PR #1, opened 2019-11-11
[TM-359] Initial repo setup

rvem:
We have nix expressions for building static `tezos-client` binary. Having this executable separate
from the `babylonnet.sh` or `mainnet.sh` scripts will be useful for tezos community.

This PR provides initial repository set up with instructions on how to build static `tezos-client` binary and
`.rpm/deb` packages based on this binary. Also set up CI to produce these packages and binary.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,24 @@
+# SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+env:
+  NIX_PATH: nixpkgs=channel:nixos-unstable
+
+steps:
+ - command: nix run nixpkgs.reuse -c reuse lint
+   label: reuse lint
+ - command: .buildkite/check-trailing-whitespace.sh
+   label: check trailing whitespace
+ - command: "nix run -f https://github.com/serokell/crossref-verifier/archive/master.tar.gz -c crossref-verify"
+   label: crossref-verify
+   soft_fail: true
+ - commands:
+   - nix-build -A tezos-client-static -o tezos-client-static
+   - nix-build -A tezos-client-rpm-package -o tezos-client-rpm
+   - nix-build -A tezos-client-deb-package -o tezos-client-deb
+   label: build and package
+   artifact_paths:
+     - ./tezos-client-static/bin/tezos-client
+     - ./tezos-client-rpm/*
+     - ./tezos-client-deb/*
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,60 @@
+# Contribution Guidelines
+
+## Reporting Issues
+
+Please [open an issue](https://github.com/serokell/tezos-client/issues/new)
+if you find a bug or have a feature request.
+Before submitting a bug report or feature request, check to make sure it hasn't already been submitted
+
+The more detailed your report is, the faster it can be resolved.
+If you report a bug, please provide steps to reproduce this bug and revision of code in which this bug reproduces.
+
+
+## Code
+
+If you would like to contribute code to fix a bug, add a new feature, or
+otherwise improve our project, pull requests are most welcome.
+
+## Legal
+
+We want to make sure that our projects come with correct licensing information
+and that this information is machine-readable, thus we are following the
+[REUSE Practices][reuse] – feel free to click the link and read about them,
+but, basically, it all boils down to the following:
+
+  * Add the following header at the very top (but below the shebang, if there
+    is one) of each source file in the repository (yes, each and every source
+    file – it is not as hard as it might sound):
+
+    ```
+    # SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
+    #
+    # SPDX-License-Identifier: MPL-2.0
+    ```
+
+    (This is an example for Nix; adapt it as needed for other languages.)
+
+    The license identifier should be the same as the one in the `LICENSE` file.
+
+  * If you are copying any source files from some other project, and they do not
+    contain a header with a copyright and a machine-readable license identifier,
+    add it, but be extra careful and make sure that information you are recording
+    is correct.
+
+    If the license of the file is different from the one used in the project and
+    you do not plan to relicense it, use the appropriate license identifier and
+    make sure the license text exists in the `LICENSES` directory.
+
+    If the file contains the entire license in its header, it is best to move the
+    text to a separate file in the `LICENSES` directory and leave a reference.
+
+  * If you are copying pieces of code from some other project, leave a note in the
+    comments, stating where you copied it from, who is the copyright owner, and
+    what license applies.
+
+  * All the same rules apply to documentation that is stored in the repository.
+
+These simple rules should cover most of situation you are likely to encounter.
+In case of doubt, consult the [REUSE Practices][reuse] document.
+
+[reuse]: https://reuse.software/spec/
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,26 @@
+## Description
+
+<!--
+Describes the nature of your changes. If they are substantial, you should
+further subdivide this into a section describing the problem you are solving and
+another describing your solution.
+-->
+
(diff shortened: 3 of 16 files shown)


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+    file – it is not as hard as it might sound):
+
+    ```haskell
+    {- SPDX-FileCopyrightText: 2019 Serokell <https://serokell.io>
+     -
+     - SPDX-License-Identifier: MPL-2.0
+     -}
+    ```
+
+    (This is an example for Haskell; adapt it as needed for other languages.)
```
gromakovsky: AFAIU, this repo is going to contain mostly `.nix` stuff, so I think it makes sense to use `.nix` syntax in example.

.github/pull_request_template.md
```diff
@@ -0,0 +1,72 @@
+<!--
+   - SPDX-FileCopyrightText: 2019 Serokell <https://serokell.io>
+   -
+   - SPDX-License-Identifier: MPL-2.0
+   -->
+[//]: # (This is a template of a pull request template.)
+[//]: # (You should modify it considering specifics of a particular repository and put it there.)
+[//]: # (Comments like this are meta-comments, they shouldn't be present in the final template.)
+[//]: # (Comments starting with '<!---' are intended to stay in the final template.)
```
gromakovsky: Remove meta-comments.

.github/pull_request_template.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Description
+
+<!--
+Describes the nature of your changes. If they are substantial, you should
+further subdivide this into a section describing the problem you are solving and
+another describing your solution.
+-->
+
+[//]: # (Here you can add a link to the corresponding issue tracker, e. g. https://issues.serokell.io/issue/AD-)
```
gromakovsky: Let's add a link to issue tracker of this repo (github's one)?

.github/pull_request_template.md
```diff
@@ -0,0 +1,72 @@
...
+  - [ ] I added an entry to the [changelog](../tree/master/CHANGES.md) if my changes are visible to the users
+        and
+  - [ ] provided a migration guide for breaking changes if possible
+
+#### Stylistic guide (mandatory)
+
+[//]: # (Update link to style guide if necesary or remove if it's not present)
+
+- [ ] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
+- [ ] My code complies with the [style guide](../tree/master/docs/code-style.md).
```
gromakovsky: Does it exist? Since it is not a Haskell repo and we don't have style guides for anything except Haskell, perhaps we can remove it.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+and that this information is machine-readable, thus we are following the
+[REUSE Practices][reuse] – feel free to click the link and read about them,
+but, basically, it all boils down to the following:
+
+  * Add the following header at the very top (but below the shebang, if there
+    is one) of each source file in the repository (yes, each and every source
+    file – it is not as hard as it might sound):
+
+    ```haskell
+    {- SPDX-FileCopyrightText: 2019 Serokell <https://serokell.io>
```
gromakovsky: I am not sure, but I suspect the copyright holder is TQ. Let's wait until we get confirmation.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,21 @@
...
+   label: check trailing whitespace
+ - commands:
+   - nix-build -A tezos-client-static -o tezos-client-static
+   - nix-build -A tezos-client-rpm-package -o tezos-client-rpm
+   - nix-build -A tezos-client-deb-package -o tezos-client-deb
+   label: build and package
+   artifact_paths:
+     - ./tezos-client-static/bin/tezos-client
+     - ./tezos-client-rpm/*
+     - ./tezos-client-deb/*
```
gromakovsky: If adding `crossref-verifier` check is easy and cheap, I'd do it here.

gromakovsky:
Does CI build binaries for `mainnet` or `babylonnet`?

rvem:
> Does CI build binaries for `mainnet` or `babylonnet`?

`mainnet` but they seem to be very similar, and both work with `babylonnet` node

.github/CONTRIBUTING.md
rvem: Do all files should be with TQ as a holder?

gromakovsky: 1. Files that are new, i. e. created during our work on `tezos-client` packaging.
2. Files that we copy-pasted from somewhere where TQ is the copyright holder.
3. Files that we copy-pasted from somewhere and that have `Unlicense` license (e. g. https://github.com/serokell/metatemplates/blob/master/templates/.github/CONTRIBUTING.md).

gromakovsky: In other words, files that shouldn't have TQ are the files that (we copy-pasted from somewhere) and (they have copyright holder other than TQ) and (their license doesn't allow to change copyright holder, i. e. we can assume that the license is not `Unlicense`).

.github/pull_request_template.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Documentation
+  - [ ] I checked whether I should update the docs and did so if necessary:
+    - [README](../tree/master/README.md)
+    - Haddock
+
+- Public contracts
+  - [ ] Any modifications of public contracts comply with the [Evolution
+  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
+  - [ ] I added an entry to the [changelog](../tree/master/CHANGES.md) if my changes are visible to the users
```
gromakovsky: 1. Well, `CHANGES.md` also doesn't exist. let's add dummy `CHANGES.md` or remove this line.
2. Please review this list carefully. E. g. why do we mention Haddock here? Are we going to have tests? How does one write tests for nix? Why do we care about evolution of public contracts? As I said, it is not a library, it doesn't have any API. I believe this checklist can be greatly simplified.

README.md
```diff
@@ -1 +1,45 @@
...
+
+Run `tezos-client man` to get more information.
+
+## For Contributors
+
+Please see [CONTRIBUTING.md](.github/CONTRIBUTING.md) for more information.
+
+## About Serokell
+
+Patak is maintained and funded with ❤️ by [Serokell](https://serokell.io/).
```
gromakovsky: Patak bardaq

.github/pull_request_template.md
rvem: Actually, it seems that only `README.md` makes sense at the current point

.github/pull_request_template.md
```diff
@@ -0,0 +1,38 @@
...
+-->
+https://github.com/serokell/tezos-client/issues/
+
+<!--
+Ideally a PR has all of the checkmarks set.
+
+If something in this list is irrelevant to your PR, you should still set this
+checkmark indicating that you are sure it is dealt with (be that by irrelevance).
+
+If you don't set a checkmark (e. g. don't add a test for new functionality),
```
gromakovsky: > (e. g. don't add a test for new functionality),

Well, now there is no such checkmark :)
Perhaps all this comment can be removed entirely since there is only 1 checkmark.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+If you report a bug, please provide steps to reproduce this bug and revision of code in which this bug reproduces.
+
+
+## Code
+
+If you would like to contribute code to fix a bug, add a new feature, or
+otherwise improve our project, pull requests are most welcome.
+
+Our pull request template contains a [checklist](./pull_request_template.md#white_check_mark-checklist-for-your-pull-request) of acceptance criteria for your pull request.
+Please read it before you start contributing and make sure your contributions adhere to this checklist.
```
gromakovsky: Since there is only 1 item, maybe this paragraph doesn't make much sense. It is useful when there are checkmarks about tests, because it's better to keep in mind that you add tests before you make a PR. In our case checklist is quite trivial.

PR #2, opened 2019-11-13
[TM-359] Update README

rvem:
## Description
Problem: "Obtain binary or packages from CI" section in the README.md
isn't quite clear. Also, there are some inaccuracies in this file.

Solution: Update README, also update `packageRpm.nix` because it
contained some incorrect stuff.

## Related issue(s)

https://issues.serokell.io/issue/TM-359

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,48 @@ Run `nix-build -A tezos-client-static -o tezos-static`
 to build staticically linked `tezos-client` executable.
 
 Run `nix-build -A tezos-client-{rpm, deb}-package -o tezos-client-package` in order
-to build native `.rpm` or `.deb` package for RedHat and Debiab-based distros.
+to build native `.rpm` or `.deb` package for Red Hat and Debian-based distros.
 
 ## Obtain binary or packages from CI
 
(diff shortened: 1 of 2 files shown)


gromakovsky:
Well, https://github.com/serokell/tezos-client/issues/ is obviously not an issue :)

README.md
```diff
@@ -19,13 +19,48 @@ Run `nix-build -A tezos-client-static -o tezos-static`
...
+sudo apt install <path to deb file>
+```
+
+### Use PPA with `tezos-client`
+
+Also if you are using Ubuntu you can use PPA in order to install `tezos-client`.
+In order to do that run the following commands:
+```
+sudo add-apt-repository ppa:serokell/tezos && sudo apt-get update
+sudo apt-get install tezos-client
```
gromakovsky: Have you tested these commands btw (I mean all commands present in this file)?

rvem: Everything except Fedora (I nuked VM with it in order to free some space for nix :(), but it this package is similar to one, I've been building earlier, so it should work as well

rvem: UPD: tested on Fedora as well

Issue #4, opened 2019-11-14
Version built packages


itkach:
looking at CI output at https://buildkite.com/serokell/tezos-client/builds/20#fd33a01d-a229-4e70-b44c-75f6a6fafe72  I see that static binary is simply called tezos-client, deb package contains version number (1.0 which is probably wrong) and rpm name contains version and arch

In the absence of any versioning from tezos-client itself I think all artifact names should include
- tezos source branch (looks like each "...net" has it's own branch, so probably makes sense to maintain several builds, but at the very least we need one for current version of protocol, i.e babylonnet)
- build architecture
- source code revision (short version)
- timestamp (unix or ISOformatted): this one should be used as a sequential version number so that package managers (apt and dnf) pulling from their respective package repos can detect that there's an upgrade

itkach:
.deb and .rpm package file names themselves should follow Debian or perhaps Ubuntu (https://askubuntu.com/questions/620533/what-is-the-meaning-of-the-xubuntuy-string-in-ubuntu-package-name) and Fedora conventions (https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/) respectively and incorporate git revision somewhere - either in package file name or in package description or other suitable metadata that can be easily inspected with dpkg and rpm command line tools

PR #5, opened 2019-11-15
Update packages and binaries names

rvem:
## Description

Problem:
1) There is no actual versioning for tezos-client, currently, we produce it with version 1.0.0, which is incorrect
2) We want to have binaries and packages for both mainnet and babylonnet.

Solution:
1) Use timestamp in ISO 8601 and short git hash as a version instead of 1.0.0, thus package manager will be able to upgrade the package once we release a new one with a later timestamp
2) Produce required files in CI as well.

https://launchpad.net/~serokell will be updated once we consider that package names are ok and this PR is merged.

## Related issue(s)

https://github.com/serokell/tezos-client/issues/4


gromakovsky:
Same comment as in #8: let's refer to the issue in commit messages (in this case to #4).

README.md
```diff
@@ -15,11 +15,13 @@ using `babylonnet.sh` or `mainnet.sh` scripts.
...
 
-Run `nix-build -A tezos-client-static -o tezos-static`
-to build staticically linked `tezos-client` executable.
+Run `nix-build -A tezos-client-{mainnet, babylonnet} -o tezos-client`
+to build staticically linked `tezos-client-mainnet` or `tezos-client-babylonnet`
+executable.
 
-Run `nix-build -A tezos-client-{rpm, deb}-package -o tezos-client-package` in order
-to build native `.rpm` or `.deb` package for Red Hat and Debian-based distros.
+Run `nix-build -A {mainnet, babylonnet}-{rpm, deb}-package -o tezos-client-package --arg timestamp $(date -I | sed 's|-||g')` in order
```
itkach: I think `$(date +"%Y%m%d")` would be better here since it's more portable (e.g. `-I` doesn't exist in Mac OS version of `date`) and explicit (and uses just one tool instead of two). Also, a more accurate timestamp, say down to minute precision `%Y%m%d%H%M`, might be better to allow multiple builds on the same day and treat have rpm/deb treat later builds as updates

README.md
```diff
@@ -15,11 +15,13 @@ using `babylonnet.sh` or `mainnet.sh` scripts.
 
 ## Build Instructions
 
-Run `nix-build -A tezos-client-static -o tezos-static`
-to build staticically linked `tezos-client` executable.
+Run `nix-build -A tezos-client-{mainnet, babylonnet} -o tezos-client`
+to build staticically linked `tezos-client-mainnet` or `tezos-client-babylonnet`
+executable.
```
itkach: would be nice to have each command flavor fully spelled out and on a separate line - for easy copy-pasting of actual executable commands

README.md
itkach: Same here, would be nice to have each command flavor fully spelled out and on a separate line - for easy copy-pasting of actual executable commands. Alternatively, a make file might be a good way to simplify invocation. E.g. `make rpm` could build  babylonnet rpm (client for current protocol version is probably of most interest which makes it a good default). Also, I wonder if rpm and deb builds can just use output of static build? that doesn't seem to be the case at the moment

rvem: > I wonder if rpm and deb builds can just use the output of static build?
It actually uses the output of static build but in terms of nix derivation

itkach:
Now getting this error when trying to build deb:

rvem:
Oh, I see, I'll fix it in a moment

README.md
```diff
@@ -42,19 +67,20 @@ Also if you are using Ubuntu you can use PPA in order to install `tezos-client`.
...
-Download `.rpm` file from the CI and double-click on it or run:
+Build or download `.rpm` file from the CI and double-click on it or run:
 ```
 sudo yum localinstall <path to the rpm file>
 ```
 
 ## Other Linux distros usage
 
-Download statically built `tezos-client` binary from the CI.
+Download or build static `tezos-client` binary from the CI.
```
gromakovsky: Probably it would be better to say "Build static `tezos-client` binary or download it from the CI".

gromakovsky:
CI failed, btw.

PR #8, opened 2019-11-22
Change link to CI

### LINKED ISSUE #3: Rename the repo
`tezos-client` is a bit bad name, it makes one think that the code of `tezos-client` is in this repo, but here we only have some scripts for packaging. We need a better name. `tezos-client-nix` or anything that has `nix` in name is also a bit bad because the fact that `nix` is used is just implementation detail.
TODO: specify name.
Even though the repo is new, it should be done carefully to avoid breaking links (e. g. to buildkite). Also maybe we'll have to change something in buildkite itself.

rvem:
## Description
Problem: Both repo and CI pipeline were renamed, however, README has
link to the old pipeline (it still works, btw).

Solution: Update link to renamed buildkite pipeline.

## Related issue(s)

Resolves #3 

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ to build native `.rpm` or `.deb` package for Red Hat and Debian-based distros.
 ## Obtain binary or packages from CI
 
 If you don't want to build these files from scratch, you can download artifacts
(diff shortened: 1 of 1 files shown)


rvem:
Just forgot about it :upside_down_face:

PR #13, opened 2019-12-04
[#12] srpm generation and Fedora copr

### LINKED ISSUE #12: Upload tezos-client's to copr
We want to provide an easy way to install tezos-client on Fedora. We want to use Fedora Copr repository for this.

# Acceptance criteria
Packages for mainnet and babylonnet are uploaded to copr, instructions in README are updated.

rvem:
## Description

Problem: We want to upload our rpm package to copr, we need `.src.rpm`
package for this.

Solution: Generate it along with `.rpm` package.

Also, update installation instructions for Fedora. 

## Related issue(s)

Resolves #12 

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,14 +95,32 @@ sudo apt-get install tezos-client-babylonnet
 
 ## Fedora (Red Hat) usage
 
+
+### Install `.rpm` package
+
 Build or download `.rpm` file from the CI and double-click on it or run:
 ```
 sudo yum localinstall <path to the rpm file>
 ```
+### Use copr package with `tezos-client`
+
+Also if you are using Fedora you can use Copr in order to install `tezos-client`.
+In order to do that run the following commands:
+```
+# use dnf
+sudo dnf copr enable @Serokell/Tezos
+sudo dnf install tezos-client-mainnet
+sudo dnf install tezos-client-babylonnet
+
+# or use yum
+sudo yum copr enable @Serokell/Tezos
+sudo yum install tezos-client-mainnet
+sudo yum install tezos-client-babylonnet
+```
 
 ## Other Linux distros usage
 
-Download or build static `tezos-client` binary from the CI.
+Build static `tezos-client` binary or download it from the CI.
 Make it executable:
 ```
 chmod +x tezos-client
--- a/default.nix
+++ b/default.nix
@@ -31,18 +31,20 @@ let
     licenseFile = "${root}/LICENSES/MPL-2.0.txt";
     description = "CLI client for interacting with tezos blockchain";
     gitRevision = mainnet.rev;
+    branchName = "mainnet";
   };
 
   packageDesc-babylonnet = packageDesc-mainnet // {
     project = "tezos-client-babylonnet";
     bin = binary-babylonnet;
     gitRevision = babylonnet.rev;
+    branchName = "babylonnet";
   };
 
   buildDeb = import ./packageDeb.nix { inherit stdenv writeTextFile dpkg; };
   buildRpm = packageDesc:
-    import ./packageRpm.nix { inherit stdenv writeTextFile rpm buildFHSUserEnv; }
-      (packageDesc // { arch = "x86_64"; });
+  import ./packageRpm.nix { inherit stdenv writeTextFile gnutar rpm buildFHSUserEnv; }
+    (packageDesc // { arch = "x86_64"; });
 
   mainnet-rpm-package = buildRpm packageDesc-mainnet;
 
--- a/packageRpm.nix
+++ b/packageRpm.nix
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
 #
 # SPDX-License-Identifier: MPL-2.0
-{ stdenv, writeTextFile, rpm, buildFHSUserEnv }:
+{ stdenv, writeTextFile, gnutar, rpm, buildFHSUserEnv }:
 pkgDesc:
 
 let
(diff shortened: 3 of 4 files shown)


rvem:
I've tested all new commands on VM with Fedora, they seem to be working fine.

README.md
```diff
@@ -92,14 +92,26 @@ sudo apt-get install tezos-client-babylonnet
...
 Build or download `.rpm` file from the CI and double-click on it or run:
 ```
 sudo yum localinstall <path to the rpm file>
 ```
+### Use copr package with `tezos-client`
+
+Also if you are using Fedora you can use Copr in order to install `tezos-client`.
+In order to do that run the following commands:
+```
+sudo yum copr enable @Serokell/Tezos
```
gromakovsky: Apparently there is `dnf` tool which is a modern replacement of `yum`: https://docs.pagure.org/copr.copr/how_to_enable_repo.html#how-to-enable-repo

Shouldn't we mention both?

packageRpm.nix
```diff
@@ -44,15 +74,16 @@ in rec {
...
+        ls SOURCES
         mkdir -p BUILD/${project}
         cp ${licenseFile} BUILD/${project}/LICENSE
 
-        mkdir -p BUILDROOT/${pkgName}/usr/local/bin
-        cp ${bin} BUILDROOT/${pkgName}/usr/local/bin/${project}
-
-        rpmbuild -bb SPECS/${project}.spec
-        cp RPMS/${arch}/${pkgName}.rpm $out
+        rpmbuild -ba SPECS/${project}.spec
```
gromakovsky: What's the difference between `-ba` and `-bb`? Can you send some links which make it easier to understand what's going on here?

packageRpm.nix
```diff
@@ -44,15 +74,16 @@ in rec {
         mkdir rpmbuild
         cd rpmbuild
         mkdir SPECS
+        mkdir SOURCES
         cp ${writeSpecFile} SPECS/${project}.spec
+        cp ${sourceArchive} SOURCES/${sourceArchive.name}
+        ls SOURCES
```
gromakovsky: Is it only for logging purposes or is the output used by the rest of the script somehow?

packageRpm.nix
```diff
@@ -12,26 +12,56 @@ let
...
 
+      %prep
+      %setup -q
+
+      %build
+
+      %install
+      mkdir -p %{buildroot}/%{_bindir}
+
+      install -m 0755 %{name} %{buildroot}/%{_bindir}/%{name}
```
gromakovsky: Where can I read about spec files?

README.md
rvem: I haven't tried dnf but I think it should work as well

packageRpm.nix
rvem: https://rpm-packaging-guide.github.io/

packageRpm.nix
rvem: Oh, I forgot to remove this line

packageRpm.nix
rvem: -bb builds only binary package, -ba builds both binary and source packages

rvem:
@gromakovsky is there anything else you want to be changed in this PR?

packageRpm.nix
```diff
@@ -12,26 +12,56 @@ let
...
     text = ''
+      %define debug_package %{nil}
+      %define _unpackaged_files_terminate_build 0
       Name:    ${project}
       Version: ${version}
       Release: ${revision}
       Summary: ${pkgDesc.description}
       License: ${pkgDesc.license}
+      BuildArch: ${arch}
+      Source0: ${project}-${version}.tar.gz
```
gromakovsky: Documentation says that it should be
> Path or URL to the compressed archive of the upstream source code (unpatched, patches are handled elsewhere). This should point to an accessible and reliable storage of the archive, for example, the upstream page and not the packager’s local storage. If needed, more SourceX directives can be added, incrementing the number each time, for example: Source1, Source2, Source3, and so on.

AFAIU, you put the license file and static binary here. Is it ok? The documentation makes me think it should contain source code from `tezos/tezos` :man_shrugging:

gromakovsky:
Okay, now I have only one question.

packageRpm.nix
rvem: AFAIR, at this point, `Source` is what package made from. In our case, it is only a static binary, but I agree that it is not entirely correct to package only it without source code. However, it seems that we are not breaking any terms doing it

rvem: I think it should be ok if we put the link to tezos/tezos as a `Source1`

default.nix
```diff
@@ -31,17 +31,19 @@ let
...
     description = "CLI client for interacting with tezos blockchain";
     gitRevision = mainnet.rev;
+    branchName = "mainnet";
   };
 
   packageDesc-babylonnet = packageDesc-mainnet // {
     project = "tezos-client-babylonnet";
     bin = binary-babylonnet;
     gitRevision = babylonnet.rev;
+    branchName = "babylonnet";
```
gromakovsky: Did it become necessary? Is it used anywhere?

gromakovsky:
Okay, one more question. Also now there are conflicts, so I'd like to take a look at the rebased version just in case.

default.nix
rvem: Forgot to remove that:(

rvem:
Rebased

rvem:
Oh wait, I broke something:(
`.rpm` files have weird content

rvem:
Okay, apparently current approach needs more changes when used outside of VM

rvem:
@gromakovsky I've fixed an issue with paths in `.rpm` package

rvem:
Also, I've checked newly generated packages on VM with Fedora

Issue #14, opened 2019-12-04
Build/package other tezos binaries 


itkach:
# Clarification and motivation

In addition to `tezos-client`, Tezos software also includes other binaries: `tezos-admin-client`, `tezos-node`, `tezos-baker`, `tezos-accuser`, `tezos-endorser`, `tezos-signer` and `tezos-protocol-compiler`. Same as with `tezos-client`, availability of easy-to-install native platform binaries would lower barrier for adoption and make it easier for developers to start experimenting and building tools and applications for Tezos platform.

# Acceptance criteria
Statically linked binaries and/or packages for popular Linux distros are available for download.

gromakovsky:
We should also rename the repo once again, probably to `tezos-packaging`.

PR #17, opened 2019-12-10
[#14] Rename repo

rvem:
## Description
Problem: We are going to package other tezos-related binaries as well,
thus `tezos-client-packaging` is not the best name for repo.

Solution: Rename it once again, now to `tezos-packaging`

Also pipeline on buildkite was renamed

## Related issue(s)

~Resolves~ Relates #14
https://issues.serokell.io/issue/OPS-901

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Reporting Issues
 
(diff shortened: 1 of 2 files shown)


rvem:
Oh, I see, `.github/`, that's why `git grep` didn't show me that file

PR #18, opened 2019-12-10
[#14] Package more tezos binaries

### LINKED ISSUE #14: Build/package other tezos binaries

rvem:
## Description
Tezos software includes more executables than just `tezos-client`. We want to package them as well.
For now, in this PR we only build static binaries for them 
(specifically `tezos-client`, `tezos-admin-client`, `tezos-node`, `tezos-baker`, `tezos-accuser`, `tezos-endorser`, `tezos-signer`, and `tezos-protocol-compiler`).

TODO:
~* Package them into `.deb` and `.rpm` as well as `tezos-client`.~
~* Update usage instructions.~

## Related issue(s)

Resolves #14

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,33 +3,16 @@
 # SPDX-License-Identifier: MPL-2.0
 ts := $(shell date +"%Y%m%d%H%M")
 
-binary:
-		nix-build -A tezos-client-babylonnet -o tezos-client
-		cp tezos-client/tezos-client-babylonnet-* tezos-client-babylonnet
-		rm tezos-client
+.PHONY: binaries binaries-mainnet rpm-packages deb-packages
 
-binary-mainnet:
-		nix-build -A tezos-client-mainnet -o tezos-client
-		cp tezos-client/tezos-client-mainnet-* ./tezos-client-mainnet
-		rm tezos-client
+binaries:
+		cp $(shell nix-build -A babylonnet-binaries --no-out-link)/* ./
 
-rpm:
-		nix-build -A babylonnet-rpm-package -o tezos-client-package --arg timestamp $(ts)
-		cp -a tezos-client-package/*.rpm .
-		rm tezos-client-package
+binaries-mainnet:
+		cp $(shell nix-build -A mainnet-binaries --no-out-link)/* ./
 
-rpm-mainnet:
-		nix-build -A mainnet-rpm-package -o tezos-client-package --arg timestamp $(ts)
-		cp -a tezos-client-package/*.rpm .
-		rm tezos-client-package
+rpm-packages:
+		nix-build -A rpm-packages -o rpm-packages --arg timestamp $(ts)
 
-deb:
-		nix-build -A babylonnet-deb-package -o tezos-client-package --arg timestamp $(ts)
-		cp -a tezos-client-package/*.deb .
-		rm tezos-client-package
-
-deb-mainnet:
-		nix-build -A mainnet-deb-package -o tezos-client-package --arg timestamp $(ts)
-		cp -a tezos-client-package/*.deb .
-		rm tezos-client-package
-
+deb-packages:
+		nix-build -A deb-packages -o deb-packages --arg timestamp $(ts)
--- a/README.md
+++ b/README.md
@@ -9,103 +9,43 @@
 [![Build status](https://badge.buildkite.com/e899e9e54babcd14139e3bd4381bad39b5d680e08e7b7766d4.svg?branch=master)](https://buildkite.com/serokell/tezos-packaging)
 
 This repo provides various form of distribution for tezos-related executables
-(unfortunately, only `tezos-client` for now, see [this issue](https://github.com/serokell/tezos-packaging/issues/14)).
+(`tezos-client`, `tezos-client-admin`, `tezos-node`, `tezos-baker`,
+`tezos-accuser`, `tezos-endorser`, `tezos-signer` and `tezos-protocol-compiler`).
 
-`tezos-client` is CLI tool used for interaction with Tezos blockchain.
-This repo contains nix expression for building staticically linked
-tezos-client binaries that can be used with remote tezos nodes without
-using `babylonnet.sh` or `mainnet.sh` scripts.
+## Obtain binaries or packages from github release
 
-## Build Instructions
-
-### Statically built binary
-
-Run one of the following commands:
-```
-nix-build -A tezos-client-mainnet -o tezos-client
-nix-build -A tezos-client-babylonnet -o tezos-client
-```
-
-Or use Makefile:
-```bash
-make binary #build tezos-client-babylonnet
-make binary-mainnet #build tezos-client-mainnet
-```
-
-To build `mainnet` or `babylonnet` versions of `tezos-client` executable
-
-### Ubuntu `.deb` package
-
-Run one of the following commands:
-```
-nix-build -A mainnet-deb-package -o tezos-client-package --arg timestamp $(date +"%Y%m%d%H%M")
-nix-build -A babylonnet-deb-package -o tezos-client-package --arg timestamp $(date +"%Y%m%d%H%M")
-```
-
-Or use Makefile:
-```bash
-make deb #build deb package with tezos-client-babylonnet
-make deb-mainnet #build deb package with tezos-client-mainnet
-```
-
-To build `.deb` package with `mainnet` or `babylonnet` `tezos-client` executable. Once you install
-such package the command `tezos-client-mainnet` or `tezos-client-babylonnet` will be available.
-
-### Fedora `.rpm` package
-
-Run one of the following commands:
-```
-nix-build -A mainnet-rpm-package -o tezos-client-package --arg timestamp $(date +"%Y%m%d%H%M")
-nix-build -A babylonnet-rpm-package -o tezos-client-package --arg timestamp $(date +"%Y%m%d%H%M")
-```
-
-Or use Makefile:
-```bash
-make rpm #build rpm package with tezos-client-babylonnet
-make rpm-mainnet #build rpm package with tezos-client-mainnet
-```
-
-To build `.rpm` package with `mainnet` or `babylonnet` `tezos-client` executable. Once you install
-such package the command `tezos-client-mainnet` or `tezos-client-babylonnet` will be available.
-
-## Obtain binary or packages from github release
-
-If you don't want to build these files from scratch, you can download assets from github release.
+Recomended way to get these binaries is to download them from assets from github release.
 Go to the [latest release](https://github.com/serokell/tezos-packaging/releases/latest)
 and download desired assets.
 
 ## Ubuntu (Debian based distros) usage
 
-### Install `.deb` package
+### Use PPA with `tezos-*` binaries
 
-Build or download `.deb` file from the CI and double-click on it or run:
-```
-sudo apt install <path to deb file>
-```
-
-### Use PPA with `tezos-client`
-
-Also if you are using Ubuntu you can use PPA in order to install `tezos-client`.
-In order to do that run the following commands:
+If you are using Ubuntu you can use PPA in order to install `tezos-*` executables.
+E.g, in order to do install `tezos-client` run the following commands:
 ```
 sudo add-apt-repository ppa:serokell/tezos && sudo apt-get update
 sudo apt-get install tezos-client-mainnet
 sudo apt-get install tezos-client-babylonnet
 ```
+Once you install such packages the commands `tezos-*` will be available.
+
+### Install `.deb` package locally
+
+As an alternative, you can download `.deb` file from release assets
+(they are located in `deb-packages.tar.gz` archive) and double-click on it or run:
+```
+sudo apt install <path to deb file>
+```
 
 ## Fedora (Red Hat) usage
 
+### Use copr package with `tezos-*` binaries
 
-### Install `.rpm` package
-
-Build or download `.rpm` file from the CI and double-click on it or run:
-```
-sudo yum localinstall <path to the rpm file>
-```
-### Use copr package with `tezos-client`
-
-Also if you are using Fedora you can use Copr in order to install `tezos-client`.
-In order to do that run the following commands:
+If you are using Fedora you can use Copr in order to install `tezos-*`
+executables.
+E.g. in order to install `tezos-client` run the following commands:
 ```
 # use dnf
 sudo dnf copr enable @Serokell/Tezos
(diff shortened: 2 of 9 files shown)


gromakovsky:
@itkach today you said that packaging into deb and rpm should be discussed. Can you please write your thoughts about that? Whether there should be 1 package per binary or 1 package with all binaries or some other grouping…

itkach:
I was thinking that it might make sense to put each binary into a separate deb/rpm but also have virtual packages that "bundle" some of them together, e.g. "tezos-all" would have all the other packages listed as a dependency, so installing "tezos-all" would pull down everything, or "tezos-baking" would pull down tezos-node, tezos-baker, tezos-client and tezos-admin-client. But perhaps this is overthinking it and individual packages, one-per-binary, is good enough.

rvem:
Also, probably bundle with `tezos-client` and `tezos-node` will be useful for contract developers

rvem:
One package per binary for now

rvem:
Launchpad PPA and Copr repo will be updated once this PR is merged

default.nix
```diff
@@ -10,13 +10,15 @@ let
...
+    protoName = "005_PsBabyM1";
+    binarySuffix = "005-PsBabyM1";
   };
   babylonnet = {
     rev = "b8731913";
     sha256 = "1pakf1s6bg76fq42mb8fj1immz9g9wwimd522cpx8k28zf0hkl5i";
     patchFile = ./nix/fix-babylonnet.patch;
-    protoName = "proto_005_PsBabyM1";
+    protoName = "005_PsBabyM1";
+    binarySuffix = "005-PsBabyM1";
```
gromakovsky: It seems to be the same as `protoName` except that `_` is replaced with `-`. I don't know what these values mean, but I wonder if it is accidentally so or `binarySuffix` is always equal to `protoName` with `_` replaced with `-`. In the latter case it seems better to define `binarySuffix` in terms of `protoName` (so that we'll have to update less things in case of new protocol), but there is another thing I don't know: whether it is hard to do that in `nix`. So that's up to you.

gromakovsky:
Btw,
> Problem: Now we build not only `tezos-client` but also other
tezos-related executables.

is not a problem per se.

Makefile
```diff
@@ -3,33 +3,18 @@
...
 ts := $(shell date +"%Y%m%d%H%M")
 
-binary:
-		nix-build -A tezos-client-babylonnet -o tezos-client
-		cp tezos-client/tezos-client-babylonnet-* tezos-client-babylonnet
-		rm tezos-client
+binaries:
+		nix-build -A babylonnet-binaries -o binaries
+		cp binaries/* ./
+		rm binaries
```
gromakovsky: AFAIU `binaries` is a directory and it is likely not empty. Will it work w/o `-r`?

rvem: At this point `binaries` directory has only one `.tar.gz` archive inside. Everything is fine for me locally:
```
XPS-15-9560 :: work/serokell/tezos-client-packaging » make binaries
nix-build -A babylonnet-binaries -o binaries
/nix/store/dn4cmd7m03qsq6ijzq0lhklb3ch6y01f-tezos-babylonnet-binaries-b8731913.tar.gz
cp binaries/* ./
rm binaries
XPS-15-9560 :: work/serokell/tezos-client-packaging » echo $?
0
```

default.nix
rvem: Fair enough

Makefile
gromakovsky: > This manual page documents the GNU version of rm.  rm removes each specified file.  By default, it does not remove directories.

Strange :man_shrugging:

rvem: ah, I misunderstood you, at this point `binaries` is actually a symlink to some path in `/nix/store`

gromakovsky: Ah, coolstory, ok then.

default.nix
```diff
@@ -6,72 +6,126 @@ with pkgs;
 
 let
   root = ./.;
-  mainnet = {
+  mainnet = rec {
     rev = "94f779a7";
     sha256 = "16lxilng5q8fr2ll6h4hf7wlvac6nmw4cx10cbgzj5ks090bl97r";
     patchFile = ./nix/fix-mainnet.patch;
+    protoName = "005_PsBabyM1";
```
yorickvP: are you sure?

default.nix
```diff
@@ -6,72 +6,126 @@ with pkgs;
...
+    "${tezos-static-babylonnet}/bin";
+
+  mkPackageDescs = { executableName, binPath, description }@executableDesc:
+    let
+      mainnetDesc = {
+        inherit description;
+        bin = "${tezos-static-mainnet}/${binPath}";
+        gitRevision = mainnet.rev;
+        project = "${executableName}-mainnet";
+        version = toString timestamp;
```
yorickvP: wait, why timestamp? If you must, grab it from git somehow maybe?

default.nix
```diff
@@ -6,72 +6,126 @@ with pkgs;
...
+      binPath = "/bin/tezos-admin-client";
+      description = "Administration tool for the node";
+    }
+    { executableName = "tezos-node";
+      binPath = "/bin/tezos-client";
+      description =
+        "Entry point for initializing, configuring and running a Tezos node";
+    }
+    { executableName = "tezos-baker";
+      binPath = "/bin/tezos-baker-${mainnet.binarySuffix}";
```
yorickvP: hardcoded to mainnet?

default.nix
```diff
@@ -6,72 +6,126 @@ with pkgs;
...
+  moveDerivations = name: drvs:
+    stdenv.mkDerivation rec {
+      inherit name drvs;
 
-  babylonnet-rpm-package = buildRpm packageDesc-babylonnet;
+      buildCommand = ''
+        mkdir -p ${name}
+        backup=$PWD
+        for drv in $drvs; do
+          cd $drv
```
yorickvP: why all the cd's instead of absolute paths?

nix/default.nix
```diff
@@ -229,13 +229,56 @@ let
...
+            owner = "OCamlPro";
+            repo = "${pname}";
+            rev = "v${version}";
+            sha256 = "0smfwrj8qhzknhzawygxi0vgl2af4vyi652fkma59rzjpvscqrnn";
+          };
+          buildInputs = [ pprint astring base ];
+          buildPhase = "";
+          installPhase = ''
+            mkdir -p $out/lib/ocaml/4.07.1/site-lib
+            mkdir -p $out/lib/ocaml/4.07.1/site-lib//../bin
```
yorickvP: ?!?!

nix/default.nix
```diff
@@ -274,18 +317,48 @@ let
...
+          dune build src/proto_${branchInfo.protoName}/bin_endorser/tezos-endorser-${branchInfo.binarySuffix}.install
+          dune build src/bin_signer/tezos-signer.install
+          dune build src/lib_protocol_compiler/tezos-protocol-compiler.install
+        '';
         installPhase = ''
           mkdir -p $out/bin
+          # tezos-client and tezos-admin
           cp _build/default/src/bin_client/main_client.exe $out/bin/tezos-client
-          cp _build/default/src/bin_client/main_admin.exe $out/bin/tezos-admin
+          cp _build/default/src/bin_client/main_admin.exe $out/bin/tezos-admin-client
```
yorickvP: maybe just loop over these produced exes? are they always called main_?

nix/fix-babylonnet.patch
```diff
@@ -16,4 +16,96 @@ index 6bee8e98..53006632 100644
 +                    )))
  
  (install
-  (section bin)
\ No newline at end of file
+  (section bin)
```
yorickvP: very annoying that there's no dune option for these

nix/default.nix
rvem: probably `//` should be `/`, but `ocp-ocamlres` really wants to have `/../bin` in order to install binary
https://github.com/OCamlPro/ocp-ocamlres/blob/master/Makefile#L2

default.nix
rvem: AFAIR, then there is going to be this `/nix/store/<hash>` prefixes

default.nix
rvem: oh, you are right

default.nix
rvem: we came up to an agreement, that we are going to use a timestamp as a version since `tezos-*` stuff lack any kind of versioning

default.nix
rvem: `proto` this should mean `protocol`
e.g. https://gitlab.com/tezos/tezos/tree/babylonnet/src/proto_005_PsBabyM1

default.nix
rvem: and not the timestamp from the tezos git revision, but the time, when static binary/package was built

nix/default.nix
rvem: > are they always called main_?
Nope, `tezos-node` has a simple `main.exe` executable.
So it seems that loop over `main*.exe` is something non-trivial since we cannot always get `tezos-smth` name by the `.exe` name

rvem:
> Did you test that all of these run and don't insta-segfault?

Yes, also CI successfully produces everything

default.nix
rvem: to be more specific, timestamp of package building, not the revision timestamp

gromakovsky:
Hm, it seems that things in `tezos-mainnet-binaries-94f779a7.tar.gz` don't have `w` permission and I can't remove them. Take a look at https://github.com/serokell/scratch/pull/7/files#diff-d70f7a208d7ea7a7314740f0e69eb9acR25, I had a similar issue there.

gromakovsky:
@rvem can you take care of ↑? Also there seem to be some conflicts.

default.nix
rvem: ah, lol, I'm idiot

rvem:
@gromakovsky `w` permission on archive files seems to be fixed now

default.nix
```diff
@@ -6,73 +6,124 @@ with pkgs;
...
-    arch = "amd64";
-    license = "MIT";
-    dependencies = "";
-    maintainer = "Serokell https://serokell.io";
-    licenseFile = "${tezos-client-static-mainnet}/LICENSE";
-    description = "CLI client for interacting with tezos blockchain";
-    gitRevision = mainnet.rev;
-    branchName = "mainnet";
-  };
+  tezos-static-mainnet = import ./nix/static.nix mainnet;
```
yorickvP: not sure if nix does common subexpr elimination like this, please move to separate `let` entry?
```suggestion
  static-nix = import ./nix/static.nix;
  tezos-static-mainnet = static-nix mainnet;
```

default.nix
yorickvP: Poor deterministic nix build, tainted by human concepts such as time.

rvem:
Also, now some things have to be updated in release script since now we have binaries in tar archive, but we want to provide binaries as is in the releases

gromakovsky:
Update about release artifacts:
1. Let's provide each individual static binary **and** archive with all binaries. I suppose two archives: one for mainnet, one for babylonnet. The former is useful when someone needs just one (or a few) binary, the latter is useful for those who need pretty much everything.
2. Let's put `.deb` and `.rpc` packages into archives and not put individual packages into release.

README.md
```diff
@@ -120,20 +114,17 @@ sudo yum install tezos-client-babylonnet
 
 ## Other Linux distros usage
 
-Build static `tezos-client` binary or download it from the CI.
+Build static `tezos-*` binary or download it from the CI.
```
gromakovsky: I think from now on we can assume that CI artifacts are only for internal usage by maintainers of this repo and people should download stuff from releases.

So here I would suggest downloading from releases rather than CI.

As a side note, I think the order of entities in README is quite wrong:
1. Usage instructions should be before build instructions.
2. For Ubuntu and Fedora we should suggest PPA/copr before downloading `.deb` or `.rpm` packages.

And btw, there we also suggest downloading from CI rather than releases.

gromakovsky:
Ok, LGTM and approved by me already. Needs an approval from @yorickvP.

default.nix
```diff
@@ -6,72 +6,109 @@ with pkgs;
...
+    binarySuffix = builtins.replaceStrings [ "_" ] [ "-" ] protocolName;
+  };
+  mainnet = rec {
     rev = "94f779a7";
     sha256 = "16lxilng5q8fr2ll6h4hf7wlvac6nmw4cx10cbgzj5ks090bl97r";
     patchFile = ./nix/fix-mainnet.patch;
+    protocol = protocol005;
   };
-  babylonnet = {
+  babylonnet = rec {
```
yorickvP:
```suggestion
  babylonnet = {
```

default.nix
```diff
@@ -6,72 +6,109 @@ with pkgs;
 
 let
   root = ./.;
-  mainnet = {
+  protocol005 = rec {
```
yorickvP:
```suggestion
  protocol005 = {
```

default.nix
```diff
@@ -6,72 +6,109 @@ with pkgs;
...
+    protocolName = "005_PsBabyM1";
+    binarySuffix = builtins.replaceStrings [ "_" ] [ "-" ] protocolName;
+  };
+  mainnet = rec {
```
yorickvP:
```suggestion
  mainnet = {
```

default.nix
```diff
@@ -6,72 +6,109 @@ with pkgs;
...
+  tezos-static-babylonnet = static-nix babylonnet;
 
-  mainnet-rpm-package = buildRpm packageDesc-mainnet;
+  packDirectory = archiveName: pathToPack:
+    stdenv.mkDerivation rec {
+      name = "${archiveName}.tar.gz";
+      phases = "archivePhase";
+      nativeBuildInputs = [ gnutar ];
+      archivePhase = ''
+        mkdir -p ${archiveName}
```
yorickvP: just figure out the right tar flags to do this without moving stuff around

default.nix
rvem: Nope, this one is required :upside_down_face:

default.nix
rvem: Updated, btw, now no moving or copying the stuff around

PR #22, opened 2019-12-23
[#16] Update license in packages

### LINKED ISSUE #16: Update license in packages

From what I see, generated packages (both .deb and .rpm) say that the license is MPL and contain `MPL-2.0.txt` file from this repo. However, MPL license applies only to the files in this repo, while `tezos-client` itself has a different license (https://gitlab.com/tezos/tezos/blob/master/LICENSE).

Now I can't find where we use license file in deb packaging, but I found it somewhere yesterday and I see that `debian/copyright` file contains MPL-2.0 license. Maybe it was attached during upload to PPA (I downloaded it from there).

# Acceptance criteria

Both packages should contain the license from `tezos/tezos` repo, not MPL-2.0.

rvem:
Problem: All packages have MPL-2.0 license, however, it is a bit
incorrect.

Solution: Use license from tezos repo instead.

## Related issue(s)

Resolves #16 

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -25,10 +25,10 @@ let
     version = toString timestamp;
     bin = binary-mainnet;
     arch = "amd64";
-    license = "MPL-2.0";
+    license = "MIT";
     dependencies = "";
     maintainer = "Serokell https://serokell.io";
-    licenseFile = "${root}/LICENSES/MPL-2.0.txt";
+    licenseFile = "${tezos-client-static-mainnet}/LICENSE";
     description = "CLI client for interacting with tezos blockchain";
     gitRevision = mainnet.rev;
     branchName = "mainnet";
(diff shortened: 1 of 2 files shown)


gromakovsky:
>  All packages has

have (in commit message)

default.nix
```diff
@@ -25,10 +25,10 @@ let
     version = toString timestamp;
     bin = binary-mainnet;
     arch = "amd64";
-    license = "MPL-2.0";
+    license = "OSL";
```
gromakovsky: AFAIU that license in exactly MIT, so I think it's better to write "MIT" here, it is more recognizable.

nix/default.nix
```diff
@@ -286,6 +286,8 @@ let
           mkdir -p $out/bin
           cp _build/default/src/bin_client/main_client.exe $out/bin/tezos-client
           cp _build/default/src/bin_client/main_admin.exe $out/bin/tezos-admin
+          # Reuse license from tezos repo in packaging
+          cp LICENSE $out/LICENSE
```
gromakovsky: Btw, is my understanding correct that if in a new revision they rename/move `LICENSE` file somewhere, this `cp` will fail and CI will fail, so we'll notice it?

rvem: yep

PR #24, opened 2020-01-09
[#11] Add autorelease script

### LINKED ISSUE #11: Make it easier to get static binary

Currently if one wants to get `tezos-client` static binary they should
> Go to the latest master build, click on build and package stage, choose Artifacts section and download files by clcking on the filenames.

It's not too many steps, but still a bit inconvenient if the only reason why I opened this repo is to get static binary.

# Acceptance criteria

A bit subjective: make it easier to obtain static `tezos-client` binary. We don't really care about `.deb` because people can use `add-apt-repository` and it's written in README. I am not sure about `.rpm`. Anyway, if it's easy to do the same for `.deb` and `.rpm`, let's do it.

Possible options:
1. Use GitHub releases, somehow ensure that up-to-date binaries are there (either create a new release automatically or at least add a checkmark to PR template).
2. Upload artifacts somewhere, get a link pointing to the latest "somewhere", put it into README.

rvem:
## Description
Problem: We want an easier and more convenient way to get static binaries
(also packages, but this is secondarily, since we have PPAs).

Solution: Add autorelease script that produces github prerelease with
all closures (binaries and packages).

## Related issue(s)

Resolves #11 

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -33,5 +33,4 @@ steps:
  - commands:
    - GITHUB_TOKEN=$(cat ~/niv-bot-token) ./scripts/autorelease.sh
    label: create auto pre-release
-   # TODO: uncomment me before merging
-   # branches: master
+   branches: master
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -29,3 +29,9 @@ steps:
      - ./mainnet-deb-package/*
      - ./babylonnet-rpm-package/*
      - ./babylonnet-deb-package/*
+   branches: !master
+ - commands:
+   - GITHUB_TOKEN=$(cat ~/niv-bot-token) ./scripts/autorelease.sh
+   label: create auto pre-release
+   # TODO: uncomment me before merging
+   # branches: master
--- a/default.nix
+++ b/default.nix
@@ -20,7 +20,7 @@ let
   tezos-client-static-babylonnet = import ./nix/static.nix babylonnet;
   binary-mainnet = "${tezos-client-static-mainnet}/bin/tezos-client";
   binary-babylonnet = "${tezos-client-static-babylonnet}/bin/tezos-client";
-  # Hopefully, there is always will be single LICENSE for all binaries and branches
+  # Hopefully, there will always be a single LICENSE for all binaries and branches
   licenseFile = "${tezos-client-static-mainnet}/LICENSE";
   packageDesc-mainnet = {
     inherit licenseFile;
--- a/default.nix
+++ b/default.nix
@@ -20,7 +20,10 @@ let
   tezos-client-static-babylonnet = import ./nix/static.nix babylonnet;
   binary-mainnet = "${tezos-client-static-mainnet}/bin/tezos-client";
   binary-babylonnet = "${tezos-client-static-babylonnet}/bin/tezos-client";
+  # Hopefully, there is always will be single LICENSE for all binaries and branches
+  licenseFile = "${tezos-client-static-mainnet}/LICENSE";
   packageDesc-mainnet = {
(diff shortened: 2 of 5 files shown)


.buildkite/pipeline.yml
```diff
@@ -29,3 +29,9 @@ steps:
      - ./mainnet-deb-package/*
      - ./babylonnet-rpm-package/*
      - ./babylonnet-deb-package/*
+   branches: !master
+ - commands:
+   - GITHUB_TOKEN=$(cat ~/niv-bot-token) nix-shell -p git gitAndTools.hub --run "bash scripts/autorelease.sh"
```
balsoft: `GITHUB_TOKEN=$(cat ~/niv-bot-token) ./scripts/autorelease.sh`

scripts/autorelease.sh
```diff
@@ -0,0 +1,31 @@
...
+# Delete release
+hub release delete auto-release
+
+# Update the tag
+git fetch # So that the script can be run from an arbitrary checkout
+git tag -f auto-release
+git push --force --tags
+
+# Combine all assets
+assets=""
```
balsoft: Why not create a tarball with all the assets?

rvem: Because customers want to be able to have direct links to the binaries:
https://github.com/serokell/tezos-packaging/issues/11#issuecomment-571239514

rvem: Moreover, the number of assets is going to increase soon, and end-users are likely will not want to download all binaries/packages

.buildkite/pipeline.yml
rvem: updated

gromakovsky:
I would also include `LICENSE` file from `tezos/tezos` to be safe. It seems to be already included into `.deb` and/or `.rpm` package, but it seems safer to add it directly.

release.nix
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
```
balsoft:
```
# SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
#
{ pkgs ? import <nixpkgs> { }, timestamp ? "19700101" }:
let closures = builtins.attrValues (import ./. { inherit pkgs timestamp; });
in pkgs.runCommandNoCC "release" { inherit closures; } ''
  mkdir -p $out
  for closure in $closures; do
    cp $closure/* $out/
  done
''
```
or alternatively if `hub release` will resolve symlinks
```
# SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
#
{ pkgs ? import <nixpkgs> { }, timestamp ? "19700101" }:
pkgs.symlinkJoin {
  name = "release";
  paths = builtins.attrValues (import ./. { inherit pkgs timestamp; });
}
```
are both simpler, more elegant and easier to maintain.

balsoft:
Also, please provide issue number.

rvem:
@balsoft https://issues.serokell.io/issue/TM-359

default.nix
```diff
@@ -20,15 +20,17 @@ let
   tezos-client-static-babylonnet = import ./nix/static.nix babylonnet;
   binary-mainnet = "${tezos-client-static-mainnet}/bin/tezos-client";
   binary-babylonnet = "${tezos-client-static-babylonnet}/bin/tezos-client";
+  # Hopefully, there is always will be single LICENSE for all binaries and branches
```
gromakovsky: Redundant "is"?

yorickvP:
```suggestion
  # Hopefully, there will always be a single LICENSE for all binaries and branches
```

mkaito: "there will always be"

PR #31, opened 2020-02-07
 [#30] Build binaries from master branch sources

### LINKED ISSUE #30: Build everything for carthagennet, research the difference

Carthagennet network is alive. I am not sure whether it is running Cartage protocol, but most likely it does, and even if it does not, it most likely will soon. So we should build stuff for carthagennet as well. There is a corresponding branch: https://gitlab.com/tezos/tezos/commits/carthagenet

However, it is not clear whether there is any substantial differences between these branches and whether they are compatible. I heard that genesis public key is embedded into code, so that's one of the reasons why branches can be incompatible. But I suppose it applies only to a subset of executables, not to all of them. I also heard that Tezos developers want to stop having per-network branches, but I don't know any details.

With three networks number of assets becomes quite large, so we should decrease it if possible.

# Acceptance criteria

1. Everything we build should be built not only for mainnet and babylonnet, but also carthagennet.
2. Figure out the difference between branches and whether we can decrease number of assets. Ideally also figure out some details about getting rid of per-network branches (maybe ask in tezos-dev).

rvem:
## Description
Problem: We want to support all existing chains. Currently, we support
babylonnet and mainnet, also there is carthagenet. Binaries per chain

Solution: Turned out, that binaries built from master branch sources can
be used with multiple chains. E.g. `tezos-client` can work with all
existing networks, `tezos-node` has `--network` option. Baker, endorser
and accuser are protocol specific, thus `tezos-*-005-PsBabyM1` can be
used with both mainnet and babylonnet node and `tezos-*-006-PsCARTHA`
can be used with carthagenet (at least for now).

Copr and Launchpad repos will be updated, once this PR is merged.

## Related issue(s)

Resolves #30 


gromakovsky:
> Binaries per chain

Is this sentence complete?

rvem:
> Is this sentence complete?

Mda, someone ate part of my commit message. Now it's quite hard to remember what I meant to say in that place.

PR #35, opened 2020-02-20
Make source patches configurable

rvem:
## Description
Problem: In one of our projects we want to provide patched static
tezos-binaries.

Solution: Make patches configurable.

## Related issue(s)

None

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -365,7 +365,7 @@ let
 
         minimumOCamlVersion = "4.07";
 
-        patches = [ branchInfo.patchFile ];
+        patches = branchInfo.patches;
         src = fetchgit {
           url = "https://gitlab.com/tezos/tezos.git/";
           rev = branchInfo.rev;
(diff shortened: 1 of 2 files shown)


nix/default.nix
```diff
@@ -365,7 +365,7 @@ let
 
         minimumOCamlVersion = "4.07";
 
-        patches = [ branchInfo.patchFile ];
+        patches = [ branchInfo.patches ];
```
gromakovsky: AFAIU it was a list of patches (with one item), but now it is a list of lists of patches, but I barely know `nix`. Can you clarify where it is equivalent? If it was Haskell I would say it should be `patches = branchInfo.patches`

rvem: :think: this looks like a bug, but nix somehow figured it out

gromakovsky: So I'd like to know whether nix accidentally figured it out or it is reliable and specified behavior.

rvem: I assume, that it performs list flattening somewhere, but I haven't found any proofs yet

gromakovsky: Anyway, don't you think that `patches = branchInfo.patches` would be less confusing?

rvem: yeah, I agree

PR #40, opened 2020-03-09
autorelease.sh: add `set -euo pipefail`, clean up on failure

zhenyavinogradov:
Add stricter failure checks and remove temporary files

### DIFF
--- a/scripts/autorelease.sh
+++ b/scripts/autorelease.sh
@@ -4,11 +4,18 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
+set -euo pipefail
+
 # Project name, inferred from repository name
 project=$(basename "$(pwd)")
 
 # The directory in which artifacts will be created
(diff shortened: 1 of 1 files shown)


zhenyavinogradov:
It was failing if the release did not exist before, now it's fixed

zhenyavinogradov:
The script failed randomly, but worked after a retry, I don't know why: https://buildkite.com/serokell/tezos-packaging/builds/253#45ea1d1b-0316-4b46-890d-48b530ba86af

yorickvP:
Duplicate value for 'name'. Maybe some of the github servers have a constraint for it?

zhenyavinogradov:
If the error happened when attaching an asset to the release, it would mean there already was an asset with the same name, which github does prohibit. But it does not make sense why it would occur for a just created release. It's also confusing that `hub` says 4 assets failed to upload but shows a single error response

PR #50, opened 2020-05-06
[#47] Tezos releases in versioning

### LINKED ISSUE #47: Tezos now has some versioning

It looks like Tezos developers finally started using some versions. E. g. you can see [v7.0-rc1 tag](https://gitlab.com/tezos/tezos/-/tree/v7.0-rc1).

# Acceptance criteria

1. Research how they do versioning.
2. Decide how we should do versioning in this repo. Write it down somewhere. Probably something is the repo should be updated accordingly (check that).
3. Make a new release for 7.0-rc1?

rvem:
## Description
Problem: Tezos now uses some versioning. `v7.0-rc1` recently was
released. We should use this versioning.

Solution: Optionally replace commit hash in the package version with
tag in case we are using sources that are not from the master branch
(git treats release tags the same way as branches).

We still want to be able to do multiple releases within single tezos
release or make releases that are based on fresh tezos code that wasn't
released.

## Related issue(s)

Resolves #47

### DIFF
--- a/deb.nix
+++ b/deb.nix
@@ -6,15 +6,16 @@ pkg:
 let
   project = pkg.meta.name;
   version = meta.version;
-  revision = meta.gitRevision;
+  release = meta.release;
+  epoch = meta.epoch;
   pkgArch = meta.arch;
-  pkgName = "${project}_0ubuntu${version}-${revision}_${pkgArch}";
+  pkgName = "${project}_0ubuntu${version}-${release}_${pkgArch}";
 
   writeControlFile = writeTextFile {
     name = "control";
     text = ''
       Package: ${project}
-      Version: ${version}-${revision}
+      Version: ${epoch}:${version}-${release}
       Priority: optional
       Architecture: ${meta.arch}
       Depends: ${meta.dependencies}
--- a/release.nix
+++ b/release.nix
@@ -9,7 +9,7 @@ let
   release-notes = writeTextDir "release-notes.md" ''
     Automatic release on ${builtins.substring 0 8 timestamp}
 
-    This release contains assets based on [revision ${source.rev}](https://gitlab.com/tezos/tezos/tree/${source.rev}) of ${source.ref} branch.
+    This release contains assets based on [${source.ref} release](https://gitlab.com/tezos/tezos/tree/${source.ref}).
 
     Binaries without protocol suffixes support the following protocols (unsupported protocols are listed for reference):
     ${builtins.concatStringsSep "\n" (map (x: "- [ ] `${x}`") protocols.ignored
@@ -24,10 +24,11 @@ let
     name = "tezos-release-no-tarball";
     paths = [ "${bundled.binaries}/bin" LICENSE release-notes ];
   };
-  releaseTarball = runCommand "release-tarball" { }
-    "mkdir $out; tar --owner=serokell:1000 --mode='u+rwX' -czhf $out/release.tar.gz -C ${releaseNoTarball} .";
+  tarballName = "binaries-${builtins.replaceStrings ["v"] [""] source.ref}-1.tar.gz";
+  binariesTarball = runCommand "binaries-tarball" { }
+    "mkdir $out; tar --owner=serokell:1000 --mode='u+rwX' -czhf $out/${tarballName} -C ${bundled.binaries}/bin .";
   LICENSE = writeTextDir "LICENSE" (builtins.readFile "${source}/LICENSE");
 in buildEnv {
   name = "tezos-release";
-  paths = [ releaseNoTarball releaseTarball ];
+  paths = [ releaseNoTarball binariesTarball ];
 }
--- a/debSource.nix
+++ b/debSource.nix
@@ -7,9 +7,9 @@ pkg:
 let
   project = lib.toLower pkg.meta.name;
   version = meta.version;
-  revision = meta.gitRevision;
+  release = meta.release;
+  epoch = meta.epoch;
   pkgArch = meta.arch;
-  pkgName = "${project}_0ubuntu${version}-${revision}_${pkgArch}";
   inherit (vmTools)
     makeImageFromDebDist commonDebPackages debDistros runInLinuxImage;
   ubuntuImage = makeImageFromDebDist (debDistros.ubuntu1804x86_64 // {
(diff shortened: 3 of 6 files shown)


gromakovsky:
And how will versions be formed for Debian and Fedora packages vs how they are formed now?

rvem:
> And how will versions be formed for Debian and Fedora packages vs how they are formed now?

`<timestamp>-<commithash>` when we package not Tezos release

`<timestamp>-<release-tag>` when we package some Tezos release

gromakovsky:
Ok, I am fine with that, but CI is failing currently. One of the reasons for it is compatibility with current versions, right?

I'd like to know what @itkach thinks about that.

rvem:
CI fails because it is restricted to have `-` symbol in the rpm spec file `Release` field:(

itkach:
I think with proper upstream release versions timestamps should no longer be used. For RPMs, `Epoch` tag in the spec can be used to indicate the switch to new versioning scheme and the precedence over timestamp-based packages. Hopefully deb has some equivalent mechanism

gromakovsky:
Tezos just recently release 7.0, so we should update to 7.0 from 7.0-rc.

gromakovsky:
Places where some form of a version appears:
1. Tags in this repo from which releases are made.
2. Some versioning is used by RPMs and DEBs. If at some point we supposed other package repositories, versions will be there as well.
3. In release artifacts we have `binaries-COMMIT_SHA.tar.gz`.

My proposal is the following. If we make a release from a Tezos release (or release candidate), we
1. Create the same tag as in `tezos/tezos` repo, `v7.0` in our current case.
2. Create deb and rpm packages with the same version as in `tezos/tezos`. `7.0` in our current case.
3. In release artifacts I guess we can also use version: `binaries-VERSION.tar.gz`, e. g. `binaries-7.0.tar.gz`.

If we make another release from Tezos version for which we already have a release (e. g. we have changed something in `tezos-packaging`), we append `-N` (a number) to the version, e. g. `7.0-1`, `7.0-2`, etc.

If we ever want to make a new release of `tezos-packaging` for some intermediate `tezos/tezos` version, we can use timestamps as we do now or something else. Let's skip this topic for now and get back to it if we ever need to make a release from an intermediate revision.

TBD:
1. We can always have `-N` part starting from `-0` or `-1`. Or we can assume that if it's missing it means `-0` or `-1`. What is better?
2. Do we want to make releases for release candidates?
3. Related to the previous one: if upstream version has a hyphen (e. g. `7.0-rc1`) what will we do with it? RPM's `Release` field prohibits it.

gromakovsky:
@rvem can we proceed with this? (2) and (3) don't matter at this point as we are making a release for 7.0 (not rc). As for (1), I don't have a preference and it sounds like a minor cosmetic thing, so I think you can just do as you wish.

rvem:
Yes, I'll update the PR according to our conclusion about versioning

itkach:
> We can always have -N part starting from -0 or -1. Or we can assume that if it's missing it means -0 or -1. What is better?

I think 1 is better, just because people tend to number things starting from one (first man in space, 1st Street).

itkach:
>     3\. Related to the previous one: if upstream version has a hyphen (e. g. `7.0-rc1`) what will we do with it? RPM's `Release` field prohibits it.

RPM Release tag can contain both RC info and additional designation, e.g.

Version: 7.0
Release: rc1.serokell.1234

rvem:
Also, AFAIU, rpm spec file requires some release, so I think we should start with `7.0-1` for both `.rpm` and `.deb` packages.

release.nix
```diff
@@ -9,7 +9,8 @@ let
   release-notes = writeTextDir "release-notes.md" ''
     Automatic release on ${builtins.substring 0 8 timestamp}
 
-    This release contains assets based on [revision ${source.rev}](https://gitlab.com/tezos/tezos/tree/${source.rev}) of ${source.ref} branch.
+    This release contains assets based on
+      else "[${source.ref} release]((https://gitlab.com/tezos/tezos/tree/${source.ref})."
```
gromakovsky, line 13: "based on else"?

rvem: fixed

gromakovsky:
Can you push a "DROP ME BEFORE MERGE" commit to check that releases are formed properly?

rvem:
`release.nix` had some typos, now everything looks fine

gromakovsky:
@balsoft will you have time to review it this week?

release.nix
```diff
@@ -9,7 +9,8 @@ let
   release-notes = writeTextDir "release-notes.md" ''
     Automatic release on ${builtins.substring 0 8 timestamp}
 
-    This release contains assets based on [revision ${source.rev}](https://gitlab.com/tezos/tezos/tree/${source.rev}) of ${source.ref} branch.
+    This release contains assets based on
+      [${source.ref} release](https://gitlab.com/tezos/tezos/tree/${source.ref}).
```
gromakovsky, line 13: Do we really want a linebreak here?

deb.nix
```diff
@@ -6,15 +6,16 @@ pkg:
 let
   project = pkg.meta.name;
   version = meta.version;
-  revision = meta.gitRevision;
+  release = meta.release;
+  epoch = meta.epoch;
```
balsoft, line 10: Why is it not used anywhere?

release.nix
```diff
@@ -24,10 +25,11 @@ let
     name = "tezos-release-no-tarball";
     paths = [ "${bundled.binaries}/bin" LICENSE release-notes ];
   };
-  releaseTarball = runCommand "release-tarball" { }
-    "mkdir $out; tar --owner=serokell:1000 --mode='u+rwX' -czhf $out/release.tar.gz -C ${releaseNoTarball} .";
+  tarballName = "binaries-${builtins.stringReplace ["v"] [""] source.ref}-1.tar.gz";
+  binariesTarball = runCommand "binaries-tarball" { }
+    "mkdir $out; tar --owner=serokell:1000 --mode='u+rwX' -czhf $out/${tarballName} -C ${bundled.binaries}/bin .";
```
balsoft, line 30: What's the reasoning for this change? Don't we want stuff like LICENSE and release notes included in the tarball?

rvem: we want a single tarball with all binaries, see old releases

rvem: Maybe LICENSE is useful, but release notes definitely shouldn't be included

deb.nix
rvem: oh lol, thanks for noticing

gromakovsky:
@rvem and don't forget to make an actual release :wink:

PR #53, opened 2020-05-26
[#9] nix less static building

rvem:
## Description
Problem: We want to have a nix-less way to build static tezos binaries.

Solution: Build them using a custom alpine docker image. This image
includes building a static version for `libusb-1.0` and `libusb-hidapi` libraries
that are required for building static tezos binaries.
In order to build the binaries, one needs to simply run
`./docker-static-build.sh` script.

## Related issue(s)

Resolves part of #9


nix-less/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
```
gromakovsky, line 1:
```suggestion
# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
```

Very minor, but would be nice to update everywhere :)

gromakovsky:
I see some " ---> Using cache" messages. Where do we setup caching?

rvem:
Docker has some image cache, currently, images are pruned on daily basis

PR #54, opened 2020-05-31
Add tls to conduit

### LINKED ISSUE #52: [BUG] No SSL or TLS support compiled into Conduit
# Description

I am using `tezos-client` from the latest release and trying to connect to a node using TLS. I get an error: "No SSL or TLS support compiled into Conduit".

# To Reproduce

1. Use `tezos-client` from the 7.0-1 release.
2. Run any command that tries to connect to a functional remote node that uses TLS, for example:
> `tezos-client -S -A mainnet-tezos.giganode.io -P 443 rpc get /chains/main/blocks/head/protocols`
3. Get the following error:
```
Error:
  Rpc request failed:
     - meth: GET
     - uri: https://mainnet-tezos.giganode.io:443/describe/chains/main/blocks/head/protocols?recurse=no
     - error: Unable to connect to the node: "No SSL or TLS support compiled into Conduit"
```

# Expected behavior

I expect it to work the same way as `./carthagenet.sh client`
```
>  ./carthagenet.sh client -S -A mainnet-tezos.giganode.io -P 443 rpc get /chains/main/blocks/head/protocols
{ "protocol": "PsCARTHAGazKbHtnKfLzQg3kms52kSRpgnDY982a9oYsSXRLQEb",
  "next_protocol": "PsCARTHAGazKbHtnKfLzQg3kms52kSRpgnDY982a9oYsSXRLQEb" }
```

# Environment

- Archlinux
- v7.0-1 release

balsoft:
### Problem

see #52 

TLDR: 

```
$ tezos-client -S -A mainnet-tezos.giganode.io -P 443 rpc get /chains/main/blocks/head/protocols
Error:
  Rpc request failed:
     - meth: GET
     - uri: https://mainnet-tezos.giganode.io:443/describe/chains/main/blocks/head/protocols?recurse=no
     - error: Unable to connect to the node: "No SSL or TLS support compiled into Conduit"
```

### Solution

Build conduit with tls support.

Now:

```
$ tezos-client -S -A mainnet-tezos.giganode.io -P 443 rpc get /chains/main/blocks/head/protocols
{ "protocol": "PsCARTHAGazKbHtnKfLzQg3kms52kSRpgnDY982a9oYsSXRLQEb",
  "next_protocol": "PsCARTHAGazKbHtnKfLzQg3kms52kSRpgnDY982a9oYsSXRLQEb" }
```

## Related issue(s)

Resolves #52

### DIFF
--- a/hacks.nix
+++ b/hacks.nix
@@ -11,7 +11,10 @@ with oself; {
   # FIXME opam-nix needs to do version resolution
   ezjsonm = osuper.ezjsonm.versions."1.1.0";
   ipaddr = osuper.ipaddr.versions."4.0.0";
-  conduit = osuper.conduit.versions."2.1.0";
+  conduit = osuper.conduit.versions."2.1.0".overrideAttrs (oa: rec {
+    buildInputs = [ oself.tls ] ++ oa.buildInputs;
+    propagatedBuildInputs = buildInputs;
+  });
   conduit-lwt-unix = osuper.conduit-lwt-unix.versions."2.0.2";
   cohttp-lwt-unix = osuper.cohttp-lwt-unix.versions."2.4.0";
   cohttp-lwt = osuper.cohttp-lwt.versions."2.4.0";


gromakovsky:
Also I'd like @rvem to take a look as I am barely familiar with Nix and Ocaml worlds.

balsoft:
@gromakovsky if you could come up with a series of "tests" for all the utilities, those tests could be implemented on CI rather trivially.

gromakovsky:
Well, I am just speaking out loud. We can have a trivial test like `tezos-client -S -A ADDR rpc get /chains/main/blocks/head/` where `ADDR` is an address of a node that uses TLS, but we need a node that uses TLS and is always accessible. Options:
1. Use an existing node, not ours, hope it will always be alive. If it dies, we will find another one and update its address in CI config.
2. Deploy our own node with TLS. Doing it in addition to the existing node (without TLS) sounds too costly (but maybe it's doable without overhead). Changing our existing node to use TLS is not costly, but a bit less convenient (one has to pass `-S` or enable TLS in their config, but then disable it for local chain). Our node can have some issues and break our CI when all is good, but we can fix it by repairing our node instead of looking for another one.

Maybe there are better options (like run node locally on CI using `tezos-node` binary built by CI and then try to connect to it using `tezos-client` built by CI).
Now that I wrote it, I suppose this option might be the best one and maybe deserves a dedicated issue.

rvem:
image
oops, missclicked

rvem:
@balsoft please update release [here](https://github.com/serokell/tezos-packaging/blob/master/default.nix#L20)

balsoft:
Oh crap :0

Issue #56, opened 2020-06-12
Ubuntu native packaging


gromakovsky:
# Clarification and motivation

Our packages for Ubuntu currently contain just static binaries. However, `.deb` packages are not allowed to statically link against `libc`. We can distribute these packages ourselves, but we can't put them into main official repositories. However, we probably want to do it one day.

We've made some researched and based on that we are going to dynamically link with natively-installed C libraries, but statically link with all OCaml libraries. We also considered using Snappy, but decided not to do it for now.

# Acceptance criteria

There is an automated way to create `.deb` packages that do not link statically with native libraries.

gromakovsky:
Our .deb and .rpm packages should include systemd units, so that after installing them one would be able to run `tezos-node` and other daemons using `systemctl`.

gromakovsky:
Also there was a request to implement packaging (i. e. `deb` package construction) in nix-less way.

gromakovsky:
What's left:
1. Systemd units.
2. Upload to launchpad.

PR #57, opened 2020-06-15
[#9] docker binaries in releases

### LINKED ISSUE #9: Nix-less packaging

It might be useful for other parties who are not familiar with Nix to be able to set up their own packaging CD. Therefore we need an option of building the client binary without the help of Nix.

Since the packages will have to be built “natively”, preferable we should use distro-native C libraries. Using distro-native OCaml libraries would be great, if possible, but if not, looks like, we’ll have to bundle all transitive OCaml dependencies into our package. (It seems that OCaml links binaries dynamically only with C libraries, and all OCaml libraries are linked statically – if that is correct, then we are in luck.)

# Acceptance criteria

* The repo contains scripts/tools that can be used to build a `tezos-client` binary without Nix.
* There are scripts/tools that can be used to build Ubuntu and Fedora packages for `tezos-client`.
* There is documentation on how to use all of the above so that anyone can setup their own CD system using their favourite CD tool.

rvem:
## Description
Problem: Now we have a way to build binaries using docker, we want to
use these binaries in releases.

Solution: Rework existing release mechanism a bit. Now instead of
building nix derivation, it uses path to the directory with tezos
binaries.

## Related issue(s)

Resolves #9


.buildkite/pipeline.yml
```diff
@@ -32,6 +32,7 @@ steps:
  - commands:
    - cd docker
    - nix run nixpkgs.docker -c ./docker-static-build.sh
+   - nix run nixpkgs.upx -c upx tezos-*
```
gromakovsky, line 35: I heard you had segfaults with this. Are they gone?

default.nix
```diff
@@ -0,0 +1,24 @@
+# SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+{ timestamp ? "19700101", docker-binaries ? null }:
```
gromakovsky, line 5: Are we still using timestamps btw?

.buildkite/pipeline.yml
rvem: it's fine after upx update :yao:

default.nix
rvem: yes, autorelease has `Automatic release on <bla-bla>` default header, but this can be changed

balsoft:
Isn't this technically nixful because we use nix to package binaries?

Otherwise, looks fine

rvem:
> Isn't this technically nixful because we use nix to package binaries?

There is #56 which will update this partially (only for Ubuntu). Apparently, one day we'll update for Fedora as well

default.nix
gromakovsky: I suppose it doesn't really matter what we write there. We can write just "Automatic release". Normally there should be only one automatic release anyway. I think we were including the date there because it was easily available, but if we now propagate it through some scripts just to put into the release tittle I think it's an extra complication.

That being said, it's a bit out of scope of the current issue, so I don't mind if we postpone it.

gromakovsky:
AFAIU there is build process and packaging. The former is nix-less while the latter is not. And the latter seems to be quite trivial, we just write some lines to deb and rpm files. Since packaging for Ubuntu most likely will be changed one day (due to #56) I think there is not much point in updating it before #56 (maybe as part of #56). And then we'll see what to do with Fedora. I suppose the main task for now was changing the **build** process to be nix-less because it's more complicated.

That being said, I still kind dislike the structure of files and folders. Let me copy my comment:
> I think it's fine to still use nix for this purpose if that's more convenient, but these files should be moved outside of `nix/` directory. I think `nix/` directory should be limited to files necessary to build things with `nix`, and files that we use to create packages and release artifacts should be separated. Alternatively, we can still store all nix files in `nix/`, but have `nix/FOLDER_WITH_FILES_FOR_BUILDING` and `nix/FOLDER_WITH_FILES_FOR_RELEASING`.

As I see it, conceptually we have:
1. Files for build process with `docker`.
2. Files for build process with `nix`.
3. Files for making `.deb` and `.rpm` packages.
4. High level files that basically define high level interface of all of this.

(4) can be in the root, that's fine. And I'd prefer to separate first three things. For example:
* `build/[docker,nix]/` and `package/`
* `docker/` and `nix/build/` and `nix/package/`

tl;dr: I don't like that files that we use to build binaries are mixed with files we use to create `.deb` and other packages.

rvem:
@gromakovsky repo structure was updated, now there are dedicated `nix/build` and `nix/package`

default.nix
rvem: I've decided to remove timestamp usage completely

gromakovsky:
Is `ocaml-overlay.nix` needed for packaging? Shouldn't packaging be independent of ocaml? Not sure about other files in `nix/`, but they don't look so suspicious at least.

rvem:
`ocaml-overlay.nix` is useful, when this project is imported to another using nix, so I decided to put it to the `nix/`

gromakovsky:
But it's useful only when you want to **build** something, not when you want to put into a package, right? So if some other projects wants to build Tezos stuff using nix, why can't it use `nix/build`?

rvem:
Well, strictly speaking, `ocaml-overlay.nix` provides nix packages for tezos-binaries, so it's questionable where it should be put :)

rvem:
@gromakovsky I've moved `ocaml-overlay.nix` to `nix/build`

PR #58, opened 2020-06-18
[#55] Extend binaries testing

### LINKED ISSUE #55: Test behavior of binaries

We have some tests checking that our binaries can at least be launched and don't crash immediately: https://github.com/serokell/tezos-packaging/blob/2d04cdadbd8e15b6a79d8a7b2c88acaab660eb74/default.nix#L61-L75

But they don't protect us from bugs like #52. I suggest that we expand our tests a bit.

# Acceptance criteria

1. As part of our tests we already run `tezos-node`. In addition to that, let's also run `tezos-client` and connect to this node, do some trivial request.
2. Let's run node with AND without TLS.
3. For each other executable please think whether there is an easy and reasonable way to sanity check some of its properties that we genuinely expect to always hold. Add tests if you find appropriate ones.

rvem:
## Description
Problem: We want to extend binaries testing to prevent unexpected bug
appearance in the future.

Solution: Use nixos tests framework for testing tezos binaries, port
and extend the existing testing scenario to it, add scenario for testing
daemons and tezos-admin-client.

## Related issue(s)

Resolves #55 


tests/tezos-binaries.nix
```diff
@@ -0,0 +1,139 @@
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: MPL-2.0
+{ path-to-binaries ? null } @ args:
```
balsoft, line 4:
```suggestion
{ path-to-binaries ? (import ../nix { }).binaries } @ args:
```

tests/tezos-binaries.nix
```diff
@@ -0,0 +1,139 @@
...
+  tezos-client = "${path-to-binaries}/tezos-client";
+  tezos-endorser = "${path-to-binaries}/tezos-endorser-006-PsCARTHA";
+  tezos-node = "${path-to-binaries}/tezos-node";
+  tezos-signer = "${path-to-binaries}/tezos-signer";
+in
+{
+  nodes.machine = { ... }: { };
+
+  testScript = ''
+    machine.succeed('export TEZOS_LOG="* -> warning"')
```
gromakovsky, line 19: Is it a good idea to write 100+ lines of python code inside a `.nix` file?

.buildkite/pipeline.yml
```diff
@@ -25,14 +25,15 @@ steps:
...
+   - buildkite-agent artifact download "binaries/bin/*" . --step "build and package"
+   - chmod +x ./binaries/bin/*
+   - nix-build tests/tezos-binaries.nix --no-out-link --arg path-to-binaries ./binaries/bin
    label: test binaries
    branches: "!master"
  - commands:
    - cd docker
    - nix run nixpkgs.docker -c ./docker-static-build.sh
-   - cd .. && ./scripts/test-binaries.sh docker
+   - cd .. && nix-build tests/tezos-binaries.nix --no-out-link --arg path-to-binaries ./docker
```
gromakovsky, line 36: Do we test binaries produced by nix AND binaries produced by docker?

rvem: yes

tests/tezos-binaries.nix
rvem: that's how nixos tests work :shrug:

.buildkite/pipeline.yml
rvem: in different steps

tests/tezos-binaries.nix
rvem: IMO, it's a bit less error-prone in comparison to bash script. Also one may want to add tests that include multiple machines, and it's easy to do that using nixos test engine

gromakovsky: But isn't it possible to put this python code into some `test.py` and somehow import it here? I know that nix is pure and `testScript = readFile(test.py)` probably won't work here directly, but is there no way to make it work? If we put it into `test.py` we will get syntax highlighting and other benefits.

balsoft: `testScript = readFile ./test.py` would indeed work if it wasn't for string interpolation. `${tezos-node}` only works in a string.

Note however that this can be circumvented in various ways.

rvem: that would be nice, but it seems that the test engine does some formatting stuff and cannot parse scripts given as a filepaths instead of plain code

rvem: also, something tricky should be done with binaries paths

balsoft: You can do something like this:
```
testScript = ''
tezos_node = "${tezos-node}"
tezos_client = "${tezos-client}"
<...>
${readFile ./test.py}
''
```

rvem: :pogchamp:

tests/test_script.py
```diff
@@ -0,0 +1,85 @@
+machine.succeed('export TEZOS_LOG="* -> warning"')
```
gromakovsky, line 4: Probably it's a good idea to start this file with a comment stating that it's supposed to be used only from `tests/tezos-binaries.nix`.

.buildkite/pipeline.yml
```diff
@@ -25,15 +25,16 @@ steps:
      - ./rpm-packages/*
    branches: "!master"
  - commands:
-   - nix-build ./nix -A binaries -o binaries
-   - ./scripts/test-binaries.sh binaries/bin
+   - buildkite-agent artifact download "binaries/bin/*" . --step "build and package"
+   - chmod +x ./binaries/bin/*
+   - nix-build tests/tezos-binaries.nix --no-out-link --arg path-to-binaries ./binaries/bin
    label: test binaries
```
gromakovsky, line 31: Can we change it to something like "test nix-built binaries"? And I suppose "build and package" should changed to something like "build with nix and package". And why do we package before building with docker? Or do we build with docker here? Below I see "build using docker".

I think this file requires some refactoring:
1. Labels are a bit outdated.
2. It'd be nice to start each step with `label` as we do in most repos.
3. I'd also add some comments probably, but that depends on how you will update the labels.
4. Time to update the copyright year :wink: 

I don't know if it's a good idea to put all this refactoring into this PR. Feel free to do it fully here or only partially here and then making another PR.

rvem: I did refactoring in this PR in a separate commit

tests/test_script.py
```diff
@@ -1,6 +1,9 @@
 # SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
 #
 # SPDX-License-Identifier: MPL-2.0
+
+# This script is used in tezos-binaries.nix and doesn't supposed to be run
```
gromakovsky, line 5:
```suggestion
# This script is used in tezos-binaries.nix and isn't supposed to be run
```

.buildkite/pipeline.yml
gromakovsky: While we are at it, is `NIX_PATH` used?

.buildkite/pipeline.yml
```diff
@@ -7,40 +7,46 @@ env:
...
-   label: crossref-verify
+ - label: reuse lint
+   command: nix run nixpkgs.reuse -c reuse lint
+ - label: check trailing whitespace
+   command: .buildkite/check-trailing-whitespace.sh
+ - label: crossref-verify
+   command: "nix run -f https://github.com/serokell/crossref-verifier/archive/68a1f9d25b6e7835fea8299b18a3e6c61dbb2a5c.tar.gz -c crossref-verify"
    soft_fail: true
- - commands:
+ - label: build and package via nix
```
gromakovsky, line 17: Can you clarify what's going on in this step? AFAIU our dep and rpm packages contain docker-built binaries, don't they? If they do, shouldn't we build with docker before this step?

rvem: no, they don't, we don't have any docker-based packaging right now

.buildkite/pipeline.yml
rvem: yes, there is plenty of `nix run nixpkgs.*`

.buildkite/pipeline.yml
gromakovsky: Ah, I thought that everything that goes into release is built without nix. And it's actually true, I just forgot that we stopped putting deb and rpm packages into releases.

.buildkite/pipeline.yml
gromakovsky: But we also have `sources.json` that specifies version of `nixpkgs`. Why are we using `NIX_PATH` in addition to that?

gromakovsky:
I would also add some empty lines to `pipeline.yml` to visually separate different parts of it (general checks that we apply to all repos, nix stuff, docker stuff, etc.).

.buildkite/pipeline.yml
rvem: I'd say that it's historically happened

PR #60, opened 2020-07-08
[#56] Nix less ubuntu packaging

rvem:
## Description

This PR adds a way for building native Ubuntu packages (both binary and source) using ubuntu docker image.

TODO:
~* Check source packages on launchpad (for now they were checked only using `pbuilder` which kinda emulates
launchpad builder behaviour).~ [test package](https://launchpad.net/~rvem/+archive/ubuntu/tezos-experiments/+packages)
~* Add information on updating launchpad using source packages.~
~* Prune nix-based ubuntu package creation.~
~* Clean up python script.~

## Related issue(s)

Resolves part of #56 (systemd units will be added later, maybe in another PR)


meta.json
```diff
@@ -0,0 +1,5 @@
+{
+    "release": "1",
+    "epoch": "1",
+    "maintainer": "Serokell <hi@serokell.io>"
```
gromakovsky, line 4: Why did you change "Serokell https://serokell.io <hi@serokell.io>" to "Serokell <hi@serokell.io>"?

docker/package/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+FROM ubuntu:18.04
+RUN apt-get update && apt-get install -y libev-dev libgmp-dev libhidapi-dev m4 perl pkg-config \
+  debhelper dh-make devscripts autotools-dev python3 python3-distutils wget
+RUN apt-get install -y software-properties-common && add-apt-repository ppa:avsm/ppa && apt-get update && apt-get install -y opam
+ENV USER dockerbuilder
+RUN useradd dockerbuilder && mkdir /tezos-packaging
```
gromakovsky, line 10: Why do we need `dockerbuilder` user?

docker/package/package_generator.py
```diff
@@ -0,0 +1,140 @@
...
+    subprocess.run(["dh_make", "-syf" f"../{dir}.tar.gz"], check=True)
+    gen_control_file(package, "debian/control")
+    subprocess.run(["wget", "-q", "-O", "debian/copyright", f"https://gitlab.com/tezos/tezos/-/raw/v{version}/LICENSE"], check=True)
+    subprocess.run("rm debian/*.ex debian/*.EX debian/README*", shell=True, check=True)
+    gen_changelog(package, "bionic", meta["maintainer"], date, "debian/changelog")
+    gen_rules("debian/rules")
+    subprocess.run(["dpkg-buildpackage", "-S" if sys.argv[1] == "source" else "-b", "-us", "-uc"],
+                   check=True)
+    os.chdir("..")
+    subprocess.run(f"rm -rf {dir}", shell=True, check=True)
```
gromakovsky, line 131:
```suggestion
    subprocess.run(f"rm -rf {dir}", shell=True, check=True)
    
```

meta.json
rvem: `dh_make` doesn't like links and likes emails in this format:(

docker/package/package_generator.py
```diff
@@ -0,0 +1,140 @@
...
+        self.desc = desc
+
+date = subprocess.check_output(["date", "-R"]).decode().strip()
+meta = json.load(open("../meta.json", "r"))
+tezos_sources = json.load(open("../nix/nix/sources.json", "r"))["tezos"]
+
+common_deps = ["libev-dev", "libgmp-dev", "libhidapi-dev", "m4", "perl", "pkg-config",
+               "wget", "unzip", "rsync"]
+
+active_protocols = json.load(open("../nix/protocols.json", "r"))["active"]
```
gromakovsky, line 21: I kinda dislike that docker way of doing things uses something from `../nix`. I think `protocols.json` can (should) be put outside of `nix/` since it's used not only by nix. There is also `../nix/nix/sources.json`, but it's more complicated, probably it's fine to use as is.

docker/package/Dockerfile
rvem: `dh_make` needs a dedicated user (not root) to create the package

.buildkite/pipeline.yml
```diff
@@ -45,6 +43,18 @@ steps:
...
    branches: "!master"
 
+ - label: deb source packages via docker
+   commands:
+   - nix run -f. pkgs.docker -c ./docker/docker-ubuntu-packages.sh source
+   artifact_paths:
+     - ./out/*
+ - label: deb binary packages via docker
+   commands:
+   - nix run -f. pkgs.docker -c ./docker/docker-ubuntu-packages.sh binary
```
gromakovsky, line 53: AFAIU most of  `docker/package/package_generator.py` is done regardless of binary/source. Don't we do something twice here?

rvem: that's true, the only difference is `dpkg-buildpackage` flag

rvem: however, the process of fetching and archiving sources is quite fast

docker/package/package_generator.py
rvem: Fair enough

docker/package/package_generator.py
rvem: https://buildkite.com/serokell/tezos-packaging/builds/421#6181921a-a4a4-4d42-8ebe-a96ed5c44fe1
:stare:

gromakovsky: Lol, probably github added spaces automatically :trollface:

docker/package/package_generator.py
```diff
@@ -0,0 +1,141 @@
...
+    gen_makefile(package, f"{dir}/Makefile")
+    subprocess.run(["tar", "-czf", f"{dir}.tar.gz", dir], check=True)
+    os.chdir(dir)
+    subprocess.run(["dh_make", "-syf" f"../{dir}.tar.gz"], check=True)
+    gen_control_file(package, "debian/control")
+    subprocess.run(["wget", "-q", "-O", "debian/copyright", f"https://gitlab.com/tezos/tezos/-/raw/v{version}/LICENSE"], check=True)
+    subprocess.run("rm debian/*.ex debian/*.EX debian/README*", shell=True, check=True)
+    gen_changelog(package, "bionic", meta["maintainer"], date, "debian/changelog")
+    gen_rules("debian/rules")
+    subprocess.run(["dpkg-buildpackage", "-S" if sys.argv[1] == "source" else "-b", "-us", "-uc"],
```
gromakovsky, line 128: Can you define something like `is_source` somewhere above to avoid using `sys.argv[1]` more than once? And I would consider three cases: `source`, `binary` and anything else (we should fail in this case).

nix/proto
```diff
@@ -29,5 +29,5 @@ case $command in
   allow) removeAll $1; add allowed $1;;
   ignore) removeAll $1; add ignored $1;;
   forget) removeAll $1;;
-  *) jq . protocols.json
+  *) jq . ../protocols.json
 esac
```
gromakovsky, line 33: Now that `protocols.json` is used not only in nix-way, I think this file should be moved outside of `nix/` as well. AFAIU it's an essential part of the main workflow and we want the main workflow to be possible w/o `nix`.

AFAIU this script doesn't really need `nix`, but it has some traces of `nix`. It essentially declares its dependencies (`jq` and `moreutils`) using `nix`. AFAIU in nix-less world it should have `#!/usr/bin/env bash` shebang and implicitly assume that `jq` and `moreutils` (specifically `sponge`) are installed. Maybe if we remove that `nix-shell` thing, the nix-way will break, but I hope it can be fixed so that both ways are supported.

Btw, I have found this issue https://github.com/stedolan/jq/issues/105 which is almost 7 years old, it's kinda funny that it has some many reactions and is still open and debatable :)

I would put this `proto` script to `scripts/proto` (since we have the `scripts/` folder already).

P. S. You will need to update `CONTRIBUTING.md`.

.buildkite/pipeline.yml
```diff
@@ -45,6 +43,18 @@ steps:
...
+ - label: deb source packages via docker
+   commands:
+   - nix run -f. pkgs.docker -c ./docker/docker-ubuntu-packages.sh source
+   artifact_paths:
+     - ./out/*
+ # - label: deb binary packages via docker
+ #   commands:
+ #   - nix run -f. pkgs.docker -c ./docker/docker-ubuntu-packages.sh binary
+ #   artifact_paths:
+ #     - ./out/*
```
gromakovsky, line 55: AFAIU we still want to require it to pass before a PR is merged, but we will trigger it manually because it's an expensive action, right? And you will uncomment it once you setup BK this way.

rvem: Yes, that's my plan, I've commented it for now in order to not slow down the CI

docker/package/package_generator.py
```diff
@@ -0,0 +1,150 @@
...
+            "CLI client for interacting with tezos blockchain"),
+    Package("tezos-admin-client",
+            "Administration tool for the node") ,
+    Package("tezos-node",
+            "Entry point for initializing, configuring and running a Tezos node"),
+    Package("tezos-signer",
+            "A client to remotely sign operations or blocks")
+]
+
+daemons = ["baker", "accuser", "endorser"]
```
gromakovsky, line 103: AFAIU node is also a daemon, isn't it? But apparently it doesn't need protocol suffix, unlike other daemons. So wouldn't it be more appropriate to call this list somehow differently? :)

PR #67, opened 2020-07-28
[TM-429] WIP Systemd units and launch scripts

### LINKED ISSUE #56: Ubuntu native packaging

### LINKED ISSUE #68: Update protocol versions in README

Instructions in README have commands like `sudo dnf install tezos-baker-005-PsBabyM1`. 005 protocol is old.

# Acceptance criteria

Instructions involving Tezos protocol use the latest protocol.

mkaito:
## Description

Draft. Untested. Pushing early to review as we go.

The aim of this PR is to add systemd service units for the Tezos node, baker, and accuser, along with any scripts necessary to get them up.

The shell script code has been copied from upstream: https://gitlab.com/tezos/tezos/-/blob/v7.2/scripts/docker/entrypoint.inc.sh

Modified to baseless assumptions for correct paths on Ubuntu.

There is a related PR to tezos-infra to spin up a Ubuntu 20.04 machine to test these units.

## Related issue(s)

https://github.com/serokell/tezos-infra/pull/26

https://issues.serokell.io/issue/TM-429

Resolves #68
Resolves #56 

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -1,5 +1,5 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 
-Files: .github/* nix/* protocols.json meta.json
+Files: .github/* nix/nix/* protocols.json meta.json
 Copyright: TQ Tezos <https://tqtezos.com>
 License: MPL-2.0
--- a/README.md
+++ b/README.md
@@ -36,6 +36,32 @@ sudo apt-get install tezos-baker-006-pscartha
 ```
 Once you install such packages the commands `tezos-*` will be available.
 
+### Systemd units for `tezos-node` and daemons
+
+`tezos-node`, `tezos-accuser-006-pscartha`, `tezos-baker-006-pscartha` and
+`tezos-endorser-006-pscartha` packages have systemd files included to the
+package.
+
+Once you've installed the packages with systemd unit, you can run the service
+with the binary from the package using the following command:
+```
+systemctl start <package-name>.service
+```
+To stop the service run:
+```
+systemctl stop <package-name>.service
+```
+
+Each service has configuration file located in `/etc/default`. Default
+configurations can be found [here](docker/package/defaults/).
+
+Files created by the services will be located in `/var/lib/tezos/` by default.
+`tezos-{accuser, baker, endorser}-006-pscartha` services can have configurable
+data directory.
+
+`tezos-{accuser, endorser}` have configurable node address, so that they can be used with both
+remote and local node.
+
 ## Fedora Copr repository with `tezos-*` binaries
 
 If you are using Fedora you can use Copr in order to install `tezos-*`
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ sudo add-apt-repository ppa:serokell/tezos && sudo apt-get update
 sudo apt-get install tezos-client
 # dpkg-source prohibits uppercase in the packages names so the protocol
 # name is in lowercase
-sudo apt-get install tezos-baker-005-psbabym1
+sudo apt-get install tezos-baker-006-pscartha
 ```
 Once you install such packages the commands `tezos-*` will be available.
 
@@ -45,11 +45,11 @@ E.g. in order to install `tezos-client` or `tezos-baker` run the following comma
 # use dnf
 sudo dnf copr enable @Serokell/Tezos
 sudo dnf install tezos-client
-sudo dnf install tezos-baker-005-PsBabyM1
+sudo dnf install tezos-baker-006-PsCARTHA
 
 # or use yum
 sudo yum copr enable @Serokell/Tezos
-sudo yum install tezos-baker-005-PsBabyM1
+sudo yum install tezos-baker-006-PsCARTHA
 ```
 Once you install such packages the commands `tezos-*` will be available.
 
--- a/docker/package/defaults/tezos-baker.conf
+++ b/docker/package/defaults/tezos-baker.conf
@@ -1,0 +1,9 @@
+# vim: ft=sh
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+# shellcheck disable=SC2034
+DATA_DIR="/var/lib/tezos/baker"
+NODE_RPC_PORT="8732"
+BAKER_ACCOUNT=""
--- a/docker/package/scripts/tezos-accuser-start
+++ b/docker/package/scripts/tezos-accuser-start
@@ -1,0 +1,29 @@
+#!/usr/bin/bash
+
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+set -euo pipefail
+
+# $PROTOCOL should be defined in the system unit environment
+accuser="/usr/bin/tezos-accuser-$PROTOCOL"
+
+accuser_dir="$DATA_DIR"
+
+accuser_config="$accuser_dir/config"
+mkdir -p "$accuser_dir"
+
+if [ ! -f "$accuser_config" ]; then
+    "$accuser" --base-dir "$accuser_dir" \
+               --addr "$NODE_HOST" --port "$NODE_RPC_PORT" \
+               config init --output "$accuser_config" >/dev/null 2>&1
+else
+    "$accuser" --base-dir "$accuser_dir" \
+               --addr "$NODE_HOST" --port "$NODE_RPC_PORT" \
+               config update >/dev/null 2>&1
+fi
+
+exec "$accuser" --base-dir "$accuser_dir" \
+     --addr "$NODE_HOST" --port "$NODE_RPC_PORT" \
+     run "$@"
--- a/docker/package/scripts/tezos-baker-start
+++ b/docker/package/scripts/tezos-baker-start
@@ -1,0 +1,38 @@
+#!/usr/bin/bash
+
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+set -euo pipefail
+
+# $PROTOCOL should be defined in the system unit environment
+baker="/usr/bin/tezos-baker-$PROTOCOL"
+
+baker_dir="$DATA_DIR"
+node_data_dir="/var/lib/tezos/node/data"
+
+baker_config="$baker_dir/config"
+mkdir -p "$baker_dir"
+
+if [ ! -f "$baker_config" ]; then
+    "$baker" --base-dir "$baker_dir" \
+             --port "$NODE_RPC_PORT" \
+             config init --output "$baker_config" >/dev/null 2>&1
+else
+    "$baker" --base-dir "$baker_dir" \
+             --port "$NODE_RPC_PORT" \
+             config update >/dev/null 2>&1
+fi
+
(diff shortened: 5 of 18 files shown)


docker/package/systemd/tezos-node.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+After=network.target
+Description=Tezos node
+
+[Service]
+EnvironmentFile=/etc/defaults/tezos
+Environment="TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=Y"
+Environment="TEZOS_LOG=* -> warning"
+ExecStart=/usr/bin/tezos-node-start
+ExecStartPost=/usr/bin/tezos-node-post-start
```
rvem, line 10: what is this script, btw?

docker/package/scripts/tezos-baker-start
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+set -euo pipefail
+
+. "./tezos.inc.sh"
+
+shift
+
+launch_baker "$@"
```
rvem, line 8: looks like baker daemon will require some additional config setup because when you start the baker you should provide baker address alias

rvem: oh, nevermind, this could be done during the service start

docker/package/scripts/tezos.inc.sh
```diff
@@ -0,0 +1,171 @@
+#!/bin/bash
+
+: "${BIN_DIR:="/usr/local/bin"}"
```
rvem, line 3: btw, currently `.deb` packages install binaries to `/usr/bin`

docker/package/scripts/tezos-baker-start
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+set -euo pipefail
+
+. "./tezos.inc.sh"
```
rvem, line 4: also, AFAIU, this won't work cause it will try to use `tezos.inc.sh` from the directory where user did `systemtcl start tezos-bla`

.reuse/dep5
```diff
@@ -1,5 +1,5 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 
-Files: .github/* nix/* protocols.json meta.json
+Files: .github/* nix/* protocols.json meta.json docker/package/defaults/* docker/package/scripts/defaults docker  /package/systemd/*
```
gromakovsky, line 3: Why did you put it here rather than as comments to the files? AFAIK comments are preferable. And AFAIU those files support comments.

README.md
```diff
@@ -36,6 +36,32 @@ sudo apt-get install tezos-baker-006-pscartha
...
+```
+To stop the service run:
+```
+systemctl stop <package-name>.service
+```
+
+Each service has configuration file located in `/etc/defaults`. Default
+configurations can be found [here](docker/package/defaults/).
+
+File created by the services will be located in `/var/lib/tezos/` by default.
```
gromakovsky, line 58: File or files?

.reuse/dep5
rvem: because I was lazy and wanted to make CI happy, will update it to the comments

README.md
```diff
@@ -36,6 +36,32 @@ sudo apt-get install tezos-baker-006-pscartha
 ```
 Once you install such packages the commands `tezos-*` will be available.
 
+### Systemd units for `tezos-node` and daemons
+
+`tezos-node`, `tezos-accuser-006-pscartha`, `tezos-baker-006-pscartha` and
+`tezos-endorser-006-pscartha` packages have systemd files included to the
+package.
+
+Once, you've installed the packages with systemd unit, you can run the service
```
gromakovsky, line 45: Why comma after "Once"? :thinking:

docker/package/systemd/tezos-node.service
```diff
@@ -0,0 +1,11 @@
...
+After=network.target
+Description=Tezos node
+
+[Service]
+EnvironmentFile=/etc/defaults/tezos-node.conf
+Environment="TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=Y"
+Environment="TEZOS_LOG=* -> info"
+ExecStart=/usr/bin/tezos-node-start
+StateDirectory=tezos
+User=tezos
```
gromakovsky, line 11: Apparently some service files have 2 `\n` in the end, some files have 1 and this file has 0.

docker/package/scripts/tezos-accuser-start
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/bash
+set -euo pipefail
+
+accuser="/usr/bin/tezos-accuser-PROTOCOL"
```
gromakovsky, line 4: How does it work? AFAIK in the binary name there is an actual protocol name, not the word PROTOCOL.

docker/package/defaults/tezos-baker.conf
```diff
@@ -0,0 +1,5 @@
+# vim: ft=sh
+
+# shellcheck disable=SC2034
+DATA_DIR="/var/lib/tezos/baker"
+BAKER_ACCOUNT=""
```
gromakovsky, line 5: "No newline at the end of file"

docker/package/scripts/tezos-accuser-start
rvem: yes, I `sed` it in the `package_generator.py`

rvem: [here](https://github.com/serokell/tezos-packaging/pull/67/files#diff-e78056cfeb41547f90c416ed3def3997R148)

gromakovsky: Alright, but I think these scripts would be move self-contained if you added comments saying that the user of these scripts (even though the only user of these scripts is in this repo) is supposed to replace `PROTOCOL` with the actual protocol. In this case the person who sees these scripts for the first time wouldn't be confused about this `PROTOCOL` thing.

Also, why do you use `sed` instead of environmental variables? The latter seems to be more typical, isn't it?

.reuse/dep5
gromakovsky: Btw, just found another bad thing here. `nix/*` is pretty big, it should be `nix/nix/*` here after recent restructuring. Could you please update it (unrelated to this PR, but a really small change)?

docker/package/scripts/tezos-accuser-start
rvem: hmm, apparently, you're right, and using env variable is better

docker/package/systemd/tezos-node.service
```diff
@@ -0,0 +1,15 @@
...
+#
+# SPDX-License-Identifier: MPL-2.0
+
+[Unit]
+After=network.target
+Description=Tezos node
+
+[Service]
+EnvironmentFile=/etc/defaults/tezos-node.conf
+Environment="TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=Y"
```
gromakovsky, line 11: Oh, btw, I think we shouldn't unconditionally disable this disclaimer in our packages. For us this disclaimer is annoying and not interesting, but for random users it can be helpful. There is a reason why it exists and I think these packages are not the right place to say that this disclaimer shouldn't exist.

rvem: hmm, actually, I don't think that this disclaimer is used in these binaries, so it indeed makes sense to remove it

rvem: Also, I've noticed that default verbosity is `notice`, not `info` :pled:

gromakovsky: If default is `notice`, does setting `"TEZOS_LOG=* -> notice"` do anything? If we don't set `TEZOS_LOG`, will it be `notice` anyway? I think it's better to just use default verbosity.

gromakovsky:
Oh, also just in case: I assume you have tested everything :)

rvem:
> I assume you have tested everything

Yes, I have tested new packages with systemd units, also checked that I didn't break other packages

docker/package/systemd/tezos-baker.service
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+[Unit]
+After=network.target
+After=tezos-node.service
+Requires=tezos-node.service
```
mkaito, line 8: This kind of prevents running it against a remote node without running a local one, is this desired?

docker/package/scripts/tezos-node-start
```diff
@@ -0,0 +1,45 @@
...
+                --data-dir "$node_data_dir"
+    fi
+
+    # Launching the node
+
+    exec "$node" run --data-dir "$node_data_dir"
+
+}
+
+launch_node --network "$NETWORK" "$@"
```
mkaito, line 45: Not much of a point in defining a function just to call it once at the bottom.

docker/package/scripts/tezos-endorser-start
```diff
@@ -0,0 +1,36 @@
...
+}
+
+launch_endorser() {
+    configure_endorser
+    exec "$endorser" --base-dir "$endorser_dir" \
+         --addr "$NODE_HOST" --port "$NODE_RPC_PORT" \
+         run "$@"
+}
+
+launch_endorser "$@"
```
mkaito, line 36: Not much of a point to define functions to call them once in the same script (unless you need the scoping)

docker/package/scripts/tezos-accuser-start
```diff
@@ -0,0 +1,38 @@
...
+
+launch_accuser() {
+    configure_accuser
+    exec "$accuser" --base-dir "$accuser_dir" \
+         --addr "$NODE_HOST" --port "$NODE_RPC_PORT" \
+         run "$@"
+}
+
+
+launch_accuser "$@"
```
mkaito, line 38: Not much of a point in defining functions just to call them once in the same scope.

docker/package/systemd/tezos-baker.service
rvem: AFAIU, Baker is supposed to be run only with local node

docker/package/scripts/tezos-node-start
rvem: updated

mkaito:
Once you fix that, feel free to merge.

docker/package/scripts/tezos-accuser-start
```diff
@@ -0,0 +1,29 @@
...
+    "$accuser" --base-dir "$accuser_dir" \
+               --endpoint "$NODE_ENDPOINT" \
+               config init --output "$accuser_config" >/dev/null 2>&1
+else
+    "$accuser" --base-dir "$accuser_dir" \
+               --endpoint "$NODE_ENDPOINT" \
+               config update >/dev/null 2>&1
+fi
+
+exec "$accuser" --base-dir "$accuser_dir" \
```
gromakovsky, line 27: Why is `exec` needed?

rvem: thus, `tezos-accuser` will exchange bash process

PR #69, opened 2020-08-13
[#66] improve tezos updating

### LINKED ISSUE #66: Improve tezos revision updating process

Our "Quick maintenance guide" says:
> Tezos revision is located in the [`sources.json`](../nix/nix/sources.json) file.
  You can either update it manually to newer revision or use `niv` tool.
  In order to do that run `niv update tezos` (this will update revision to latest).
  `niv update tezos -a rev=...` will update wources to some commit.

I'd like to pass just the tag and have `niv` to the trick. So I tried `niv update tezos -a rev=v7.3`. Here is what I got:
```
     "tezos": {
         "ref": "v7.2",
         "repo": "https://gitlab.com/tezos/tezos",
-        "rev": "6b9f3bc3c90d99d4a2bd5cd26e6f3efbcb977090",
+        "rev": "v7.3",
         "type": "git"
     }
 }
```

Apparently I should do `niv update tezos -a rev=ba45727cadd0416936fbd49400bcc986a55064ca -a ref=v7.3`.
Two problems:
1. Specifying `ba45727cadd0416936fbd49400bcc986a55064ca` is redundant (in theory).
2. Our guide doesn't say that both `-a` should be passed.

# Acceptance criteria

1. `s/wources/sources`.
2. This guide should say that the command should be run from `nix/`.
3. Research whether it's possible to specify only the tag and have `niv` figure out the revision. If it's not, open an issue in `niv`.
4. Update the `niv` command (in contribution guidelines) to have both `-a` arguments.

rvem:
## Description
Fixed instructions in `.github/CONTRIBUTING.md`.
Also added automatic tezos version fetching during static docker build.

## Related issue(s)

Resolves #66 

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -4,6 +4,7 @@
 
 env:
   NIX_PATH: nixpkgs=https://github.com/serokell/nixpkgs/archive/master.tar.gz
+  SET_VERSION: "export TEZOS_VERSION=\"$(cat nix/nix/sources.json | jq -r '.tezos.ref')\""
 
 steps:
  - label: reuse lint
@@ -31,6 +32,7 @@ steps:
 
  - label: build via docker
    commands:
+   - eval "$SET_VERSION"
    - cd docker
    - nix run -f.. pkgs.docker -c ./docker-static-build.sh
    - nix run -f.. pkgs.upx -c upx tezos-*
@@ -45,12 +47,14 @@ steps:
 
  - label: test deb source packages via docker
    commands:
+   - eval "$SET_VERSION"
    - nix run -f. pkgs.docker -c ./docker/docker-ubuntu-packages.sh source
    artifact_paths:
      - ./out/*
    branches: "!master"
  - label: test deb binary packages via docker
    commands:
+   - eval "$SET_VERSION"
    # Building all binary packages will take significant amount of time, so we build only one
    # in order to ensure package generation sanity
    - nix run -f. pkgs.docker -c ./docker/docker-ubuntu-packages.sh binary tezos-baker-006-PsCARTHA
--- a/docker/README.md
+++ b/docker/README.md
@@ -11,8 +11,10 @@
 Static binaries building using custom alpine image.
 
 [`docker-static-build.sh`](docker-static-build.sh) will build tezos binaries
-image defined in [Dockerfile](build/Dockerfile). In order to build them, just run the script
+image defined in [Dockerfile](build/Dockerfile). In order to build them you should specify
+`TEZOS_VERSION` env variable and run the script:
 ```
+export TEZOS_VERSION="v7.3"
 ./docker-static-build.sh
 ```
 After that, directory will contain built static binaries.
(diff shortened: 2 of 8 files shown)


gromakovsky:
> Problem: Our contributing doc has updated instuctions for updating

wat

I guess it should be "has outdated instructions"

docker/README.md
```diff
@@ -13,7 +13,7 @@ Static binaries building using custom alpine image.
 [`docker-static-build.sh`](docker-static-build.sh) will build tezos binaries
 image defined in [Dockerfile](build/Dockerfile). In order to build them, just run the script
 ```
-./docker-static-build.sh
+cd .. && ./docker/docker-static-build.sh
```
gromakovsky, line 16: I slightly dislike that now Docker-way is not completely nix-less. `nix/sources.json` is a simple JSON file, but imo it still complicates the process a bit for foreign users. What if we pass `.tezos.ref` as a command line argument or env variable instead? And do `$(cat /sources.json | jq -r '.tezos.ref')` in the CI config?

rvem: Well, docker packaging also uses `nix/sources.json` to determine sources version.
> What if we pass .tezos.ref as a command line argument or env variable instead?

Thus we'll again have two places when the version is defined. I think that we can extract tezos sources json to the root and modify `nix/sources.nix` to use this file, but then it won't be possible to use `niv` for updating

gromakovsky: > Thus we'll again have two places when the version is defined.

No, it will be in only one place: in `nix/sources.nix`. And CI config will be doing something like `$(cat nix/nix/sources.json | jq -r '.tezos.ref')`. So basically we will move the place where we get this version from `docker/docker-static-build.sh` to CI config. This way people who want to run this script on their machine will be able to do it without bothering about `nix/nix/sources.nix`.

rvem: Ah, okay, I didn't get what you've been proposing

.buildkite/pipeline.yml
```diff
@@ -4,7 +4,7 @@
 
 env:
   NIX_PATH: nixpkgs=https://github.com/serokell/nixpkgs/archive/master.tar.gz
```
gromakovsky, line 6: Oh, remind me please, why do we need it when we have `nixpkgs` in `sources.json`? I guess I had this question some time ago already :pled: Maybe it's better to add a comment here, otherwise I will ask again soon :pled:

gromakovsky: (I know it's unrelated to this PR :pled:)

docker/README.md
rvem: Changed to env variable usage

gromakovsky:
> I guess it should be "has outdated instructions"

You still failed to type "instructions" btw :petro:

.buildkite/pipeline.yml
rvem: hmm, I don't remember now, it was required for something earlier :thinking:

rvem: Apparently, it's unused now since we have `pkgs` in `default.nix`

rvem: >  I guess I had this question some time ago already

Probably in another repository :yao:

rvem: aha, I know, its used in `tests/tezos-binaries.nix`

Issue #64, opened 2020-07-27
rpi binaries


chiptus:
First, let me say how I love this project, and how useful it for me.

I'm currently trying to have a tezos-client on one of my RPI (3b+ arch aarch64) and I can't use your binaries as I get `-bash: ./tezos-client: cannot execute binary file: Exec format error`. Is it because of the different architecture?

 Can we add support for this little machines?

Thanks!

gromakovsky:
Yeah, we were discussing it some time ago. It's on our roadmap. Currently it's hard to estimate how hard it might be, but I guess we'll start working on this issue some time in August and then we'll see.

chiptus:
anything I can do to help?

gromakovsky:
@chiptus did you try running binaries from releases or did you try building them yourself (using the `docker/docker-static-build.sh` script)? If it's the former, can you try running `docker/docker-static-build.sh` on your RPI?

chiptus:
I only tried to build from source using opam.

chiptus:
failed: 

```
The command '/bin/sh -c git clone --single-branch --branch v1.0.23 https://github.com/libusb/libusb.git --depth 1' returned a non-zero code: 139
```

rvem:
Which OS are you using on your Raspberry Pi?

chiptus:
respberry os 64bit (beta)

rvem:
Do you remember how you've installed the docker?
Can it be something similar to https://github.com/schollz/find3/issues/1?

chiptus:
I used `curl -sSL https://get.docker.com | sh` as far as I remember

rvem:
I've been playing with qemu and arm alpine docker images in order to build static binaries for `armv7l` or aarch64.
However, I haven't succeeded much in it:(
The issue with `armv7l` caused by the fact that currently ocaml 4.09.1 cannot be easily built on this architecture, thus it's impossible to use opam and its `opam switch create` command in order to build tezos.
The issue with `aarch64` caused by the fact that both `curl` and `wget` fail with SIGSEGV when trying to fetch something from the link that has redirects, thus it's impossible to fetch some sources of the tezos opam dependencies (e.g. astring).
I'll play with `aarch64` a bit more and will try to find the reason of this segfaults.

The other possible direction of the work is to spin up some arm machine on AWS and try to build static tezos binaries there.
When I get some success, I'll send you the binaries for testing.

chiptus:
thx

[...]

PR #70, opened 2020-08-21
[#64] Add aarch64 static binaries building

rvem:
## Description
Problem: We want to support rpi and arm architecture in general.
Unfortunately, it's quite nontrivial to use opam and ocaml on armv7l,
since OCaml 4.09.1 has issues with building on this architecture.

Solution: Add static binaries building for 64bit arm architecture via
qemu + aarch64 docker alpine image.

This PR resolves only part of #64.
Needs to be done:
* Build aarch64 binaries automatically on the remote arm machine in CI.
* Include built binaries to the releases.
* (?) Provide native `.deb` packages with tezos-binaries for RPI.

## Related issue(s)

Resolves part of #64 

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -19,6 +19,27 @@ export TEZOS_VERSION="v7.3"
 ```
 After that, directory will contain built static binaries.
 
+This script can optionally accept argument to define resulting binaries target architecture.
+Currently supported architectures are: `host` and `aarch64`, so that
+one can build native binaries for current architecture or build `aarch64` binaries on
+`x86_64` machine.
+
+### Compiling for `aarch64` on `x86_64` prerequisites
+
+Docker image defined in [`Dockerfile.aarch64`](build/Dockerfile.aarch64) uses qemu for
+compilation on `aarch64`. In particular it uses `qemu-aarch64-static` binary from
+[qemu-user-static repo](https://github.com/multiarch/qemu-user-static/).
+In order to be able to compile tezos using `aarch64` emulator you'll need to run:
+```
+docker run --rm --privileged multiarch/qemu-user-static:register --reset
+```
+This command will register qemu emulator in `binfmt_misc`.
+
+Once this is done you should run the following command to build `aarch64` static tezos binaries:
+```
+./docker-static-build.sh aarch64
+```
+
 ## Ubuntu packages
 
 We provide a way to build both binary and source native Ubuntu packages.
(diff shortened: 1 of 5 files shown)


docker/build/Dockerfile.aarch64
```diff
@@ -0,0 +1,25 @@
...
+COPY --from=binary-fetch qemu-aarch64-static /usr/bin/qemu-aarch64-static
+RUN apk update
+RUN apk --no-cache --virtual add rsync git m4 build-base patch unzip \
+  bubblewrap wget pkgconfig gmp-dev libev-dev hidapi-dev eudev-dev perl opam libusb-dev bash \
+  autoconf automake libtool linux-headers
+RUN git clone --single-branch --branch v1.0.23 https://github.com/libusb/libusb.git --depth 1
+RUN cd libusb && autoreconf -fvi && ./configure && make && make install
+RUN git clone --single-branch --branch hidapi-0.9.0 https://github.com/libusb/hidapi.git --depth 1
+RUN cd hidapi && autoreconf -fvi && ./bootstrap && ./configure && make && make install
+RUN rm -rf libusb hidapi
```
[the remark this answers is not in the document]
rvem: I haven't found a way to share parts of the dockerfiles directly, so I've extracted libraries and tezos building into dedicated shell scripts and used them in the dockerfiles. The amount of duplication slightly decreased.

docker/build/build-tezos.sh
```diff
@@ -0,0 +1,14 @@
+#! /usr/bin/env bash
+# This script builds static tezos binaries. It expects patched required for static building to be
```
gromakovsky, line 7:
```suggestion
# This script builds static tezos binaries. It expects patches required for static building to be
```

docker/README.md
```diff
@@ -19,6 +19,25 @@ export TEZOS_VERSION="v7.3"
 ```
 After that, directory will contain built static binaries.
 
+This script can optionally accept argument to define resulting binaries target architecture.
+Currently supported architectures are: `x86_64` and `aarch64`.
+
+### Cross-compiling for `aarch64` prerequisites
```
gromakovsky, line 25: Apparently we have decided that "Cross-compiling" is not the best term here.

gromakovsky (changes requested):
`x86_64` should be updated to `host` or something like that, because it can in fact be run on any host.

gromakovsky (approved):

PR #71, opened 2020-08-26
Allow all old protocols

rvem:
## Description
Problem: We doesn't compile all protocols for nix binaries, this lead to
a problem when bootstrapping mainnet node, it says that old protocols
are missing during block validation.

Solution: Allow all old protocols.

## Related issue(s)

https://issues.serokell.io/issue/OPS-997


balsoft (approved):
This increases compilation time and size of binaries quite a bit, but probably nobody cares.

PR #72, opened 2020-08-27
Fix auto releases

rvem:
## Description
Problem: `nix-shell -p` requires nixpkgs to be in
`NIX_PATH`, which we removed earlier.

Solution: Use shell.nix file instead of using `-p` option.

## Related issue(s)

Resolves #

### DIFF
--- a/scripts/autorelease.sh
+++ b/scripts/autorelease.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env nix-shell
-#!nix-shell -p gitAndTools.hub git rename -i bash
+#!nix-shell shell.nix -i bash
 # SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
 #
 # SPDX-License-Identifier: MPL-2.0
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -44,21 +44,21 @@ steps:
...
+ #     - ./out/*
+ #   branches: "!master"
+ # - label: test deb binary packages via docker
+ #   commands:
+ #   - eval "$SET_VERSION"
+ #   # Building all binary packages will take significant amount of time, so we build only one
+ #   # in order to ensure package generation sanity
+ #   - nix run -f. pkgs.docker -c ./docker/docker-ubuntu-packages.sh binary tezos-baker-006-PsCARTHA
+ #   - rm -rf out
+ #   branches: "!master"
```
balsoft, line 61: Is this relevant?

rvem: I'll drop commit with these changes, they're here just to speed up the CI, since these steps are quite slow

rvem:
@balsoft ping

balsoft (approved):

PR #73, opened 2020-09-03
[#64] Arm building in CI

### LINKED ISSUE #64: rpi binaries

rvem:
## Description
Problem: We want to have automated way for building static arm binaries.

Solution: Use dedicated buildkite-agent which is run on arm64 machine
for building static arm binaries.

~Currently depends on [OPS-1058](https://issues.serokell.io/issue/OPS-1058), pipeline setup
can change a bit once this issue is resolved.~

## Related issue(s)

Resolves #64

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -37,6 +37,21 @@ steps:
    - nix run -f.. pkgs.upx -c upx tezos-*
    artifact_paths:
      - ./docker/tezos-*
+ # arm builer is an ubuntu machine without nix
+ - label: build arm via docker
+   commands:
+   - eval "$SET_VERSION"
+   - cd docker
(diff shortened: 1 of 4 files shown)


gromakovsky:
I have created #76 and I propose to limit #64 to static binaries building only. So I think this PR resolves the last part of #64 and you can update its description to contain "Resolves #64" (unless I am missing something else that should be done).

zhenyavinogradov (approved):

PR #75, opened 2020-09-08
[WIP] [LIGO-8] Use opam to get tezos

balsoft:
## Description

Currently, we use a git checkout to get the tezos sources for our nix build. Getting it from opam has multiple benefits, especially for LIGO team.

## Related issue(s)

None

### DIFF
--- a/nix/build/hacks.nix
+++ b/nix/build/hacks.nix
@@ -19,6 +19,7 @@ with oself; {
   cohttp-lwt-unix = osuper.cohttp-lwt-unix.versions."2.4.0";
   cohttp-lwt = osuper.cohttp-lwt.versions."2.4.0";
   macaddr = osuper.macaddr.versions."4.0.0";
+  index = osuper.index.versions."1.2.1";
 
   lwt = lwt4;
 
@@ -35,7 +36,14 @@ with oself; {
   bigstring = osuper.bigstring.overrideAttrs (_: { doCheck = false; });
 
   # FIXME vendors/index
-  index = oself.index-super;
+
+  # ???
+
+  blake2 = osuper.blake2.override { dune = oself.dune_2; };
+  hacl = osuper.hacl.override { dune = oself.dune_2; };
+  secp256k1-internal =
+    osuper.secp256k1-internal.override { dune = oself.dune_2; };
+  uecc = osuper.uecc.override { dune = oself.dune_2; };
 
   # FIXME dependencies in tezos-protocol-compiler.opam
   tezos-protocol-compiler = osuper.tezos-protocol-compiler.overrideAttrs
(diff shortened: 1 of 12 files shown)


gromakovsky (changes requested):
1.
> Getting it from opam has multiple benefits, especially for LIGO team.

I'd like to learn more about these benefits and motivation for this PR (why is it needed?).
2. Where is the version of tezos specified now? Is it derived from `opam-repository`? Don't forget to update `CONTRIBUTING.md` and check other documents that may require some updates.
3. How can I build a specific commit of tezos? Is it still possible? Sometimes Tezos developers ask us to test whether a patch fixes an issue we have. It is very convenient to use this repo for such testing, because I don't want to build tezos locally.

gromakovsky:
> 3. How can I build a specific commit of tezos? Is it still possible?

Aha, apparently it's still possible with nix-less way because there we just do `git clone`.

balsoft:
1. The main benefit is consistency. Each tezos release is tested against the OPAM packages from the same snapshot, thus we get behavior more consistent with what you would get from `opam build` if we take Tezos from OPAM and not from git.
2. Yes, the version of tezos is taken from `opam-repository`.
3. There is a way -- insert `(opam-nix.callOPAMPackage (builtins.fetchTarball https://gitlab.com/tezos/tezos/-/archive/<commit>/tezos-<commit>.tar.gz))` in `nix/build/ocaml-overlay.nix` . Or, as you have noticed, you can also just change the `git clone` invocation in the docker script.

zhenyavinogradov (approved):
I don't know why the tests fail, but the code looks good

balsoft:
@rvem could you take a look at the tests?

nix/build/hacks.nix
```diff
@@ -27,6 +35,9 @@ with oself; {
   conf-libev = self.libev;
   conf-hidapi = self.hidapi;
   conf-pkg-config = self.pkg-config;
+  conf-rust = null;
```
rvem, line 38: :joy:

balsoft: Oh right, yeah. Well. Rust doesn't build with Musl, so it will have to be done in a super-hacky way by reimporting nixpkgs.

rvem: Btw, I've recently tried to build new tezos with rust deps with musl inside the alpine image and everything went just fine

balsoft:
```
$ p pkgsMusl.rustc
error: --- ThrownError --- nix-shell
missing bootstrap url for platform x86_64-unknown-linux-musl
(use '--show-trace' to show detailed location information)

```

balsoft: I guess we can use `pkgsCross.musl64` but ocaml doesn't like that

balsoft:
It turns out using OPAM as source of packages is no good because of terrible publishing delays.

PR #78, opened 2020-09-17
Allow using podman instead of docker

rvem:
## Description
Problem: We cannot longer use docker in our CI pipeline.

Solution: Allow using podman instead of docker, use it in CI and update
instructions to mention an ability to use podman.

## Related issue(s)
https://issues.serokell.io/issue/OPS-1063
https://issues.serokell.io/issue/OPS-1064

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -4,6 +4,7 @@
 
 env:
   SET_VERSION: "export TEZOS_VERSION=\"$(cat nix/nix/sources.json | jq -r '.tezos.ref')\""
+  USE_PODMAN: "True"
 
 steps:
  - label: reuse lint
@@ -29,11 +30,11 @@ steps:
    - nix-build tests/tezos-binaries.nix --no-out-link --arg path-to-binaries ./binaries/bin
    branches: "!master"
 
- - label: build via docker
+ - label: build via podman
    commands:
    - eval "$SET_VERSION"
    - cd docker
-   - nix run -f.. pkgs.docker -c ./docker-static-build.sh
+   - ./docker-static-build.sh
    - nix run -f.. pkgs.upx -c upx tezos-*
    artifact_paths:
      - ./docker/tezos-*
@@ -42,7 +43,8 @@ steps:
    commands:
    - eval "$SET_VERSION"
    - cd docker
-   - ./docker-static-build.sh
+   # arm builder doesn't have podman installed
+   - USE_PODMAN="False" ./docker-static-build.sh
    - upx tezos-*
    - >
      for f in ./tezos-*; do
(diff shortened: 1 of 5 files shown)


gromakovsky:
> Problem: We cannot longer use docker in our CI pipeline.

:coolstory:

It doesn't explain anything. Why can't we use docker?
(I sort of can figure it out from the aforementioned issues, but they are private)

rvem:
Alright, I will write more in the `Problem:` section

.buildkite/pipeline.yml
```diff
@@ -62,7 +64,7 @@ steps:
  - label: test deb source packages via docker
```
gromakovsky, line 64: Some "via docker" things do not use docker anymore :wink:

rvem: :yao:

.buildkite/pipeline.yml
```diff
@@ -33,7 +35,7 @@ steps:
    commands:
    - eval "$SET_VERSION"
    - cd docker
-   - nix run -f.. pkgs.docker -c ./docker-static-build.sh
+   - nix run -f.. pkgs.podman pkgs.conmon pkgs.runc pkgs.shadow -c ./docker-static-build.sh
```
zhenyavinogradov, line 38: podman shouldn't need `nix run`, it's available globally on our CI server, and it's better to use the same version as the server provides

rvem: the problem is that `podman` isn't available when running buildkite job for some reason, see:
https://buildkite.com/serokell/tezos-packaging/builds/540#b82f4312-e7dd-4c6f-8c4e-17d80c7087a8/63-68

rvem: maybe buildkite-agent service needs to be restarted :thinking:

zhenyavinogradov: Yeah, try restarting it, it might have the old `PATH` set

rvem: Nope, that doesn't help:(

rvem: Maybe there is no PATH in the service configuration at all? :thinking:

rvem: there is `PATH`, but there is no `podman` in it even after restart

rvem: https://github.com/serokell/serokell-profiles/blob/master/profiles/servers/jupiter/default.nix#L280 :coolstory:

zhenyavinogradov: _Why do computers have to be so hard._ So we need to add podman to the list, enabling a package globally works for gitlab-runner but not for buildkite

rvem: https://github.com/serokell/serokell-profiles/pull/536

rvem: works now :tada:

rvem:
@gromakovsky @zhenyavinogradov please rereview

.buildkite/pipeline.yml
```diff
@@ -4,6 +4,8 @@
 
 env:
   SET_VERSION: "export TEZOS_VERSION=\"$(cat nix/nix/sources.json | jq -r '.tezos.ref')\""
+  USE_PODMAN: "True"
+  STORAGE_DRIVER: "fuse-overlayfs"
```
gromakovsky, line 8: What is it used by?

rvem: Hmm, it's used to force storage driver for podman (the alternative is vfs which consumes a lot of disk space), however, looks like it can be safely removed now

gromakovsky (approved):

zhenyavinogradov (approved):

PR #82, opened 2020-09-24
[#62] Native fedora packaging

### LINKED ISSUE #62: Fedora native packaging

This issue is similar to #56, but for Fedora.

# Acceptance criteria

1. Figure out whether Fedora packages with fully static binaries are upstreamable.
2. If they are not, update our packaging for Fedora to produce upstreamable packages.
3. Add systemd units to Fedora packages (that are supposed to be developed in #56).
4. Fedora packaging should be nix-less. I. e. it should be possible to make Fedora packages without `nix`.

rvem:
## Description
Problem: Currently our fedora packages contain static binary and created
via nix, we want to have proper packages that actually build tezos
during the package build. Also, we want to do it in a nix-less way.

Solution: Extend script for package generation to support fedora
packages generation. Add Dockerfile for Fedora image with all required
dependency and make it possible to build packages inside this image.

~Currently based on #80.~

## Related issue(s)

Resolves #62


docker/package/package_generator.py
```diff
@@ -66,6 +79,75 @@ def gen_control_file(pkg: Package, out):
...
+'''
+        systemd_files =f'''
+%{{_bindir}}/{package.systemd_unit.startup_script}
+%{{_unitdir}}/%{{name}}.service
+%{{_sysconfdir}}/default/%{{name}}
+'''
+        systemd_macros=f'''
+%post
+%systemd_post %{{name}}.service
+useradd tezos || true
```
zhenyavinogradov, line 103: It creates a group with the same name as well, right? And does it create a home directory?

docker/package/Dockerfile-ubuntu
```diff
@@ -22,4 +22,4 @@ COPY docker/package/package_generator.py /tezos-packaging/docker/package_generat
 COPY docker/package/systemd_files_generator.py /tezos-packaging/docker/systemd_files_generator.py
 COPY docker/package/defaults /tezos-packaging/docker/defaults
 COPY docker/package/scripts /tezos-packaging/docker/scripts
-ENTRYPOINT ["./package_generator.py"]
+ENTRYPOINT ["./package_generator.py"]
```
zhenyavinogradov, line 25: Trailing newline missing

zhenyavinogradov (commented):
Was fun to read, but I didn't find much to diss

docker/package/package_generator.py
rvem: > It creates a group with the same name as well, right?
Yes
> And does it create a home directory?
Yes, but unfortunately not in `/var/lib`, will update this place to use `/var/lib` instead of `/home` :(

rvem: Updated

zhenyavinogradov (approved):

PR #83, opened 2020-09-29
Fix default sources in ocaml-overlay.nix

rvem:
## Description
Problem: There is invalid default path in ocaml-overlay.nix.
However, we want to use this overlay in on of our projects.

Solution: Fix default path.

## Related issue(s)

https://issues.serokell.io/issue/OPS-997


gromakovsky (approved):
Why was it not caught by anything? :thinking:

rvem:
Because the default value wasn't used

PR #88, opened 2020-10-06
[#79] update node systemd files

### LINKED ISSUE #79: Make it possible to run multiple nodes systemd services
Currently, it's impossible to run multiple systemd services with `tezos-node`s without manually copying and editing
startup script because the node data directory is hardcoded in this script. However, one may want to run multiple
nodes as a systemd service, e.g. have both carthagenet and delphinet nodes on a single machine.

Also, README should clarify how to run multiple similar services.

# Acceptance criteria
Startup script and other systemd related files for `tezos-node` service are updated.

README has a section that explains how to run multiple similar services. E.g. two bakers that are using
different protocols and different local nodes, or two accusers that use the same protocol, or two nodes
that target different networks.

rvem:
## Description
Make it possible to run multiple similar systemd services with tezos binaries and explain
how to do that.

## Related issue(s)

Resolves #79


docker/package/scripts/tezos-node-start
```diff
@@ -6,41 +6,36 @@
 
 set -euo pipefail
 
-: "${BIN_DIR:="/usr/bin"}"
```
gromakovsky, line 9: AFAIU it means "if `BIN_DIR` is unset, set it to `"/usr/bin"`, otherwise just use the value assigned to it". Correct? If so, does it mean that `BIN_DIR` previously was configurable, but now you removed this feature because it's unused and is not useful?

Also if it's correct, then probably statement "because node data dir is hardcoded in the startup script" is not entirely valid because one could configure it by specifying `DATA_DIR` (which would be suffixed with `/node/data`).

docker/package/defaults/tezos-node.conf
```diff
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: MPL-2.0
 
 # shellcheck disable=SC2034
+DATA_DIR="/var/lib/tezos/node"
```
gromakovsky, line 7: Btw, are these things set only for a local scope and unset when this "local scope" is exited? Imagine I run another application that uses `DATA_DIR` env variable. Is there a chance that it will see `DATA_DIR="/var/lib/tezos/node"`?

docker/package/defaults/tezos-baker.conf
```diff
@@ -5,5 +5,6 @@
 
 # shellcheck disable=SC2034
 DATA_DIR="/var/lib/tezos/baker"
```
gromakovsky, line 7: Btw, is it ok that both baker and node use `DATA_DIR` variable (for different purposes, with different values), are they isolated from each other?

README.md
```diff
@@ -115,6 +121,31 @@ file, they can be found in [`scripts`](./docker/package/scripts) and
...
+Adjust new `/etc/default/tezos-node-1` config file and point to `EnrironentFile` in
+`/lib/systemd/system/tezos-node-1.service` to it.
+After that, you should be able to run new service:
+```
+systemctl start tezos-node-1
+```
+
+Another case for running multiple similar systemd services is when one wants to have
+multiple daemons that target different protocols.
+Since daemons for different protocols are provided in the different packages, they will
```
gromakovsky, line 144: > Since daemons for different protocols are provided in the different packages

Is it true? Don't our latest packages have daemons for two protocols?

gromakovsky (changes requested):

docker/package/defaults/tezos-node.conf
rvem: It's environment only within systemd service executable, e.g. in scope of `tezos-node-start`.

docker/package/scripts/tezos-node-start
rvem: All packages install binaries to `/usr/bin`, so yes, having it configurable looks useless.

> Also if it's correct, then probably statement "because node data dir is hardcoded in the startup script" is not entirely valid

Well, yes, but I think that having `DATA_DIR` explicitly defined in the config file is less confusing

README.md
rvem: no, we have separate packages for `-007-PsDELPH1` and for `-006-PsCARTHA`

docker/package/defaults/tezos-baker.conf
rvem: yes, as I said earlier this environment variables are used only in the scope of running service executables

README.md
gromakovsky: Ah, apparently we have package per executable, not one package for each release (1 for Ubuntu, 1 for Fedora), I forgot about it.

README.md
```diff
@@ -15,8 +15,9 @@ This repo provides various form of distribution for tezos-related executables
 Daemon binaries (as well as packages for them) have suffix that defines their target protocol,
 e.g. `tezos-baker-007-PsDELPH1` can be used only on the chain with 007 protocol.
 
-Other binaries can be used with all protocols. E.g. `tezos-node` can be set up to run different
-networks, more about this you can read in [this article](https://tezos.gitlab.io/user/multinetwork.html).
+Other binaries can be used with all protocols if they're new enough. E.g.
+007 protocol is supported only from `v7.4`. `tezos-node` can be set up to run
+different networks, more about this you can read in [this article](https://tezos.gitlab.io/user/multinetwork.html).
```
gromakovsky, line 20: Is it grammatically correct btw? Isn't "you can read more about this" more :proper:?

gromakovsky (approved):
Nothing to report besides :english:

README.md
```diff
@@ -115,6 +122,31 @@ file, they can be found in [`scripts`](./docker/package/scripts) and
...
 
+### Multiple similar systemd services
+
+It's possible to run multiple same services, e.g. two `tezos-node`s that run different
+networks.
+
+In order to do that you should copy systemd service file and config file, e.g.:
+```
+cp /lib/systemd/system/tezos-node.service /lib/systemd/system/tezos-node-zeronet.service
+cp /etc/default/tezos-node /etc/default/tezos-node-zeronet
```
gromakovsky, line 133: Let's provide `tezos-node-mainnet`, `tezos-node-carthagenet` and `tezos-node-delphinet` services out of the box. Ideally without code duplication. It's a bit inconvenient to do these commands manually.

rvem: Updated. I've changed packages and services logic creation a bit, please check

README.md
```diff
@@ -101,43 +101,40 @@ data directory.
 `tezos-{accuser, endorser}` have configurable node address, so that they can be used with both
 remote and local node.
 
-### Systemd units on other systems
+<!-- TODO: Currently this instructions are broken -->
```
gromakovsky, line 104: What are we going to do with it and when?

rvem: I'll fix it soon

gen_systemd_service_file.py
```diff
@@ -4,13 +4,18 @@
...
+    binary_name = sys.argv[1]
+    for package in packages.packages:
+        if binary_name == package.name:
+            for systemd_unit in package.systemd_units:
+                if systemd_unit.suffix is None:
+                    print_service_file(systemd_unit.service_file,
+                                       f"{package.name}.service")
+                else:
+                    print_service_file(systemd_unit.service_file,
+                                       f"{package.name}-{systemd_unit.suffix}.service")
```
gromakovsky, line 21: Probably it would be a bit shorter if you defined `suffix = None if systemd_unit.suffix is None else f"-{systemd_unit.suffix}"` and then typed `print_service_file` just once.

README.md
rvem: Should be fixed now

README.md
```diff
@@ -101,26 +101,27 @@ data directory.
...
+
+E.g.:
+```
+./gen_systemd_service_file.py tezos-node-mainnet
+# or
+./gen_systemd_service_file.py tezos-baker-007-PsDELPH1
+```
+After that you'll have `.service` file in the current directory.
+
+Apart, from `.service` file you'll need service startup script and default configuration
```
gromakovsky: Comma looks weird.

docker/package/model.py
```diff
@@ -0,0 +1,64 @@
...
+        self.desc = desc
+        self.systemd_units = systemd_units
+        self.target_proto = target_proto
+
+def print_service_file(service_file: ServiceFile, out):
+    after = "".join(map(lambda x: f"After={x}\n", service_file.unit.after))
+    requires = "".join(map(lambda x: f"Requires={x}\n", service_file.unit.requires))
+    environment = "".join(map(lambda x: f"Environment=\"{x}\"\n", service_file.service.environment))
+    environment_file = "" if service_file.service.environment_file is None else f"EnvironmentFile={service_file.service.environment_file}"
+    file_contents = f'''# SPDX-FeleCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
```
gromakovsky, line 48:
```suggestion
    file_contents = f'''# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
```

README.md
```diff
@@ -101,12 +108,14 @@ from scratch.
 
 For this you'll need `.service` file to define systemd service. The easiest way
 to get one is to run [`gen_systemd_service_file.py`](gen_systemd_service_file.py).
-You should specify service name as a first argument and optionally provide target protocol
-as a second argument, e.g.:
+You should specify service name as an argument. Note that there are three
+predefined services for `tezos-node`: `tezos-node-{mainnet, carthagenet, delphinet}`.
```
gromakovsky, line 112: What if I want to run `tezos-node` in another network? E. g. create my own or connect to one created by a friend of mine.

README.md
```diff
@@ -115,6 +124,27 @@ file, they can be found in [`scripts`](./docker/package/scripts) and
 [`defaults`](./docker/package/defaults) folders respectively.
```
gromakovsky, line 124: It's not clear what to do with these files, but probably it's something to be described in `systemd` documentation rather than ours. However, if there is a link to something nice that explains what to do with these files (some cool `systemd` documentation), I think it makes sense to include it here.

README.md
```diff
@@ -115,6 +124,27 @@ file, they can be found in [`scripts`](./docker/package/scripts) and
...
 
 
+### Multiple similar systemd services
+
+It's possible to run multiple same services, e.g. two `tezos-node`s that run different
+networks.
+
+`tezos-node` packages provide three services out of the box:
+`tezos-node-delphinet`, `tezos-node-carthagenet` and `tezos-node-mainnet` that run
+`delphinet`, `carthagenet` and `mainnet` networks respectively.
```
gromakovsky, line 134: Same question as above: do we support custom networks? It's fine to require some manual configuration for custom networks, but it should be described.

README.md
```diff
@@ -115,6 +124,27 @@ file, they can be found in [`scripts`](./docker/package/scripts) and
...
+In order to start it run:
+```
+systemctl start tezos-node-<network>
+```
+
+Another case for running multiple similar systemd services is when one wants to have
+multiple daemons that target different protocols.
+Since daemons for different protocols are provided in the different packages, they will
+have different service files. The only thing that needs to be changed is config file.
+One should provide desired node address, data directory for daemon files and node directory
```
gromakovsky, line 145: Here you mean files like `tezos-accuser.conf`, right? How are people supposed to find these files in their Ubuntu/Fedora systems? If there is some standard location for `systemd` configs and it's some basic knowledge for systemd users, feel free to ignore the last question.

gromakovsky (changes requested):
Also I see this:
> `tezos-node`, `tezos-accuser-006-pscartha`, `tezos-baker-006-pscartha` and
`tezos-endorser-006-pscartha` packages have systemd files included to the
Ubuntu and Fedora packages.

AFAIU now there are also 007 packages with systemd files. So it should be updated I suppose (maybe in a way that wouldn't require updating it when a new protocol comes out).

docker/package/model.py
rvem: kek

README.md
rvem: It's described in the `### Systemd units on Ubuntu or Fedora` paragraph

README.md
rvem: Generated service file describes the required path for the `EnvironmentFile` (in case `EnvironmentFile` is needed).

E.g. generated baker service will have something like
```
...
EnvironmentFile=/etc/default/tezos-baker-006-PsDELPH1
...
```

README.md
rvem: No, we don't support them out of the box. Also, I'd say that currently, it's hard to support them via the existing node service structure, because the `tezos-node-start` startup script does `config init/update` and AFAIU, it's not possible to easily make node use some custom network via this command and the node config file should be edited manually

gromakovsky: AFAIU in order to run node in a network other than these three one has to:
1. Specify a different data dir.
2. Edit the configuration manually.

I see that `docker/package/defaults/tezos-node.conf` has been removed, so now it's harder to do that, but can we at least describe a way to do it manually? Maybe in a separate PR.

README.md
gromakovsky: What is `EnvironmentFile`? Is it the same kind of file as "default configuration file"?

gromakovsky:
```

Conflicting files
docker/package/package_generator.py

```

Looks kinda annoying btw 😉

README.md
rvem: the problem is that `tezos-node-start` will try to override the config, so this startup script will require additional changes as well

README.md
rvem: yes, it's a file where additional environment variables are defined

README.md
gromakovsky: Meh, ok, but it was impossible before this PR as well, right? If yes, I propose to create a new issue about supporting custom networks. Can you please do it?

rvem: See #99

gromakovsky (approved):

PR #91, opened 2020-10-08
[#89][#87] Package along with optional deps

### LINKED ISSUE #87: [BUG] No SSL or TLS support compiled into Conduit 2
# Description

For me tezos-client reports the error in the title when I'm trying to connect to a node via TLS.

# To Reproduce

1. Have the following `~/.tezos-client/config`:
```
{ "base_dir": "/home/martoon/.tezos-client",
  "node_addr": "dalphanet-tezos.giganode.io", "node_port": 443, "tls": true,
  "web_port": 8080, "confirmations": 0 }
```
2. Try some operation with `tezos-client` (e.g. run a contract).
3. Get the output like this

```
Warning:
  Failed to acquire the protocol version from the node
  Rpc request failed:
     - meth: GET
     - uri: https://dalphanet-tezos.giganode.io:443/chains/main/blocks/head/protocols
     - error: Unable to connect to the node: "No SSL or TLS support compiled into Conduit"

Error:
  Unrecognized command.
  Try using the man command to get more information.
```

# Expected behavior

Everything works smoothly.

# Environment

- Ubuntu 18.04
- Installed with apt-get, version `1:0ubuntu7.4-1`

# Additional context

This error is said to be resolved some time ago, a similar problem has been reported in #52 and #54 and fixed.

### LINKED ISSUE #89: [BUG] Binaries from ubuntu and fedora packages doesn't have ledger support
# Description
Opam packages for dedicated tezos binaries doesn't have a `ledgerwallet-tezos` dependency, thus it's not
possible to use binaries built from these packages along with the ledger.

Suddenly, our ubuntu and fedora packaging relies on these opam packages, thus there is no ledger support
for binaries from them.

# To Reproduce
```
tezos-client -A delphi.testnet.tezos.serokell.team -P 8732 man | grep ledger
```
The output is empty

# Expected behavior
Bunch of info about ledger related commands

# Environment

- Ubuntu 20.04
- 7.4-1 binaries from our ppa

rvem:
## Description
Problem: Ledger backend, as well as tls support, is provided
via optional opam dependencies, so they're not included by default.

Solution: Make it possible to explicitly list which optional
dependencies should be bundled along with the desired tezos package.
Add tls and ledger support to all required packages.

TODO: Now we need to bump ubuntu packages versions since the packages sources were updated
(there are new prefetched opam dependencies in them). We should find out what is the best strategy to do that.
See #95, ubuntu packages will be updated only after it's resolved

## Related issue(s)

Resolves #87
Resolves #89 

### DIFF
--- a/docker/package/package_generator.py
+++ b/docker/package/package_generator.py
@@ -8,11 +8,12 @@ from distutils.dir_util import copy_tree
 from systemd_files_generator import gen_service, print_service_file
 
 class Package:
-    def __init__(self, name, desc, systemd_unit=None, target_proto=None):
+    def __init__(self, name, desc, systemd_unit=None, target_proto=None, optional_opam_deps=[]):
         self.name = name
         self.desc = desc
         self.systemd_unit = systemd_unit
         self.target_proto = target_proto
+        self.optional_opam_deps = optional_opam_deps
 
     def get_full_name(self):
         return self.name if self.target_proto is None else f"{self.name}-{self.target_proto}"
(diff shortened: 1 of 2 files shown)


docker/package/package_generator.py
```diff
@@ -193,15 +194,18 @@ def gen_install(pkg: Package, out):
...
 packages = [
     Package("tezos-client",
-            "CLI client for interacting with tezos blockchain"),
+            "CLI client for interacting with tezos blockchain",
+            optional_opam_deps = ["tls", "ledgerwallet-tezos"]),
     Package("tezos-admin-client",
-            "Administration tool for the node") ,
+            "Administration tool for the node",
+            optional_opam_deps = ["tls"]),
     Package("tezos-node",
```
gromakovsky, line 202: Can you run a node accessible via https if you don't explicitly require `tls` for it?

rvem: tezos-node have tls as an explicit dependency

gromakovsky (approved):

PR #92, opened 2020-10-09
Update tezos ref to fix compatibility with newer Nix versions

notgne2:
## Description

Problem: Builds fail in Nix versions 2.4 and above due to a change in default handling of `ref`
Solution: Use a more explicit `ref` for the `tezos` source

## Related issue(s)

None


balsoft (approved):
plz no video standups

PR #96, opened 2020-10-15
[#95] Change ubuntu packages versioning

### LINKED ISSUE #95: Fix ubuntu versioning
Looks like out ubuntu versioning doesn't follow debian versioning policy.
The correct version for the current release will `7.4-0ubuntu1`, see [this](https://www.ducea.com/2006/06/17/ubuntu-package-version-naming-explanation/) for more info about ubuntu packages versioning.

# Acceptance criteria
Ubuntu packages have proper versioning. This will probably require additional epoch bump.
Packages on the launchpad are updated.

rvem:
## Description
Problem: Current ubuntu versioning scheme is incorrect and may
confuse users.

Solution: Use proper versioning scheme:
`<package-name>_<upstream-version>-0ubuntu<release-number>.`

## Related issue(s)

Resolves #95

### DIFF
--- a/docker/package/package_generator.py
+++ b/docker/package/package_generator.py
@@ -56,7 +56,11 @@ active_protocols = json.load(open("../protocols.json", "r"))["active"]
 
 version = os.environ["TEZOS_VERSION"][1:]
 release = f"{meta['release']}"
-package_version = f"{meta['epoch']}:0ubuntu{version}-{release}"
+
+ubuntu_epoch = 2
+fedora_epoch = 1
+
+package_version = f"{ubuntu_epoch}:0ubuntu{version}-{release}"
 
 pwd = os.getcwd()
 home = os.environ["HOME"]
@@ -121,7 +125,7 @@ systemctl enable %{{name}}.service
 Name:    {full_package_name}
 Version: {version}
 Release: {release}
-Epoch: 1
+Epoch: {fedora_epoch}
 Summary: {pkg.desc}
 License: MPL-2.0
 BuildArch: x86_64
--- a/docker/package/package_generator.py
+++ b/docker/package/package_generator.py
@@ -60,8 +60,6 @@ release = f"{meta['release']}"
 ubuntu_epoch = 2
 fedora_epoch = 1
 
-package_version = f"{ubuntu_epoch}:0ubuntu{version}-{release}"
-
 pwd = os.getcwd()
 home = os.environ["HOME"]
 
@@ -171,7 +169,7 @@ install: {pkg.get_full_name()}
         f.write(makefile_contents)
 
 def gen_changelog(pkg: Package, ubuntu_version, maintainer, date, out):
-    changelog_contents = f'''{pkg.get_full_name().lower()} ({package_version}) {ubuntu_version}; urgency=medium
+    changelog_contents = f'''{pkg.get_full_name().lower()} ({ubuntu_epoch}:{version}-0ubuntu{release}) {ubuntu_version}; urgency=medium
 
   * Publish {version}-{release} version of {pkg.name}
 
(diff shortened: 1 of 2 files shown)


docker/package/package_generator.py
```diff
@@ -56,7 +56,11 @@ def __init__(self, name, startup_script, config_file):
 
 version = os.environ["TEZOS_VERSION"][1:]
 release = f"{meta['release']}"
-package_version = f"{meta['epoch']}:0ubuntu{version}-{release}"
+
+ubuntu_epoch = 2
+fedora_epoch = 1
```
gromakovsky, line 61: Is it used?

docker/package/package_generator.py
```diff
@@ -234,7 +232,7 @@ def gen_install(pkg: Package, out):
 for package in packages:
     if package_to_build is None or package.get_full_name() == package_to_build:
         if is_ubuntu:
-            dir = f"{package.get_full_name().lower()}-0ubuntu{version}"
+            dir = f"{package.get_full_name().lower()}-{version}"
```
gromakovsky, line 235: Based on what do we construct this `dir` string?

docker/package/package_generator.py
```diff
@@ -167,7 +169,7 @@ def gen_makefile(pkg: Package, out):
         f.write(makefile_contents)
 
 def gen_changelog(pkg: Package, ubuntu_version, maintainer, date, out):
-    changelog_contents = f'''{pkg.get_full_name().lower()} ({package_version}) {ubuntu_version}; urgency=medium
+    changelog_contents = f'''{pkg.get_full_name().lower()} ({ubuntu_epoch}:{version}-0ubuntu{release}) {ubuntu_version}; urgency=medium
```
gromakovsky, line 172: Btw, does `0` in `0ubuntu` have any meaning?

rvem: https://www.ducea.com/2006/06/17/ubuntu-package-version-naming-explanation/

gromakovsky: > 2006

🧓 

But the article is good, thanks.

docker/package/package_generator.py
rvem: for ubuntu, this dir is source for `orig.tar.gz` archive, so it has the same name for simplicity.

docker/package/package_generator.py
rvem: fixed

docker/package/package_generator.py
gromakovsky: But we changed it, didn't we? If we did, why? I see that we archive it later, so it's supposed to exist already. Where is it created?

rvem: > If we did, why? 

Because we changed the versioning scheme, not it's `<upstream version>-0ubuntu<release number>`, so  
whole `0ubuntu<release number>` is like a release tag now, so the version of the sources is just `<upstream version>`

> Where is it created?

`os.rename(f"{opam_package}-bundle", dir)` a bit lower

gromakovsky: So we choose this name ourselves, right? There is no requirement like "there should be a folder with name <name-format> with all sources", right?

rvem: well, AFAIU, for ubuntu we actually need a folder with such a name to be able to run `dh_make` in it

gromakovsky: AFAIU the directory was something like `tezos-node-0ubuntu7.4`, but now it is `tezos-node-7.4`. So which folder is actually required?

rvem: we switched versioning, previously version without release info was `0ubuntu7.4`, not it is `tezos-node-7.4`, now the latter is required

gromakovsky: Ok

gromakovsky (approved):

PR #105, opened 2020-11-09
[#97] Switch autoreleases from hub to gh

### LINKED ISSUE #97: Use the official GitHub CLI to create releases

We are using the `hub` utility to interact with GitHub from CI, but I heard that there is another tool which is officially developed by GitHub and lets you interact with GitHub. Probably it's quite similar to `hub`. I think we should consider switching to it because "being developed by GitHub" sounds like an essential advantage here. 

# Acceptance criteria

1. Evaluate what should be done in order to switch to this tool.
2. If the change is considered good enough, do the switch.

rvem:
## Description
Problem: There is an official CLI from github, now it can handle
releases.

Solution: Switch to it from hub tool.

## Related issue(s)

Resolves #97 

### DIFF
--- a/scripts/autorelease.sh
+++ b/scripts/autorelease.sh
@@ -27,19 +27,15 @@ shopt -s extglob
 cp -L "$TEMPDIR"/"$project"/!(*.md) "$assets_dir"
 
 # Delete release if it exists
-hub release delete auto-release || true
+gh release delete auto-release --yes || true
 
 # Update the tag
 git fetch # So that the script can be run from an arbitrary checkout
 git tag -f auto-release
 git push --force --tags
 
-# Combine all assets
-assets=()
-for file in $assets_dir/*; do
-    echo $file
-    assets+=("-a" $file)
-done
-
 # Create release
-hub release create "${assets[@]}" -F "$TEMPDIR"/"$project"/release-notes.md --prerelease auto-release
+gh release create -F "$TEMPDIR"/"$project"/release-notes.md --prerelease auto-release
+
+# Upload assets
+gh release upload auto-release "$assets_dir"/*
--- a/scripts/shell.nix
+++ b/scripts/shell.nix
@@ -2,8 +2,28 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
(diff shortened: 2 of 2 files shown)


scripts/autorelease.sh
```diff
@@ -37,9 +37,11 @@ git push --force --tags
 # Combine all assets
 assets=()
 for file in $assets_dir/*; do
-    echo $file
-    assets+=("-a" $file)
+    assets+=($file)
 done
```
zhenyavinogradov, line 41: Looks like you can simply pass `$assets_dir/*` to `gh release upload` call below

scripts/shell.nix
```diff
@@ -2,8 +2,28 @@
...
+          buildGoModule = args: super.buildGoModule (args // rec {
+            version = "1.2.0";
+            vendorSha256 = "0ybbwbw4vdsxdq4w75s1i0dqad844sfgs69b3vlscwfm6g3i9h51";
+            src = self.fetchFromGitHub {
+              owner = "cli";
+              repo = "cli";
+              rev = "v${version}";
+              sha256 = "17hbgi1jh4p07r4p5mr7w7p01i6zzr28mn5i4jaki7p0jwfqbvvi";
+            };
+          });
```
zhenyavinogradov, line 19: Just to note, in general it's better to push/cherry-pick updates to packages like this into our nixpkgs fork. A cherry-picked commit would even be automatically overlayed on our next nixpkgs rebase.

zhenyavinogradov (approved):

scripts/autorelease.sh
rvem: :smart:

scripts/autorelease.sh
```diff
@@ -27,19 +27,15 @@ shopt -s extglob
...
-    echo $file
-    assets+=("-a" $file)
-done
-
 # Create release
-hub release create "${assets[@]}" -F "$TEMPDIR"/"$project"/release-notes.md --prerelease auto-release
+gh release create -F "$TEMPDIR"/"$project"/release-notes.md --prerelease auto-release
+
+# Upload assets
+gh release upload auto-release "$assets_dir/*"
```
zhenyavinogradov, line 41: `*` won't expand inside quotes :pk:

PR #108, opened 2020-11-11
[#99] Support custom networks within the node service

### LINKED ISSUE #99: Support custom networks within the node service
Currently, `tezos-node` service use [`tezos-node-start`](https://github.com/serokell/tezos-packaging/blob/b248887/docker/package/scripts/tezos-node-start) script, which enforces config init/update on each service
restart. It uses the `NETWORK` env variable to define the target node network. 

However, one may want to run his node in some custom network. Such a custom network can be defined via
a custom node config file. So the easiest way to support it is to make it possible to define path to the desired node
config file, instead of creating it in the scope of the start up script.

# Acceptance criteria
Node systemd service can run custom networks as well as existing mainnet and testnets.

FonGadke:
## Description

Problem: We want to run node in some custom network. A custom network
can be defined via a custom node config file.

Solution: Using a variable CUSTOM_NODE_CONFIG, you can specify the path
to the node configuration file.

## Related issue(s)

None

Resolves #99 

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,11 @@ In order to start it run:
 systemctl start tezos-node-<network>
 ```
 
+In addition to node services where the config is predefined to a specific network
+(e.g. `tezos-node-mainnet` or `tezos-node-delphinet`), it's possible to run `tezos-node-custom`
+service and provide a path to the custom node config file via the
+`CUSTOM_NODE_CONFIG` variable in the `tezos-node-custom.service` file.
+
 Another case for running multiple similar systemd services is when one wants to have
 multiple daemons that target different protocols.
 Since daemons for different protocols are provided in the different packages, they will
--- a/docker/package/packages.py
+++ b/docker/package/packages.py
@@ -69,19 +69,27 @@ packages = [
             "A client to decode and encode JSON")
 ]
 
-node_units = []
-for network in networks:
-    env = [f"DATA_DIR=/var/lib/tezos/node-{network}", f"NETWORK={network}", "NODE_RPC_ADDR=127.0.0.1:8732",
-           "CERT_PATH=", "KEY_PATH="]
+
+def mk_node_unit(suffix, env, desc):
     service_file = ServiceFile(Unit(after=["network.target"], requires=[],
-                                    description=f"Tezos node {network}"),
+                                    description=desc),
                                Service(environment=env,
                                        exec_start="/usr/bin/tezos-node-start",
                                        state_directory="tezos", user="tezos"
-                                   ))
-    node_units.append(SystemdUnit(suffix=network,
-                                  service_file=service_file,
-                                  startup_script="tezos-node-start"))
+                               ))
+    return SystemdUnit(suffix=suffix, service_file=service_file, startup_script="tezos-node-start")
+
+
+node_units = []
+common_node_env = ["NODE_RPC_ADDR=127.0.0.1:8732", "CERT_PATH=", "KEY_PATH="]
+for network in networks:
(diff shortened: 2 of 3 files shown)


rvem:
Currently, there are 3 services bundled with the `tezos-node` and it isn't proper to modify them in order to run a daemon that runs on the custom network. I propose adding one more for running custom networks so that `CUSTOM_NODE_CONFIG` will be an environment variable, e.g.
```
...
[Service]
Environment="CUSTOM_NODE_CONFIG=<path to the custom config file>"
...
```

README.md
```diff
@@ -183,6 +183,11 @@ In order to start it run:
 systemctl start tezos-node-<network>
 ```
 
+In addition to node services where config is predefined to a specific network
+(e.g. `tezos-node-mainnet` or `tezos-node-delphinet`), it's possible to run `tezos-node-custom`
+service, where it's possible to provide a path to the custom node config file via
+`CUSTOM_NODE_CONFIG` variable in the `tezos-node-custom.service` file.
```
pasqu4le, line 189:
```suggestion
In addition to node services where the config is predefined to a specific network
(e.g. `tezos-node-mainnet` or `tezos-node-delphinet`), it's possible to run `tezos-node-custom`
service and provide a path to the custom node config file via the
`CUSTOM_NODE_CONFIG` variable in the `tezos-node-custom.service` file.
```

docker/package/packages.py
```diff
@@ -82,6 +82,20 @@
...
+                                  Unit(after=["network.target"], requires=[],
+                                       description=f"Tezos node with custom config"),
+                                  Service(environment=[
+                                      "DATA_DIR=/var/lib/tezos/node-custom",
+                                      "NODE_RPC_ADDR=127.0.0.1:8732",
+                                      "CERT_PATH=", "KEY_PATH=",
+                                      "CUSTOM_NODE_CONFIG="],
+                                          exec_start="/usr/bin/tezos-node-start",
+                                          state_directory="tezos", user="tezos")),
+                              startup_script="tezos-node-start"))
```
pasqu4le, line 97: From what I can tell there's very little difference between this and any other `network` node unit.

I suppose adding `"custom"` to the `networks` list would be a bit improper (and require an ugly check in the loop), but couldn't the common parts be moved to a function instead?

rvem: Moved common parts to a `mk_node_unit` function

docker/package/packages.py
```diff
@@ -69,32 +69,25 @@
...
+    return SystemdUnit(suffix=suffix, service_file=service_file, startup_script="tezos-node-start")
+
+
+node_units = []
+common_node_env = ["NODE_RPC_ADDR=127.0.0.1:8732", "CERT_PATH=", "KEY_PATH="]
+for network in networks:
+    env = [f"DATA_DIR=/var/lib/tezos/node-{network}", f"NETWORK={network}"] + common_node_env
+    node_units.append(mk_node_unit(suffix=network, env=env, desc=f"Tezos node {network}"))
+
+node_units.append(mk_node_unit(suffix="custom", env=["DATA_DIR=/var/lib/tezos/node-custom"] + common_node_env,
```
pasqu4le, line 89: isn't the `env` missing a `"CUSTOM_NODE_CONFIG="`?

pasqu4le (commented):
One more comment/question, the rest LGTM

docker/package/packages.py
rvem: right, lost it for some reason :facepalm:

PR #117, opened 2020-11-19
[#104] Change license to MIT

### LINKED ISSUE #104: Update licensing information
1) Add licensing information to the root of the repository.
2) Consider switching license to MIT.

# Acceptance criteria
There is a license at the root of the repo. License switching is discussed and switched if necessary.

rvem:
## Description

Problem: It's highly likely that parts of our code will be reused in the
main tezos repository.

Solution: Switch license to MIT, thus our licensing and licensing in
tezos repo is the same and it's fine to migrate parts of our code.

## Related issue(s)

Resolves #104


LICENSES/MIT.txt
```diff
@@ -0,0 +1,20 @@
+Open Source License
```
gromakovsky, line 1: I wonder where this "Open Source License" thing comes from (originally). Anyway, I think it's simpler to use this license https://github.com/tqtezos/baseDAO/blob/master/LICENSE, it seems to differ only in the header.

Also it seems more proper to name this file `LicenseRef-MIT-TQ.txt` because we may have another MIT license later (and that's how we name MIT license files usually).

gromakovsky (changes requested):
> There is a license at the root of the repo.

You may have missed this requirement.

gromakovsky (approved):

PR #119, opened 2020-11-23
[#118] Simplify Ubuntu release updating

### LINKED ISSUE #118: Make it possible to use existing '.orig.tar.gz' source archive for ubuntu package generation
Launchpad requires the source archive to be the same for the same package version.
However, packages created for the same tezos version via `opam-bundle` may have different contents and thus different hash,
because at the moment this source archive also has all opam dependencies sources.
Such behaviour is prohibited by the launchpad, so it's quite hard to update package release (without bumping the version). 

So we should make it possible to use existing `.orig.tar.gz` files as a source archive 
(instead of generating one in place via `opam-bundle)`) during the ubuntu source package creation.

# Acceptance criteria
It should be possible to provide the existing `.orig.tar.gz` file while creating ubuntu source packages.

rvem:
## Description
Problem: Launchpad requires the source archive to be the same for
the same package version, but a different release number.

Solution: Allow using existing source archive when building ubuntu
packages.

## Related issue(s)

Resolves #118

### DIFF
--- a/docker/package/package_generator.py
+++ b/docker/package/package_generator.py
@@ -2,29 +2,37 @@
 #
 # SPDX-License-Identifier: LicenseRef-MIT-TQ
 
-import os, shutil, sys, subprocess, json
+import os, shutil, sys, subprocess, json, argparse
 from distutils.dir_util import copy_tree
 
 from .model import Package, print_service_file
 from .packages import packages
 
-if sys.argv[1] == "ubuntu":
+is_ubuntu = None
+is_source = None
+package_to_build = None
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--os", required=True)
+parser.add_argument("--type", help="package type", required=True)
+parser.add_argument("--package", help="specify binary to package")
+args = parser.parse_args()
+
+if args.os == "ubuntu":
     is_ubuntu = True
-elif sys.argv[1] == "fedora":
+elif args.os == "fedora":
     is_ubuntu = False
 else:
-    raise Exception(f"Unexpected package target OS, only 'ubuntu' and 'fedora' are supported.")
+    raise Exception("Unexpected package target OS, only 'ubuntu' and 'fedora' are supported.")
 
-if sys.argv[2] == "source":
+if args.type == "source":
     is_source = True
-elif sys.argv[2] == "binary":
+elif args.type == "binary":
     is_source = False
 else:
     raise Exception("Unexpected package format, only 'source' and 'binary' are supported.")
 
-package_to_build = None
-if len(sys.argv) == 4:
-    package_to_build = sys.argv[3]
+package_to_build = args.package
 
 date = subprocess.check_output(["date", "-R"]).decode().strip()
 meta = json.load(open(f"{os.path.dirname(__file__)}/../../meta.json", "r"))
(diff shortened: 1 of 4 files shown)


docker/package/package_generator.py
```diff
@@ -2,25 +2,57 @@
...
+is_source = None
+package_to_build = None
 
-if sys.argv[2] == "source":
-    is_source = True
-elif sys.argv[2] == "binary":
-    is_source = False
-else:
-    raise Exception("Unexpected package format, only 'source' and 'binary' are supported.")
+options, _ = getopt.getopt(sys.argv[1:], "", ['os=', 'type=', 'package='])
```
gromakovsky: It looks like bash-style of command line options parsing which is hard to read, error-prone and scales poorly. Can we use `argparse` instead?

docker/package/package_generator.py
```diff
@@ -2,25 +2,57 @@
...
+#     is_ubuntu = False
+# else:
+#     raise Exception(f"Unexpected package target OS, only 'ubuntu' and 'fedora' are supported.")
+
+# if sys.argv[2] == "source":
+#     is_source = True
+# elif sys.argv[2] == "binary":
+#     is_source = False
+# else:
+#     raise Exception("Unexpected package format, only 'source' and 'binary' are supported.")
```
gromakovsky: I suppose it should be removed from the history.

gromakovsky:
AFAIU this PR adds a feature that's not used here (by CI), but can be used manually if we make some changes to an existing package and don't want to change its version. Is it correct?

docker/package/package_generator.py
```diff
@@ -269,6 +286,8 @@ def gen_install(pkg: Package, out):
                            check=True)
             os.chdir("..")
         else:
+            if source_archive is not None:
+                raise Exception("Sources archive provision doesn't supported for Fedora packages")
```
gromakovsky, line 290:
```suggestion
                raise Exception("Sources archive provision isn't supported for Fedora packages")
```

docker/package/package_generator.py
rvem: :see_no_evil:

rvem:
> AFAIU this PR adds a feature that's not used here (by CI), but can be used manually if we make some changes to an existing package and don't want to change its version. Is it correct?

Yes

docker/package/package_generator.py
```diff
@@ -2,29 +2,34 @@
...
+parser.add_argument("--os", default="ubuntu")
+parser.add_argument("--type", help="package type")
+parser.add_argument("--package", help="specify binary to package")
+parser.add_argument("--sources", help="specify source archive for single ubuntu package")
+args = parser.parse_args()
 
-package_to_build = None
-if len(sys.argv) == 4:
-    package_to_build = sys.argv[3]
+is_ubuntu = args.os if args.os is None else args.os == "ubuntu"
```
gromakovsky, line 23: AFAIU `is None` case is impossible because you provide a `default` value. You can also set `required=True` to be even more sure. And please set it for `--type` if it's required (instead of checking manually).

rvem: ah, crap, I forgot to remove the default value for `--os`:(

docker/package/package_generator.py
rvem: Will remove during the final rebase before merging

gromakovsky (approved):

PR #121, opened 2020-11-30
[#114] Sign and hash artifacts

### LINKED ISSUE #114: Sign artifacts and provide hash sums

People want to be more confident that artifacts they use can be trusted. E. g. if I obtain our static `tezos-client` somehow, I want to be sure it was provided by Serokell and is genuine as long as I trust Serokell (or, if we move packaging to `tezos/tezos`, trust Tezos developers).

There are at least two natural ways to do that:
1. Sign each artifact using our key (e. g. we can generate a fresh key specifically for this repo), provide our public key somewhere, provide signatures along with artifacts.
2. Provide hash sum (e. g. sha256) for each artifact.

As an example, I see that `stack` releases include `.asc` and `.sha256` files for each artifact: https://github.com/commercialhaskell/stack/releases/tag/v2.5.1

For `.deb` and `.rpm` packages it might be (and actually is quite likely) possible to include signatures into packages themselves. We should explore this opportunity.

# Acceptance criteria

1. We provide at least one way to (quickly and easily) verify our artifacts.
2. For packages we use some standard approach if it exists.

rvem:
## Description
We want to provide some way for people to check whether binaries/packages they're using are genuine
and created by Serokell.

This PR adds sha256 sum calculation along with PGP signing for all autorelease artifacts.

Also, this PR adds instruction about how native packages should be signed before publishing.

## Related issue(s)

Resolves #114

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -12,6 +12,9 @@ let
     Binaries that target arm64 architecture has `-arm64` suffix in the name.
     Other binaries target x86_64.
 
+    Release artifacts are signed with the following key: 0x7EAF9B150ACE940CF8C008A0BF847A85AC7BF43E.
+    You can check it on http://keys.gnupg.net/.
+
     Descriptions for binaries included in this release:
     ${builtins.concatStringsSep "\n"
     (map ({ name, description, ... }: "- `${name}`: ${description}")
--- a/scripts/autorelease.sh
+++ b/scripts/autorelease.sh
@@ -22,12 +22,15 @@ assets_dir=$TEMPDIR/assets
 nix-build -A release -o "$TEMPDIR"/"$project" --arg timestamp "$(date +\"%Y%m%d%H%M\")" \
           --arg docker-binaries ./binaries/docker --arg docker-arm-binaries ./arm-binaries/docker
 mkdir -p "$assets_dir"
-for asset in "$assets_dir"/*; do
-    sha256sum "$asset" > "$asset.sha256"
-done
 # Move archive with binaries and tezos license to assets
 shopt -s extglob
 cp -L "$TEMPDIR"/"$project"/!(*.md) "$assets_dir"
+# Iterate over assets, calculate sha256sum and sign them in order
+# to include this additional info to the release assets as well
+for asset in $assets_dir/*; do
+    sha256sum "$asset" | sed 's/ .*/ /' > "$asset.sha256"
+    gpg --armor --detach-sign "$asset"
+done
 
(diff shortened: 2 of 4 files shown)


release.nix
```diff
@@ -12,6 +12,8 @@ let
     Binaries that target arm64 architecture has `-arm64` suffix in the name.
     Other binaries target x86_64.
 
+    Release artifacts are signed with the following key: 0x7EAF9B150ACE940CF8C008A0BF847A85AC7BF43E
```
gromakovsky, line 15: Perhaps we could also suggest a keyserver from which this key can be fetched.

scripts/autorelease.sh
```diff
@@ -22,12 +22,13 @@ assets_dir=$TEMPDIR/assets
...
           --arg docker-binaries ./binaries/docker --arg docker-arm-binaries ./arm-binaries/docker
 mkdir -p "$assets_dir"
-for asset in "$assets_dir"/*; do
-    sha256sum "$asset" > "$asset.sha256"
-done
 # Move archive with binaries and tezos license to assets
 shopt -s extglob
 cp -L "$TEMPDIR"/"$project"/!(*.md) "$assets_dir"
+for asset in $assets_dir/*; do
+    sha256sum "$asset" | sed 's/ .*/ /' > "$asset.sha256"
```
gromakovsky, line 29: What is this `sed` for?

scripts/autorelease.sh
```diff
@@ -22,12 +22,13 @@ assets_dir=$TEMPDIR/assets
 nix-build -A release -o "$TEMPDIR"/"$project" --arg timestamp "$(date +\"%Y%m%d%H%M\")" \
           --arg docker-binaries ./binaries/docker --arg docker-arm-binaries ./arm-binaries/docker
 mkdir -p "$assets_dir"
-for asset in "$assets_dir"/*; do
-    sha256sum "$asset" > "$asset.sha256"
-done
 # Move archive with binaries and tezos license to assets
 shopt -s extglob
 cp -L "$TEMPDIR"/"$project"/!(*.md) "$assets_dir"
+for asset in $assets_dir/*; do
```
gromakovsky, line 28: Looks like most of the steps here are commented, so I'd leave a comment above as well.

scripts/autorelease.sh
rvem: strip `/tmp/bla-bla` suffix from the hash message

gromakovsky (approved):

PR #127, opened 2021-01-14
[#103] macOS packaging

### LINKED ISSUE #103: Build binaries for macOS
There is a rumor that now we have a mac builder. We should use it in order to build Tezos binaries for macOS.

In addition, we can have brew formulas for them.

# Acceptance criteria
CI builds binaries that target macOS, these binaries are included in the releases.

We provide brew formulas for Tezos binaries

rvem:
## Description

Roadmap:
* ~brew formulas for Tezos binaries~
* ~CI sanity check for formulas~
* ~automate brew bottle creation~ ~and incorporate them into the releases~
* ~update README and other usage instructions~

Out of scope of this issue:
* include systemd stuff to the brew packages, see #130 

Testing locally on macOS machine:
```
brew tap serokell/tezos-packaging https://github.com/serokell/tezos-packaging.git
cd /usr/local/Homebrew/Library/Taps/serokell/homebrew-tezos-packaging && git checkout rvem/#103-tezos-brew-formulae
brew install tezos-client # or any other binary
```
Note, that there are no bottles yet, so you'll have to build a package from scratch.

## Related issue(s)

Resolves #103

### DIFF
--- a/Formula/tezos-codec.rb
+++ b/Formula/tezos-codec.rb
@@ -8,6 +8,12 @@ class TezosCodec < Tezos
   init
   desc "A client to decode and encode JSON"
 
+  bottle do
+    root_url "https://github.com/serokell/tezos-packaging/releases/download/v8.1-1/"
+    cellar :any
+    sha256 "3be0ac706f19bd6a673dd64c5d20b7973b353d80273dc56f3905cb39e6acc03d" => :mojave
+  end
+
   def install
     make_deps
     install_template "src/bin_codec/codec.exe",
--- a/Formula/tezos-codec.rb
+++ b/Formula/tezos-codec.rb
@@ -1,0 +1,17 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+require File.join(File.dirname(__FILE__), "tezos")
+
+class TezosCodec < Tezos
+  init
+  desc "A client to decode and encode JSON"
+
+  def install
+    make_deps
+    install_template "src/bin_codec/codec.exe",
+                     "_build/default/src/bin_codec/codec.exe",
+                     "tezos-codec"
+  end
+end
--- a/Formula/tezos-codec.rb
+++ b/Formula/tezos-codec.rb
@@ -9,9 +9,10 @@ class TezosCodec < Tezos
   desc "A client to decode and encode JSON"
 
   bottle do
-    root_url "https://github.com/serokell/tezos-packaging/releases/download/v8.1-1/"
+    root_url "https://github.com/serokell/tezos-packaging/releases/download/#{TezosCodec.version}/"
     cellar :any
     sha256 "3be0ac706f19bd6a673dd64c5d20b7973b353d80273dc56f3905cb39e6acc03d" => :mojave
+    sha256 "e738541a16f804f74c4e35ef1a03f6af6fa7a2d42c2adcd7191cde63bc8b1d17" => :catalina
   end
 
   def install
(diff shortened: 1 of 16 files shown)


gromakovsky:
I didn't review it yet, but here are some notes about what we want:
1. There should be one formula per-binary, e. g. it should be possible to install only `tezos-client`.
2. However, maybe some binaries can be grouped into one formula if they are always used together.
3. If "virtual" formulas (which don't contain anything other than dependencies) are possible, it makes sense to create one or more (e. g. `tezos` formula).

rvem:
> 3\. If "virtual" formulas (which don't contain anything other than dependencies) are possible, it makes sense to create one or more (e. g. `tezos` formula).

AFAIU, the formula output should be either binaries or libraries, so looks like creating such "virtual" is impossible (however, I can be wrong, I'll research this topic a bit). One simple way that comes to my mind is to have `tezos` package with all binaries and make all other packages depend on it. However, this will consume a lot of additional disk space in the case when user wants only one binary (e.g. only `tezos-client`).

gpevnev (approved):

Formula/tezos-codec.rb
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+require File.join(File.dirname(__FILE__), "tezos")
+
+class TezosCodec < Tezos
+  init
+  desc "A client to decode and encode JSON"
+
+  bottle do
+    root_url "https://github.com/serokell/tezos-packaging/releases/download/v8.1-1/"
```
gpevnev, line 12:
```suggestion
    root_url "https://github.com/serokell/tezos-packaging/releases/download/#{version}/"
```

gpevnev: Just in case

gpevnev: http://ruby-for-beginners.rubymonstas.org/bonus/string_interpolation.html

rvem: :pogchamp:

gpevnev: Or even fugure out some way to abstract this in tezos client if you really want

rvem: please, no, enough ruby for today :joy:

Formula/tezos.rb
```diff
@@ -0,0 +1,46 @@
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
```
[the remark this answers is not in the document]
rvem: :stare:

README.md
```diff
@@ -80,6 +80,28 @@ chmod +x tezos-client
...
+It's possible to provide prebuilt macOS packages for brew called bottles. They're supposed
+to be built before making the new release and included to it. In order to build all bottles run
+`build-bottles.sh` script:
+```
+./scripts/build-bottles.sh
+```
+
+Note that this might take a while, because builds don't share common parts and for each binary
+dependencies are compiled from scratch. Once the bottles are built, the corresponding sections in the
+formulas should be updated. Also, bottles should be downloaded to the release artifacts.
```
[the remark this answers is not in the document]
rvem: :facepalm:

README.md
```diff
@@ -80,6 +80,28 @@ chmod +x tezos-client
 
 Run `./tezos-client` or add it to your PATH to be able to run it anywhere.
 
+## Brew tap for macOS
```
gromakovsky, line 83: Maybe let's link here somewhere around the top of README? Or maybe some kind of a ToC. It's currently hard to understand whether MacOS is supported.

rvem: Yes, I agree, it's time to add ToC to the README

gromakovsky (approved):
Update copyright years if you want, that's optional.

Issue #123, opened 2020-12-07
version 7.5 is still not in ppa


chiptus:
Running `sudo apt update && sudo apt upgrade` doesn't upgrade tezos-client. The version that is in the ppa is still 7.4:0

rvem:
Unfortunately, our ubuntu packaging is based on the tezos opam packages:(
So we blocked by the https://gitlab.com/tezos/tezos/-/issues/1021

rvem:
@chiptus, sorry for such a long response. v8.1 was published to the launchpad, should become available soon (there are still some builds running, everything should be available within 20-25 minutes, I think)

chiptus:
thanks! and thanks for letting me know!

>

rvem:
Let me know if this issue can be closed now

chiptus:
IMO yes, closing

PR #129, opened 2021-01-20
v8.1 native packages

rvem:
## Description
It's time to resurrect native packaging because opam repository was
finally updated. Luckily, the existing packaging approach works nearly out
of the box, only the dependency set should be updated a bit.

## Related issue(s)

Relates #123 

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -86,8 +86,9 @@ steps:
    # in order to ensure package generation sanity
    - ./docker/docker-tezos-packages.sh --os ubuntu --type binary --package tezos-baker-007-PsDELPH1
    - rm -rf out
-   branches: "!master"
-   timeout_in_minutes: 30
+   # It takes much time to build binary package, so we do it only on master
+   branches: "master"
+   timeout_in_minutes: 90
    agents:
      queue: "docker"
  - label: test rpm source packages via docker
@@ -107,8 +108,9 @@ steps:
    # in order to ensure package generation sanity
    - ./docker/docker-tezos-packages.sh --os fedora --type binary --package tezos-baker-007-PsDELPH1
    - rm -rf out
-   branches: "!master"
-   timeout_in_minutes: 30
+   # It takes much time to build binary package, so we do it only on master
+   branches: "master"
+   timeout_in_minutes: 90
    agents:
      queue: "docker"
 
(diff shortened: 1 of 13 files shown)


docker/package/packages.py
```diff
@@ -98,7 +98,8 @@
 
 default_testnets = {
     "006-PsCARTHA": "carthagenet",
```
gromakovsky, line 100: What about dropping it btw?

.buildkite/pipeline.yml
```diff
@@ -100,17 +89,6 @@ steps:
    timeout_in_minutes: 30
    agents:
      queue: "docker"
- - label: test rpm binary packages via docker
```
gromakovsky: Maybe instead of running them in PRs let's at least run them in master or on schedule? Master builds are rare and can be cancelled, and if they fail we will likely notice that.

docker/package/packages.py
```diff
@@ -71,7 +71,7 @@
 
 node_units = []
 for network in networks:
-    env = [f"DATA_DIR=/var/lib/tezos/node-{network}", f"NETWORK={network}", "NODE_RPC_PORT=8732",
+    env = [f"DATA_DIR=/var/lib/tezos/node-{network}", f"NETWORK={network}", "NODE_RPC_PORT=:8732",
```
gromakovsky, line 74: Is `NODE_RPC_PORT` still used? Also did you grep `NODE_RPC_PORT` and `NODE_HOST` to make sure they are no longer mentioned in all places where should stop being mentioned?

gromakovsky (changes requested):
I remember read that now node identity is generated automatically, so our systemd units can be simplified a bit. Is it done in this PR? If not, can it be done in this PR?

docker/package/packages.py
rvem: This is actually a part of `tezos-node` systemd service, where the node rpc port should be specified a bit different via `--rpc-addr=ADDR:PORT` option. However, it's indeed weird that only port can be specified

> Also did you grep NODE_RPC_PORT and NODE_HOST to make sure they are no longer mentioned in all places where should stop being mentioned?

Yes

rvem:
> If not, can it be done in this PR?

Yes, updated
`tezos-node-start` script

gromakovsky (approved):

PR #131, opened 2021-01-25
[#125] 'Baking with tezos-packaging on Ubuntu' article

### LINKED ISSUE #125: Write HOWTO on setting up baking on Ubuntu using packages produced by this project

Tezos aims to enable wide participation in the network operation and governance. This project enables that by taking care of some of the hardest aspects of deploying a working Tezos installation. Clear documentation, a step-by-step guide, on how to use these packages to produce a full, maintainable baking instance of Tezos would further advance the goals of wide adoption and participation. I suggest to start with a guide targeting Ubuntu specifically because Ubuntu is one of the popular platforms among Tezos bakers and packages are already available and contain everything necessary.

# Acceptance criteria

After following the step-by-step guide new user ends up with functioning Tezos installation that includes a node fully synced with the chosen network and a baker and endorser that bake and endorse blocks using a private key from either a file or a Ledger Nano device.

rvem:
## Related issue(s)

Resolves #125

### DIFF
--- a/docs/baking.md
+++ b/docs/baking.md
@@ -1,0 +1,223 @@
+<!--
+   - SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+   -
+   - SPDX-License-Identifier: LicenseRef-MIT-TQ
+   -->
+# Baking with tezos-packaging on Ubuntu
+
+Tezos-packaging provides an easy way to install and set up the infrastructure for
+interacting with the Tezos blockchain.
+
+This article provides a step-by-step guide for setting up a baking instance for Tezos on Ubuntu.
+
+## Prerequisites
+
+### Installing required packages
+
+In order to run a baking instance, you'll need the following Tezos binaries:
+`tezos-client`, `tezos-node`, `tezos-baker-<proto>`, `tezos-endorser-<proto>`.
+
+The currently used proto is `008-PtEdo2Zk` (used on `mainnet` and `edo2net`).
+Also, note that the corresponding Ubuntu packages have protocol
+suffix in lowercase, e.g. the list of available baker packages can be found
+[here](https://launchpad.net/~serokell/+archive/ubuntu/tezos/+packages?field.name_filter=tezos-baker&field.status_filter=published).
+
+To install them, run the following commands:
+```
+# Add PPA with Tezos binaries
+sudo add-apt-repository ppa:serokell/tezos
+sudo apt-get update
+# Install packages
+sudo apt-get install tezos-client tezos-node tezos-baker-<proto> tezos-endorser-<proto>
+```
+
+Packages for `tezos-node`, `tezos-baker-<proto>` and `tezos-endorser-<proto>` provide
+systemd units for running the corresponding binaries in the background.
+
+## Setting up the tezos-node
+
+The easiest way to set up a node running on the `mainnet` or one of the
+testnets (e.g. `edo2net`) is to use one of the predefined
+`tezos-node-<network>.services` systemd services provided in the `tezos-node`
+package.
+
+However, by default, these services will start to bootstrap the node from scratch,
+which will take a significant amount of time.
+In order to avoid this, we suggest bootstrapping from a snapshot instead.
+
+### Setting up node service
+
+`tezos-node-<network>.service` has `/var/lib/tezos/node-<network>` as a data directory
+and `http://localhost:8732` as its RPC address by default.
+
+In case you want to use a different data directory or RPC address,
+you should update the service configuration. To edit the service configuration, run:
+```
+sudo systemctl edit --full tezos-node-<network>.service
+```
+
+### Bootstrapping the node
+
+In order to run a baker locally, you'll need a fully-synced local `tezos-node`.
+
+The fastest way to bootstrap the node is to import a snapshot.
+Snapshots can be downloaded from the following websites:
+* [Tulip Snapshots](https://snapshots.tulip.tools/#/)
+* [Tezos Giganode Snapshots](https://snapshots-tezos.giganode.io/)
+* [XTZ-Shots](https://xtz-shots.io/)
+
+Download the snapshot for the desired network. We recommend to use rolling snapshots. This is
+the smallest and the fastest mode that is sufficient for baking (you can read more about other
+`tezos-node` history modes [here](https://tezos.gitlab.io/user/history_modes.html#history-modes)).
+
+All commands within the service are run under the `tezos` user.
+
+In order to import the snapshot, run the following command:
+```
+sudo -u tezos tezos-node snapshot import --data-dir /var/lib/tezos/node-<network>/ <path to the snapshot file>
+```
+
+#### Troubleshooting node snapshot import
+
+Snapshot import requires that the `tezos-node` data directory doesn't contain the `context` and `store` folders.
+Remove them if you're getting an error about their presence.
+
+In case you're getting an error similar to:
+```
+tezos-node: Error:
+              Invalid block BLjutMj47caB
+                Failed to validate the economic-protocol content of the block: Error:
+                                                                                Invalid signature for block BLjutMj47caB. Expected: tz1Na5QB98cDA.
+```
+
+you should init/update config in the data directory to match the desired network
+before importing the snapshot. To do that, run one of the following commands
+(depending on whether the config in the given data directory was initialized previously):
+```
+sudo -u tezos tezos-node config init --data-dir /var/lib/tezos/node-<network> --network <network>
+#or
+sudo -u tezos tezos-node config update --data-dir /var/lib/tezos/node-<network> --network <network>
+```
+
+### Starting the node
+
+After the snapshot import, you can finally start the node by running:
+```
+sudo systemctl start tezos-node-<network>.service
+```
+
+Note that even after the snapshot import the node can still be out of sync. It may require
+some additional time to completely bootstrap. In order to check whether the node is bootstrapped,
+you can use `tezos-client`:
+```
+sudo -u tezos tezos-client bootstrapped
+```
+
+To stop the node, run:
+```
+sudo systemctl stop tezos-node-<network>.service
+```
+
+If you want the service to start automatically at boot, use `enable`:
+```
+sudo systemctl enable tezos-node-edo2net.service
+```
+and to disable it from doing so, instead use:
+```
+sudo systemctl disable tezos-node-edo2net.service
+```
+
+You can check node logs via `journalctl`:
+```
+journalctl -u tezos-node-<network>.service
+```
+
+## Setting up baker and endorser daemons
+
+### Setting up daemon data directories
+
+Data directories for baker and endorser daemons are defined in the
+`/etc/default/tezos-baker-<proto>` and `/etc/default/tezos-endorser-<proto>`.
+By default, both these daemons have `/var/lib/tezos/.tezos-client` set as a `DATA_DIR`.
+
+Additionally, you need to specify `NODE_DATA_DIR` in the `/etc/default/tezos-baker-<proto>`
+to point at the desired node data directory, e.g. `/var/lib/tezos/node-<network>`.
+
+
+### Importing the baker key
+
+Import your baker secret key to the data directory. There are multiple ways to import
+the key:
+
+1) You have faucet JSON file from https://faucet.tzalpha.net/.
+
+In order to activate account run:
+```
+sudo -u tezos tezos-client activate account <alias> <path-to-downloaded-json>
+```
+
+2) You know either the unencrypted or password-encrypted secret key for your address.
+
+In order to import such a key, run:
+```
+sudo -u tezos tezos-client import secret key <alias> <secret-key>
+```
+3) The secret key is stored on a ledger.
+
+Open the Tezoz Wallet app on your ledger and run the following
+to import the key:
+```
+sudo -u tezos tezos-client import secret key <alias> <ledger-url>
+```
+Apart from importing the key, you'll also need to set it up for baking. Open Tezos Baking app
+on your ledger and run the following:
+```
+sudo -u tezos tezos-client setup ledger to bake for <alias>
+```
+
+### Registering the baker
+Once the key is imported, you'll need to register your baker, in order to do that run the following
+command:
+```
+sudo -u tezos tezos-client register key <alias> as delegate
+```
+
+Check a blockchain explorer (e.g. https://tzkt.io/ or https://tzstats.com/) to see the baker status and
+baking rights of your account.
+
+### Updating baker and endorser daemons configuration
+Apart from that, you'll need to update the `BAKER_ACCOUNT` and `ENDORSER_ACCOUNT` (in
+`/etc/default/tezos-baker-<proto>` and `/etc/default/tezos-endorser-<proto>` respectively) in
+accordance to the **alias** of the imported key.
+
+### Starting daemons
+
+Once the key is imported and the configs are updated, you can start the baker and endorser daemons:
+```
+sudo systemctl start tezos-baker-<proto>.service
+sudo systemctl start tezos-endorser-<proto>.service
+```
+
+If the node isn't bootstrapped yet, the baker and endorser daemons will wait for it to bootstrap.
+
+Note that if you're baking with the ledger key, you should have the Tezos Baking app open.
+
+Once the services are started, you can check their logs via `journalctl`:
+```
+journalctl -u tezos-baker-<proto>.service
+journalctl -u tezos-endorser-<proto>.service
+```
+
+If everything was set up correctly, you shouldn't see any errors in the logs.
+
+Logs for successfully started baker service should begin with:
+```
+Node is bootstrapped.
+Baker started.
+```
+
+Logs for successfully started endorser service should begin with:
+```
+Node is bootstrapped.
+Endorser started.
+```
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ different networks, you can read more about this in [this article](https://tezos
 * [Brew tap for macOS](#macos)
 * [Systemd services for Tezos binaries](#systemd)
 * [Building instructions](#building)
+* [Setting up baking instance on Ubuntu](#baking-on-ubuntu)
 
 <a name="static-linux"></a>
 ## Obtain binaries from github release
@@ -238,6 +239,12 @@ Each signer service has dedicated config file in e.g. `/etc/default/tezos-signer
 This repository provides two distinct ways for building and packaging tezos binaries:
 * [Docker-based](./docker/README.md)
 * [Nix-based](./nix/README.md)
+
+<a name="baking-on-ubuntu"></a>
+## Setting up baking instance on Ubuntu
+
+Read [the article](./docs/baking.md) to find out an easy way to set up
+baking instance on Ubuntu using packages provided by our launchpad PPA.
 
 ## For Contributors
 


docs/baking.md
```diff
@@ -0,0 +1,159 @@
+<!--
+   - SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
+   -
+   - SPDX-License-Identifier: LicenseRef-MIT-TQ
+   -->
+# Baking with tezos-packaging on Ubuntu
+
+Tezos-packaging provides an easy way to install and set up infrastructure for
+interacting with Tezos blockchain. This article provides a step-by-step guide for
+setting up baking instance for Tezos on Ubuntu.
```
pasqu4le, line 10:
```suggestion
interacting with Tezos blockchain.

This article provides a step-by-step guide for setting up baking instance for Tezos on Ubuntu.
```
Just for legibility.

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Prerequisites
+
+### Installing required packages
+
+In order to run baking instance, you'll need the following Tezos binaries:
```
pasqu4le, line 16:
```suggestion
In order to run a baking instance, you'll need the following Tezos binaries:
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+```
+# Add PPA with Tezos binaries
+sudo add-apt-repository ppa:serokell/tezos
+sudo apt-get update
+# Install packages
+sudo apt-get install tezos-client tezos-node tezos-baker-<proto> tezos-endorser-<proto>
+```
+
+Packages for `tezos-node`, `tezos-baker-<proto>` and `tezos-endorser-<proto>` provide
+systemd units for running the corresponding binaries in the background
```
pasqu4le, line 29:
```suggestion
systemd units for running the corresponding binaries in the background.
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+
```
pasqu4le, line 30: I think it's fairly clear what `<proto>` is, but still we should probably list the options here somehow IMO.

Especially because it seems that the names are not exactly the same as the one from our release page (although only for capitalization AFAICT).

What do you think about and adding a short sentence here pointing at launchpad.net?
E.g. for `baker`s link to:
https://launchpad.net/~serokell/+archive/ubuntu/tezos/+packages?field.name_filter=tezos-baker&field.status_filter=published

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+### Downloading zcash-params
+
+Since v8.0 it's required to have `sapling-output.params` and `sapling-spend.params` files from
```
pasqu4le, line 33:
```suggestion
Since v8.0 it's required to have the `sapling-output.params` and `sapling-spend.params` files from
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+the `zcash-params` within the `XDG_DATA_DIRS` to run the `tezos-node`.
+We'll need these params to be accessed by different
+users (because systemd services are usually run as a separate user), thus the suggested way of getting
+them via `https://raw.githubusercontent.com/zcash/zcash/master/zcutil/fetch-params.sh` script which will
+download them to the `HOME` directory of the current user is not convenient. Moreover, it will
+download large `sprout-groth16.params` which is unused at the runtime.
```
pasqu4le, line 39:
```suggestion

We'll need these params to be accessible by different
users (because `systemd` services are usually run as a separate user), thus the suggested way of getting
them via [the zcash script](https://raw.githubusercontent.com/zcash/zcash/master/zcutil/fetch-params.sh), which will
download them to the `HOME` directory of the current user, is not convenient. Moreover, it will
download the large `sprout-groth16.params` which is unused at the runtime.
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+So, it's better to download them manually to the `/usr/local/share/zcash-params`, in order
+to do that run the following commands:
+```
+sudo mkdir -p /usr/local/share/zcash-params
+sudo wget https://gitlab.com/tezos/opam-repository/-/raw/v8.1/zcash-params/sapling-output.params -O /usr/local/share/zcash-params/sapling-output.params
+sudo wget https://gitlab.com/tezos/opam-repository/-/raw/v8.1/zcash-params/sapling-spend.params -O /usr/local/share/zcash-params/sapling-spend.params
+```
```
pasqu4le, line 46: What do you think about making our own script just for this and putting it into `tezos-packaging/scripts/` instead?

It could be a copy of the one mentioned above (but with a different destination and without `groth16`) or perhaps a simpler new one.

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Setting up the tezos-node
+
+### Bootstrapping the node
+
+In order to run baker locally, you'll need fully-synced local `tezos-node`.
```
pasqu4le, line 52:
```suggestion
In order to run a baker locally, you'll need a fully-synced local `tezos-node`.

```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+The easiest way to set up a node running on the `mainnet` or on one of the
+testnets (e.g. `delphinet` or `edonet`) is to use one of the predefined
+`tezos-node-<network>.services` systemd services provided in the `tezos-node`
+package. However, by default, these services will start to bootstrap the node
+from scratch, which will take significant amount of time. The fastest way to bootstrap
+the node is to import a snapshot. The snapshots can be downloaded from the https://snapshots.tulip.tools/#/ .
```
pasqu4le, line 58:
```suggestion
from scratch, which will take significant amount of time.

The fastest way to bootstrap the node is to import a snapshot.
Snapshots can be downloaded from https://snapshots.tulip.tools/#/ .
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+
+`tezos-node-<network>.service` has `/var/lib/tezos/node-<network>` as a data directory
+and `http://localhost:8732` as a an RPC address by default.
```
pasqu4le, line 61:
```suggestion
and `http://localhost:8732` as its RPC address by default.
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+These parameters can be changed in the service configuration, which is located in
+`/lib/systemd/system/tezos-node-<network>.service`. Note, that if you've updated the `.service`
+file, you should reload systemctl daemon via running `sudo systemctl daemon-reload`.
+
+All commands within the service are run under `tezos` user.
```
pasqu4le, line 66:
```suggestion
All commands within the service are run under the `tezos` user.
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+
+Download the snapshost for desired network and with desired history mode
```
pasqu4le, line 68:
```suggestion
Download the snapshost for desired network and with the desired history mode
```

pasqu4le: One more thing: it's a bit odd that the snapshots and were to find them is separated to this sentence by the part explaining the dir and parameter of the node.

Perhaps this section that's now in the middle can be moved above (just before the sentences that mention snapshots)? Or maybe it would be better to move it just below this?

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+(you can read more about `tezos-node` history modes [here](https://tezos.gitlab.io/user/history_modes.html#history-modes)).
+
+Snapshot import requires `tezos-node` data directory doesn't contain `context` and `store` folders.
```
pasqu4le, line 71:
```suggestion
Snapshot import requires that the `tezos-node` data directory doesn't contain the `context` and `store` folders.
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+                                                                                Invalid signature for block BLjutMj47caB. Expected: tz1Na5QB98cDA.
+```
+
+You should init/update config in the data directory to match the desired network
+before importing the snapshot. To do the run the following:
+```
+sudo -u tezos tezos-node config init --data-dir /var/lib/tezos/node-<network> --network <network>
+#or
+sudo -u tezos tezos-node config update --data-dir /var/lib/tezos/node-<network> --network <network>
+```
```
pasqu4le, line 93: I suppose these aren't interchangeable, right?

Wouldn't the `init` one always be necessary the first time around and the `update` one after that?

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Starting the node
+
+After the snapshot import, you can finally start the node by running
+```
+sudo systemctl start tezos-node-<network>.service
+```
+
+Note that even after the snapshot import node can still be out of sync, it may require
```
pasqu4le, line 102:
```suggestion
Note that even after the snapshot import the node can still be out of sync, it may require
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+some additional time to completely bootstrap.
+
+To stop the node run:
+```
+sudo systemctl stop tezos-node-<network>.service
+```
+
+You can check node logs via `journalctl`:
```
pasqu4le, line 110:
```suggestion
You can check the node logs via `journalctl`:
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+```
+journalctl -u tezos-node-<network>.service
+```
+
+## Setting up baker and endorser daemons
+
+Once you have fully synced local `tezos-node`, you can start baker and endorser daemons.
```
pasqu4le, line 117:
```suggestion
Once you have fully synced your local `tezos-node`, you can start the baker and endorser daemons.
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+point both `DATA_DIR`s to `/var/lib/tezos/baker`. You should also
+update `NODE_DATA_DIR` in the `/etc/default/tezos-baker-<proto>` to point at the desired
+node data directory, e.g. `/var/lib/tezos/node-<network>`.
+
+Create the aformentioned daemons data directory:
+```
+sudo -u tezos mkdir -p /var/lib/tezos/baker
+```
+
+Import you baker secret key to the data directory. There are two ways to import
```
pasqu4le, line 131:
```suggestion
Import your baker secret key to the data directory. There are two ways to import
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+such key:
+1) You know either unencrypted or password-encrypted secret key for your address. In order
+to import such key run:
+```
+sudo -u tezos tezos-client -d /var/lib/tezos/baker import secret key <alias> <secret-key>
+```
+2) Secret key is stored on ledger. Open Tezow Wallet app on your ledger and run the following
```
pasqu4le, line 138:
```suggestion
2) The secret key is stored on a ledger.

Open the Tezoz Wallet app on your ledger and run the following
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+
+Create the aformentioned daemons data directory:
+```
+sudo -u tezos mkdir -p /var/lib/tezos/baker
+```
+
+Import you baker secret key to the data directory. There are two ways to import
+such key:
+1) You know either unencrypted or password-encrypted secret key for your address. In order
+to import such key run:
```
pasqu4le, line 134:
```suggestion
1) You know either the unencrypted or password-encrypted secret key for your address.

In order to import such key run:
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+```
+sudo -u tezos tezos-client -d /var/lib/tezos/baker import secret key <alias> <secret-key>
+```
+2) Secret key is stored on ledger. Open Tezow Wallet app on your ledger and run the following
+to import the key:
+```
+sudo -u tezos tezos-client -d /var/lib/tezos/baker import secret key <alias> <ledger-url>
+```
+Apart from importing the key, you'll also need to set it up for baking. Open Tezos Baking app
```
pasqu4le, line 143: From the rendered doc it isn't quite clear if what follows is needed in both cases or not.

In fact I'd suggest to divide this whole "Setting up baker and endorser daemons" section better into sub-steps(I suppose with `h3`/`###`:
- Set up daemons data directory
- Import you baker secret key to the data directory
- Start the daemons

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+on your ledger and run the following:
+```
+sudo -u tezos tezos-client -d /var/lib/tezos/baker setup ledger to bake for <alias>
+```
+
+After importing the key, you'll need to update `BAKER_ACCOUNT` and `ENDORSER_ACCOUNT` in
+`/etc/default/tezos-baker-<proto>` and `/etc/default/tezos-endorser-<proto>` respectively with
+accordance to the alias of imported key.
```
pasqu4le, line 151:
```suggestion
After importing the key, you'll need to update the `BAKER_ACCOUNT` and `ENDORSER_ACCOUNT` in
`/etc/default/tezos-baker-<proto>` and `/etc/default/tezos-endorser-<proto>` respectively, in
accordance to the alias of the imported key.
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+
+Once the key is imported and the configs are updated, you can start baker and endorser daemons:
```
pasqu4le, line 153:
```suggestion
Once the key is imported and the configs are updated, you can start the baker and endorser daemons:
```

docs/baking.md
```diff
@@ -0,0 +1,159 @@
...
+```
+sudo systemctl start tezos-baker-<proto>.service
+sudo systemctl start tezos-endorser-<proto>.service
+```
+
+Note that if you're baking with the ledger key, you should have Tezos Baking app open.
```
pasqu4le, line 159:
```suggestion
Note that if you're baking with the ledger key, you should have the Tezos Baking app open.
```

pasqu4le (commented):
I haven't actually tried the guide, so this review is more about the form.

One more question: how come the PR has `Relates` and not `Resolves`?

rvem:
> One more question: how come the PR has Relates and not Resolves?

I assume that the issue should be resolved only once the article is published somewhere (where somewhere is still to be defined)

docs/baking.md
rvem: Articles have always been tricky for me :joy: :

docs/baking.md
rvem: Apparently, it makes some sense

docs/baking.md
rvem: Yes, I think that the paragraph that starts from '`tezos-node-<network>.service` has `/var/lib/tezos/node-<network>` as a data directory' can be moved to the very top of the 'Setting up the tezos-node' part, right before the 'Bootstrapping the node'

docs/baking.md
rvem: Yes, only one of these two commands will work. It depends on whether the config was initialized previously by the node startup script.

Apparently, the wording should be changed to something like ''To do the run one of the following commands depending on whether the config in the given data directory was initialized previously:"

rvem:
@pasqu4le, update, please rereview

docs/baking.md
pasqu4le: Naaa, these are just minor fixes, overall it's well written IMO

docs/baking.md
```diff
@@ -0,0 +1,182 @@
...
+### Downloading zcash-params
+
+Since v8.0 it's required to have the `sapling-output.params` and `sapling-spend.params` files from
+the `zcash-params` within the `XDG_DATA_DIRS` to run the `tezos-node`.
+
+We'll need these params to be accessible by different
+users (because `systemd` services are usually run as a separate user), thus the suggested way of getting
+them via [the zcash script](https://raw.githubusercontent.com/zcash/zcash/master/zcutil/fetch-params.sh), which will
+download them to the `HOME` directory of the current user, is not convenient. Moreover, it will
+download the large `sprout-groth16.params` which is unused at the runtime.
```
pasqu4le, line 46:
```suggestion
download the large `sprout-groth16.params` which is unused at runtime.
```
I think here the "the" is unnecessary.

docs/baking.md
```diff
@@ -0,0 +1,182 @@
...
+So, it's better to download them to the `/usr/local/share/zcash-params`, the easiest way to do that
+is to run [`fetch-sapling-params.sh`](../scripts/fetch-sapling-params.sh) script:
+```
+sudo ./fetch-salping-params.sh
```
pasqu4le, line 50:
```suggestion
sudo ./fetch-sapling-params.sh
```
Also, I wonder if we should provide an easier way to grab this, this kinda presumes that the repo has been cloned or at least the file downloaded, perhaps it would be better to include the command here that download an executes the script.

On the same point I guess is the link just above, but I think we'll cross that bridge when we'll know where this is going to end up.

docs/baking.md
```diff
@@ -0,0 +1,182 @@
...
+
+Snapshot import requires that the `tezos-node` data directory doesn't contain the `context` and `store` folders.
+Remove them before the snapshot import.
+
+In order to import the snapshot run the following command:
+```
+sudo -u tezos-node snapshot import --data-dir /var/lib/tezos/node-<network>/ <path to the snapshot file>
+```
+
+In case you're getting the error similar to
```
pasqu4le, line 92:
```suggestion
In case you're getting an error similar to
```

docs/baking.md
```diff
@@ -0,0 +1,182 @@
...
+Sudo is required because the files are downloaded to the `/usr/local/share/zcash-params`.
+This script also checks that hashsums of the downloaded files match the expected values.
+
+## Setting up the tezos-node
+
+The easiest way to set up a node running on the `mainnet` or on one of the
+testnets (e.g. `delphinet` or `edonet`) is to use one of the predefined
+`tezos-node-<network>.services` systemd services provided in the `tezos-node`
+package. However, by default, these services will start to bootstrap the node
+from scratch, which will take significant amount of time.
```
pasqu4le, line 62:
```suggestion
package. 

However, by default, these services will start to bootstrap the node from scratch,
which will take a significant amount of time.
In order to avoid this, here we suggest bootstrapping from a snapshot instead.
```
Just to push towards "keep reading" and not "start waiting" :sweat_smile:

scripts/fetch-sapling-params.sh
```diff
@@ -0,0 +1,22 @@
...
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+# This script is an alternative for
+# https://raw.githubusercontent.com/zcash/zcash/master/zcutil/fetch-params.sh.
+# This script downloads only `sapling-*.params` files to the `/usr/local/share/zcash-params`,
+# so that the downloaded files can be used by different users. Since it downloads files
+# to `/usr/local/share`, it should be run with sudo.
+set -euo pipefail
+
```
pasqu4le, line 12: Since `sudo` is pretty much necessary, how about adding this for a better error message?
```suggestion

if [ "$EUID" -ne 0 ]; then 
        echo "This script requires root privileges."
        echo "Please run it again with sudo or as root."
        exit 1
fi

```

pasqu4le (commented):
Ok a couple more minor suggestions.

Aside from these I will try the instructions on a fresh Ubuntu VM (hopefully very soon) and approve after that.

docs/baking.md
rvem: :facepalm:

scripts/fetch-sapling-params.sh
pasqu4le: I think this suggestion is what caused the trailing whitespaces check to fail (line 13).

@rvem  Can you check if that's the case and fix it?

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+
+Download the snapshost for desired network and with the desired history mode
+(you can read more about `tezos-node` history modes [here](https://tezos.gitlab.io/user/history_modes.html#history-modes)).
+
+Snapshot import requires that the `tezos-node` data directory doesn't contain the `context` and `store` folders.
+Remove them before the snapshot import.
+
+In order to import the snapshot run the following command:
+```
+sudo -u tezos-node snapshot import --data-dir /var/lib/tezos/node-<network>/ <path to the snapshot file>
```
pasqu4le, line 92:
```suggestion
sudo -u tezos tezos-node snapshot import --data-dir /var/lib/tezos/node-<network>/ <path to the snapshot file>
```
missing the user :smile: 

Also, I'm trying following this on a fresh VM and it seems that there is no `tezos` user at this point, should it have been created by some of the previous steps (I doubt it, not much has happened), by one of the following ones, or should it be created manually?
If the latter is the case I think just above where we mention that all commands are run under `tezos` it should also say how to create the user.

pasqu4le: There's another problem at this point, if one has never `start`ed the node service with `systemctl` before there in no `/var/lib/tezos` directory and this command will fail (with an `Invalid data directory` error which is not very helpful).

So I suggest that we either:
1. add the commands to create the data directory manually in the `Setting up node data directory` section
2. rearrange this part of the guide a bit and have the user `start` the node in that section and the `stop` it and `import` the snapshot after.

What do you think?

P.s. starting the `node` seems to not create the `tezos` user automatically, unless I've done something wrong

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+users (because `systemd` services are usually run as a separate user), thus the suggested way of getting
+them via [the zcash script](https://raw.githubusercontent.com/zcash/zcash/master/zcutil/fetch-params.sh), which will
+download them to the `HOME` directory of the current user, is not convenient. Moreover, it will
+download the large `sprout-groth16.params` which is unused at runtime.
+So, it's better to download them to the `/usr/local/share/zcash-params`, the easiest way to do that
+is to run [`fetch-sapling-params.sh`](../scripts/fetch-sapling-params.sh) script:
+```
+sudo ./fetch-sapling-params.sh
+```
+
```
itkach, line 52: is there anything that prevents sapling param files from being packaged into deb and installed as a dependency? these files are, in fact a hard runtime dependency for most tezos binaries, so if installing tezos-node does not also install sapling param the installation is broken/incomplete

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+which will take a significant amount of time.
+In order to avoid this, here we suggest bootstrapping from a snapshot instead.
+
+### Setting up node data directory
+
+`tezos-node-<network>.service` has `/var/lib/tezos/node-<network>` as a data directory
+and `http://localhost:8732` as its RPC address by default.
+These parameters can be changed in the service configuration, which is located in
+`/lib/systemd/system/tezos-node-<network>.service`. Note, that if you've updated the `.service`
+file, you should reload systemctl daemon via running `sudo systemctl daemon-reload`.
```
itkach, line 73: users should be cautioned against editing .service files directly since they are part of the package; it [seems that current recommended way](https://serverfault.com/questions/413397/how-to-set-environment-variable-in-systemd-service?answertab=votes#tab-top) is to use `systemctl edit`

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Bootstrapping the node
+
+In order to run a baker locally, you'll need a fully-synced local `tezos-node`.
+
+The fastest way to bootstrap the node is to import a snapshot.
+Snapshots can be downloaded from https://snapshots.tulip.tools/#/ .
```
itkach, line 80: it would be good to give some alternatives here, https://snapshots-tezos.giganode.io/ and https://xtz-shots.io/ in particular are great

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+
+All commands within the service are run under the `tezos` user.
+
+Download the snapshost for desired network and with the desired history mode
```
itkach, line 84: typo: snapshost -> snapshots

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+(you can read more about `tezos-node` history modes [here](https://tezos.gitlab.io/user/history_modes.html#history-modes)).
+
```
itkach, line 86: I think it would be better to make a concrete recommendation here that would work for most new users and save them time/effort by not requiring them to do a research, e.g. "download rolling snapshot" - that's smallest, fastest and is sufficient for baking. Tradeoffs, limitations and further reading should be separated into a note or a footnote so that they can be easily skipped if a user wants to get to the end of the happy path quickly

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+Snapshot import requires that the `tezos-node` data directory doesn't contain the `context` and `store` folders.
+Remove them before the snapshot import.
```
itkach, line 88: similarly, this is not something a user should encounter if the instructions to get to this point were correct and followed precisely, so should be a troubleshooting note/footnote rather than part of main narrative

docs/baking.md
itkach: `/var/lib/tezos` should be created as part of .deb installation

docs/baking.md
itkach: it looks like this section doesn't actually set up anything, it's just information; would be good to explain why user may want/need to change the defaults

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+              Invalid block BLjutMj47caB
+                Failed to validate the economic-protocol content of the block: Error:
+                                                                                Invalid signature for block BLjutMj47caB. Expected: tz1Na5QB98cDA.
+```
+
+You should init/update config in the data directory to match the desired network
+before importing the snapshot. To do that run one of the following commands
+depending on whether the config in the given data directory was initialized previously:
+```
+sudo -u tezos tezos-node config init --data-dir /var/lib/tezos/node-<network> --network <network>
```
itkach, line 107: is it possible to avoid having to specify `--data-dir` in all the commands? isn't it coming from tezos node config file? isn't there a default?

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Starting the node
+
+After the snapshot import, you can finally start the node by running
+```
+sudo systemctl start tezos-node-<network>.service
+```
+
+Note that even after the snapshot import the node can still be out of sync, it may require
+some additional time to completely bootstrap.
```
itkach, line 120: how does one check if node is fully synced, both initially and on-going?

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+
+You can check the node logs via `journalctl`:
+```
+journalctl -u tezos-node-<network>.service
+```
+
+## Setting up baker and endorser daemons
+
+Once you have fully synced your local `tezos-node`, you can start the baker and endorser daemons.
+
```
itkach, line 135: do you actually have to wait for the node to fully sync or can you start baker and endorser and they will figure out when the node is ready themselves?

docs/baking.md
```diff
@@ -0,0 +1,185 @@
...
+`/etc/default/tezos-baker-<proto>` and `/etc/default/tezos-endorser-<proto>`. Make
+sure to point them at the same directory to have the same set of known keys, e.g.
+point both `DATA_DIR`s to `/var/lib/tezos/baker`. You should also
+update `NODE_DATA_DIR` in the `/etc/default/tezos-baker-<proto>` to point at the desired
+node data directory, e.g. `/var/lib/tezos/node-<network>`.
+
+Create the aformentioned daemons data directory:
+```
+sudo -u tezos mkdir -p /var/lib/tezos/baker
+```
```
itkach, line 148: why does this require user action at all? the defaults should be correct out of the box, `/var/lib/tezos/baker` should be part of `.deb` package

docs/baking.md
rvem: Yes, I think it makes sense to add one more `tezos-sapling-params` package and use it as a dependency for all other packages. Created #132

docs/baking.md
rvem: I assumed that both `tezos` user and `/var/lib/tezos` are both created during the package installation, but looks like it's not true :confused:. 
I Will check that additionally

docs/baking.md
rvem: You can indeed start both daemons, but they won't do anything unless the node is completely bootstrapped

docs/baking.md
rvem: Agree, created #133

docs/baking.md
rvem: AFAIU, in case when  `--data-dir` isn't presented it'll use `$HOME/.tezos-node` by default. It's possible to specify `--config-file` instead of `--data-dir`, but you'll still have to provide at least one of these options

itkach: these commands are to be executed as  `tezos` user, who's $HOME directory should be set to `/var/lib/tezos` and then defaults work just fine; also, a new user on a happy path/fast track to running node and baking on `mainnet` - which I think is the primary use case - shouldn't need to do any of this at all because `mainnet` is the default. To simplify things further (for the user, not necessarily for the packager 😈 ) `config init` should probably run at package installation time, so that user only ever needs to run `config update`, and that also becomes a note - explanation of how to set up for a testnet instead of mainnet - rather than the main story line.

docs/baking.md
itkach: _unless_ or _until_? if user can set everything up upfront and later just needs to check  that everything looks good once the node is synced is less mental effort than pausing the entire set up until node is bootstrapped and resuming it later, having to remember where were you in this complex workflow

rvem: > unless or until?

:facepalm: 
of course, until

scripts/fetch-sapling-params.sh
rvem: This script should be useless after #132

docs/baking.md
rvem: The aforementioned problems with user and directory creations should be fixed now

docs/baking.md
rvem: `config init` call is now a part of `tezos-node` package postinsallation

docs/baking.md
```diff
@@ -0,0 +1,175 @@
...
+
+All commands within the service are run under the `tezos` user.
+
+Download the snapshot for desired network. We recommend to use rolling snapshot, this is
+the smallest and the fastest mode that is sufficient for baking (you can read more about other
+`tezos-node` history modes [here](https://tezos.gitlab.io/user/history_modes.html#history-modes)).
+
+In order to import the snapshot run the following command:
+```
+sudo -u tezos tezos-node snapshot import --data-dir /var/lib/tezos/node-<network>/ <path to the snapshot file>
```
pasqu4le, line 77: Testing this on a fresh VM it seems there are permission problems here for the `tezos` user.

Can you check that's the case?

docs/baking.md
pasqu4le: There is now a problem with this, which is that the `--network` option does not work for `edo2net`, which I think will be somewhat common considering it's the only other network presented aside from `mainnet`

docs/baking.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Setting up baker and endorser daemons
+
+### Setting up daemons data directory
+
+Data directories for baker and endorser daemons are defined in the
+`/etc/default/tezos-baker-<proto>` and `/etc/default/tezos-endorser-<proto>`.
+By default, both this daemons have `/var/lib/tezos/client` set as a `DATA_DIR`.
+
+Additionally, you should specify `NODE_DATA_DIR` in the `/etc/default/tezos-baker-<proto>`
```
pasqu4le, line 134:
```suggestion
Additionally, you need to specify `NODE_DATA_DIR` in the `/etc/default/tezos-baker-<proto>`
```
AFAIU this is a necessity at this point in the tutorial

docs/baking.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Starting daemons
+
+Once the key is imported and the configs are updated, you can start the baker and endorser daemons:
+```
+sudo systemctl start tezos-baker-<proto>.service
+sudo systemctl start tezos-endorser-<proto>.service
+```
+
+If the node isn't bootstrapped yet, baker and endorser daemons will wait for it to bootstrap.
```
pasqu4le, line 173:
```suggestion
If the node isn't bootstrapped yet, the baker and endorser daemons will wait for it to bootstrap.
```

docs/baking.md
```diff
@@ -0,0 +1,175 @@
...
+```
+sudo systemctl start tezos-node-<network>.service
+```
+
+Note that even after the snapshot import the node can still be out of sync, it may require
+some additional time to completely bootstrap. In order to check whether the node is bootstrapped,
+you can use `tezos-client`:
+```
+tezos-client --endpoint <node-rpc-address> bootstrapped
+```
```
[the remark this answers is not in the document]
pasqu4le: Actually even better, if the tezos client default config had the right values this could be just replaced with:
```
sudo -u tezos tezos-client bootstrapped
```

docs/baking.md
```diff
@@ -0,0 +1,175 @@
...
+In order to import such key run:
+```
+sudo -u tezos tezos-client -d /var/lib/tezos/client import secret key <alias> <secret-key>
+```
+2) The secret key is stored on a ledger.
+
+Open the Tezoz Wallet app on your ledger and run the following
+to import the key:
+```
+sudo -u tezos tezos-client -d /var/lib/tezos/client import secret key <alias> <ledger-url>
```
pasqu4le, line 153: This and the one just above also could be used without the `-d` argument if we change the default config

docs/baking.md
pasqu4le: Additionally, while we fix this I think that in the next release we should add aliases for `tezos-node-<network>` that are just `tezos-node` with the correct `data-dir`.

This would provide a great UX as these aliases names would match the service names as well.

pasqu4le: Ok so #148 is blocked, let's not wait for it for this article.

docs/baking.md
```diff
@@ -0,0 +1,175 @@
+<!--
+   - SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
```
pasqu4le, line 2: Especially considering that this is a fresh file, let's start updating the copyright here.
```suggestion
   - SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
```

docs/baking.md
```diff
@@ -0,0 +1,175 @@
...
+you can use `tezos-client`:
+```
+sudo -u tezos tezos-client bootstrapped
+```
+
+To stop the node run:
+```
+sudo systemctl stop tezos-node-<network>.service
+```
+
```
pasqu4le, line 120: I would add another small suggestion here about enabling the service as well, like:
```
If you want the service start automatically at boot use `enable`:
<enable example here>
and to disable it from doing so, instead use:
<disable example here>
```

pasqu4le (approved):
A couple minor nitpicks, feel free to squash them in (and to edit the commit message)

I went through it all with the newest packages and found no problems. :ok_hand:

docs/baking.md
rvem: :see_no_evil:

itkach:
Following the article, I got to the point where I am starting baker and endorser and getting errors from both:
Endorser:
```
ubuntu@tez1:~$ journalctl -u tezos-endorser-008-ptedo2zk.service
-- Logs begin at Thu 2021-03-04 13:32:42 EST, end at Thu 2021-03-04 15:29:09 EST. --
Mar 04 15:26:49 tez1 systemd[1]: Started Tezos endorser.
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]: Error:
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   Erroneous command line argument 2 (tz1TtTLRuWZG4SRw8u3Pi9nb2wLhrwB4QV2h).
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   no public key hash alias named tz1TtTLRuWZG4SRw8u3Pi9nb2wLhrwB4QV2h
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]: Usage:
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   tezos-endorser-008-PtEdo2Zk [global options] command [command options]
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   tezos-endorser-008-PtEdo2Zk --help (for global options)
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   tezos-endorser-008-PtEdo2Zk [global options] command --help (for command options)
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]: To browse the documentation:
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   tezos-endorser-008-PtEdo2Zk [global options] man (for a list of commands)
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   tezos-endorser-008-PtEdo2Zk [global options] man -v 3 (for the full manual)
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]: Global options (must come before the command):
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -d --base-dir <path>: client data directory
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -c --config-file <path>: configuration file
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -t --timings: show RPC request times
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   --chain <hash|tag>: chain on which to apply contextual commands (possible tags are 'main' and >
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -b --block <hash|tag>: block on which to apply contextual commands (possible tags are 'head' a>
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -w --wait <none|<int>>: how many confirmation blocks before to consider an operation as includ>
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -p --protocol <hash>: use commands of a specific protocol
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -l --log-requests: log all requests to the node
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -A --addr <IP addr|host>: [DEPRECATED: use --endpoint instead] IP address of the node
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -P --port <number>: [DEPRECATED: use --endpoint instead] RPC port of the node
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -S --tls: [DEPRECATED: use --endpoint instead] use TLS to connect to node.
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -E --endpoint <uri>: HTTP(S) endpoint of the node RPC interface; e.g. 'http://localhost:8732'
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -R --remote-signer <uri>: URI of the remote signer
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -f --password-filename <filename>: path to the password filename
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   -M --mode <client|mockup>: how to interact with the node
Mar 04 15:26:49 tez1 systemd[1]: tezos-endorser-008-ptedo2zk.service: Main process exited, code=exited, status=1/FAILURE
Mar 04 15:26:49 tez1 systemd[1]: tezos-endorser-008-ptedo2zk.service: Failed with result 'exit-code'.
```

Baker:

```
ubuntu@tez1:~$ journalctl -u tezos-baker-008-ptedo2zk.service
-- Logs begin at Thu 2021-03-04 13:45:09 EST, end at Thu 2021-03-04 15:32:49 EST. --
Mar 04 15:26:42 tez1 systemd[1]: Started Tezos baker.
Mar 04 15:26:43 tez1 tezos-baker-start[947652]: Error:
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   Erroneous command line argument 5 ().
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   Directory doesn't exist: ''
Mar 04 15:26:43 tez1 tezos-baker-start[947652]: Usage:
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   tezos-baker-008-PtEdo2Zk [global options] command [command options]
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   tezos-baker-008-PtEdo2Zk --help (for global options)
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   tezos-baker-008-PtEdo2Zk [global options] command --help (for command options)
Mar 04 15:26:43 tez1 tezos-baker-start[947652]: To browse the documentation:
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   tezos-baker-008-PtEdo2Zk [global options] man (for a list of commands)
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   tezos-baker-008-PtEdo2Zk [global options] man -v 3 (for the full manual)
Mar 04 15:26:43 tez1 tezos-baker-start[947652]: Global options (must come before the command):
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -d --base-dir <path>: client data directory
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -c --config-file <path>: configuration file
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -t --timings: show RPC request times
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   --chain <hash|tag>: chain on which to apply contextual commands (possible tags are 'main' and 'te>
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -b --block <hash|tag>: block on which to apply contextual commands (possible tags are 'head' and >
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -w --wait <none|<int>>: how many confirmation blocks before to consider an operation as included
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -p --protocol <hash>: use commands of a specific protocol
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -l --log-requests: log all requests to the node
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -A --addr <IP addr|host>: [DEPRECATED: use --endpoint instead] IP address of the node
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -P --port <number>: [DEPRECATED: use --endpoint instead] RPC port of the node
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -S --tls: [DEPRECATED: use --endpoint instead] use TLS to connect to node.
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -E --endpoint <uri>: HTTP(S) endpoint of the node RPC interface; e.g. 'http://localhost:8732'
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -R --remote-signer <uri>: URI of the remote signer
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -f --password-filename <filename>: path to the password filename
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   -M --mode <client|mockup>: how to interact with the node
Mar 04 15:26:43 tez1 systemd[1]: tezos-baker-008-ptedo2zk.service: Main process exited, code=exited, status=1/FAILURE
Mar 04 15:26:43 tez1 systemd[1]: tezos-baker-008-ptedo2zk.service: Failed with result 'exit-code'.
```

Config files:

```
ubuntu@tez1:~$ cat /etc/default/tezos-endorser-008-ptedo2zk
# vim: ft=sh
#
# SPDX-License-Identifier: LicenseRef-MIT-TQ

# shellcheck disable=SC2034
ENDORSER_ACCOUNT="tz1TtTLRuWZG4SRw8u3Pi9nb2wLhrwB4QV2h"
ubuntu@tez1:~$ cat /etc/default/tezos-endorser-008-ptedo2zk
# vim: ft=sh
#
# SPDX-License-Identifier: LicenseRef-MIT-TQ

# shellcheck disable=SC2034
ENDORSER_ACCOUNT="tz1TtTLRuWZG4SRw8u3Pi9nb2wLhrwB4QV2h"
```

also tried the same with alias instead of tz address, same result. 

```
ubuntu@tez1:~$ tezos-client list known addresses
Warning:

                 This is NOT the Tezos Mainnet.

           Do NOT use your fundraiser keys on this network.

igor-edo-2: tz1TtTLRuWZG4SRw8u3Pi9nb2wLhrwB4QV2h (unencrypted sk known)
igor-edo-1: tz1RFYRWWU5DgKd5otN2HAkKZoyRsbLiPjwM (unencrypted sk known)
```

I attempted to set this up with a testnet address from https://faucet.tzalpha.net/. I don't think I need to separately import any keys since tezos-client knows them from `tezos-client activate` command. Btw, the article should also describe setting up baking with test accounts from https://faucet.tzalpha.net/.

Also, I don't see a step with actual baker registration, `tezos-client register key <baking_address> as delegate` - shouldn't there be one?

itkach:
Article should recommend that user checks the logs after starting daemons and describe what success looks like

pasqu4le:
@itkach from the commands you shared it seems possible that you imported the keys using `tezos-client` from your own user and not `tezos`:
> ubuntu@tez1:~$ tezos-client list known addresses

Those have different data dirs (and known addresses) does `sudo -u tezos tezos-client list known addresses` show them as well?

rvem:
> Mar 04 15:26:43 tez1 tezos-baker-start[947652]: Error:
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   Erroneous command line argument 5 ().
Mar 04 15:26:43 tez1 tezos-baker-start[947652]:   Directory doesn't exist: ''

Have you specified `NODE_DATA_DIR` in the `/etc/default/tezos-baker-008-ptedo2zk`?

> Mar 04 15:26:49 tez1 tezos-endorser-start[948484]: Error:
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   Erroneous command line argument 2 (tz1TtTLRuWZG4SRw8u3Pi9nb2wLhrwB4QV2h).
Mar 04 15:26:49 tez1 tezos-endorser-start[948484]:   no public key hash alias named tz1TtTLRuWZG4SRw8u3Pi9nb2wLhrwB4QV2h

Both baker and endorser daemons must use public hash key alias instead of plain `tz1` address (according to the manual).

itkach:
> @itkach from the commands you shared it seems possible that you imported the keys using `tezos-client` from your own user and not `tezos`

indeed, that was my mistake

itkach:
> Have you specified `NODE_DATA_DIR` in the `/etc/default/tezos-baker-008-ptedo2zk`?

Indeed, I missed that instruction. Is there a reason it's not populated though? Seems like it should be 
```NODE_DATA_DIR="/var/lib/tezos/node-edo2net/"```
out of the box and user shouldn't need to change this.

**Comment by itkach** [comment]
> Also, I don't see a step with actual baker registration, tezos-client register key <baking_address> as delegate - shouldn't there be one?

for keys from a file - definitely, not sure if "setup ledger to bake" does this automatically, didn't test that yet

**Comment by itkach** [comment]
> Seems like it should be
> `NODE_DATA_DIR="/var/lib/tezos/node-edo2net/"`
> out of the box and user shouldn't need to change this.

to answer myself, nodes connect to a particular network which may switch protocols and bakers are protocol specific

**Comment by rvem** [comment]
> Also, I don't see a step with actual baker registration, tezos-client register key <baking_address> as delegate - shouldn't there be one?

Yes, looks like I missed this step:(

File: docs/baking.md [hunk 48]
```diff
@@ -0,0 +1,223 @@
...
+### Registering the baker
+Once the key is imported, you'll need to register your baker, in order to do that run the following
+command:
+```
+sudo -u tezos tezos-client register key <alias> as delegate
+```
+
+Check blockchain explorer (https://tzkt.io/ or https://tzstats.com/) to see the baker status and
```
pasqu4le, line 185:
```suggestion
Check a blockchain explorer (e.g. https://tzkt.io/ or https://tzstats.com/) to see the baker status and
```

docs/baking.md
```diff
@@ -0,0 +1,223 @@
...
+```
+sudo systemctl start tezos-baker-<proto>.service
+sudo systemctl start tezos-endorser-<proto>.service
+```
+
+If the node isn't bootstrapped yet, the baker and endorser daemons will wait for it to bootstrap.
+
+Note that if you're baking with the ledger key, you should have the Tezos Baking app open.
+
+Once the services were started, you can check their logs via `journalctl`:
```
pasqu4le, line 205:
```suggestion
Once the services are started, you can check their logs via `journalctl`:
```
Or `have been`, I guess it would also be correct.

pasqu4le (commented):
A couple more nitpicks :)

PR #137, opened 2021-02-11
Draft: [#135] Store opam dependencies separately from Tezos sources

### LINKED ISSUE #135: Consider storing opam dependencies separately from tezos source in Ubuntu source packages
Currently, all opam dependencies sources are stored along with the tezos sources in the `.orig.tar.gz` source archive.

However, sometimes set of opam dependencies can change, while tezos doesn't change its version. Such change can
cause a problem, because `.orig.tar.gz` must be the same for packages with the same upstream version (e.g. same for
`8.1-1` and `8.1-2`), so we cannot update opam dependencies without bumping the upstream version.

We should check, whether it's possible to store opam dependencies sources in the `.debian.tar.xz` archive.
Specifically, we should do the following:
1) Check whether files from `.debian.tar.xz` can be used during the source package build.
2) If they can be used, find a way to put all the opam dependencies to this archive.
    Maybe the existing approach with `opam-bundle` can be adapted for this.

# Acceptance criteria
1) Research, whether it's possible to store opam dependencies separately from upstream sources.
2) If it's possible, implement such packaging and use it for Ubuntu source packages.

rvem:
## Description
Problem: Sometimes (quite rarely, tbh) we want to update the set of
opam dependencies without updating the Tezos sources and without
bumping packages version. However, it's impossible for Ubuntu packages
when the Tezos sources and its opam dependencies are stored together in
`.orig.tar.gz` archive.

Solution: Move all opam dependencies to the `.debian.tar.xz` archive
with metafiles, thus it's possible to update these dependencies along
with package release number update.

## Related issue(s)

Resolves #135

  can be successfully built.

### DIFF
--- a/docker/package/Dockerfile-fedora
+++ b/docker/package/Dockerfile-fedora
@@ -5,7 +5,7 @@
 FROM fedora:32
 RUN dnf update -y
 RUN dnf install -y libev-devel gmp-devel hidapi-devel libffi-devel m4 perl pkg-config \
-  rpmdevtools python3 wget opam rsync which cargo
+  rpmdevtools python3 wget opam rsync git which cargo
 ENV USER dockerbuilder
 RUN useradd dockerbuilder && mkdir /tezos-packaging
 ENV HOME /tezos-packaging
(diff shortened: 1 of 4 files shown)


rvem:
Apparently, we no longer need this feature since native packaging mechanism was reworked to use tezos sources and tezos opam-repository fork directly

PR #142, opened 2021-02-16
[#132] Native package with sapling params

rvem:
## Description
Sapling params are a hard runtime dependency for some of the Tezos binaries.
This PR adds a native package that wraps the required `sapling-spend.params` and `sapling-output.params`.
This package is later used as a dependency for `tezos-client`, `tezos-node` and `tezos-baker` packages.

## Related issue(s)

Resolves #132 

#### Related changes (conditional)

- [x] I checked whether I should update the [README](../tree/master/README.md)

- [x] I checked whether native packaging works, i.e. native binary packages
  can be successfully built.

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).

PR #143, opened 2021-02-17
[#133] Native packages postinstalls

### LINKED ISSUE #133: Create default directories for systemd services
Some of the systemd services provided by the native packages have default directories for storing
services data. These directories should be created automatically during the process of `.deb` package installation.
Also, looks like `tezos` user isn't added during the package installation.

# Acceptance criteria
Directories such as `/var/lib/tezos/node-*` and `/var/lib/tezos/baker` are created automatically during the process of the
package installation. `tezos` user and group are also created during the package installation.

rvem:
## Description
Problem: We want users and default data directories to be created
automatically along with the package installation.

Solution: Add scripts that do that in the package post-installation
stage.

Currently based on #142 

## Related issue(s)

Resolves #133

  can be successfully built.

### DIFF
--- a/docker/package/packages.py
+++ b/docker/package/packages.py
@@ -127,6 +127,8 @@ default_testnets = {
     "008-PtEdoTez": "edonet"
 }
 
+daemon_postinst = postinst_steps_common + "\nmkdir -p /var/lib/tezos/client\n"
+
 for proto in active_protocols:
     service_file_baker = ServiceFile(Unit(after=["network.target"],
                                           description="Tezos baker"),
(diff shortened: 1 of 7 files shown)


docker/package/packages.py
```diff
@@ -136,18 +143,24 @@
...
+mkdir -p /var/lib/tezos/accuser'''))
     packages.append(OpamBasedPackage(f"tezos-endorser-{proto}", "Daemon for endorsing",
                                      [SystemdUnit(service_file=service_file_endorser,
                                                   startup_script="tezos-endorser-start",
                                                   config_file="tezos-endorser.conf")],
                                      proto,
-                                     optional_opam_deps=["tls", "ledgerwallet-tezos"]))
+                                     optional_opam_deps=["tls", "ledgerwallet-tezos"],
+                                     postinst_steps=postinst_steps_common + '''
+mkdir -p /var/lib/tezos/endorser'''))
```
pasqu4le, line 164: So looking back at the part of the article that originated this issue I think there's still two things that are required from the user (and probably shouldn't):

1, the endorser and baker have two different `DATA_DIR`, which could be fixed by changing here and in `docker/package/defaults/tezos-endorser.conf`. Although there may be some downside for doing that?
2. If I'm not mistaken it's still necessary to set the `NODE_DATA_DIR` for the baker manually since it's not specified in `docker/package/defaults/tezos-baker.conf`, but since we are now using and creating all the default dirs, I suppose that should be changed?

rvem: 1) Well, pointing both baker and endorser to the same `/var/lib/tezos/baker` could be confusing, because AFAIU, endorser can run without a baker and vice versa.
2) Yes, because usually there are at least two networks that one might use: mainnet and testnet that correspond to the current mainnet. So I propose to require the user to set the `NODE_DATA_DIR` manually so that he knows which network he's using

rvem: Regarding 1.: maybe it's worth pointing them to a single directory called `/var/lit/tezos/client` and mention that this directory stores all keys related information

pasqu4le: Ok, regarding 2, that's a good point, I agree :ok_hand: 

Actually the same kind of reasoning can be applied to the first point, as indeed the `backer` and `endorser` don't have to run on the same keys and there are situations where that could be used.

So I think even there the user should be left to change that in the tutorial himself.

No change required, all things considered :+1:

pasqu4le (approved):

PR #145, opened 2021-02-17
[#144] edo2net node service

### LINKED ISSUE #144: Add predefined edo2net systemd service for tezos-node
In v8.2 edo protocol was updated, a bit later new testnet called edo2net was started. 
However, `tezos-node` doesn't have a predefined `edo2net` `--network` option. This makes
testnet usage quite inconvenient with v8.2 binaries because one has to manually provide node config for
edo2net.

Note that after #108 and #143, we should have `tezos-node` service with a custom config
and machinery for package post-installation scripts. 
I propose the following way to support edo2net: define one more `tezos-node` service.
Create `edo2net` node config file during `tezos-node` postinstallation stage.

# Acceptance criteria
`tezos-node` native packages have predefined systemd service for running edo2net node.

rvem:
## Description
Problem: v8.2 tezos-node doesn't have a predefined network for edo2net,
which makes it really inconvenient to run a node that uses this new
testnet.

Solution: Add a tezos-node-edo2net service, which uses edo2net config
created during the package installation.

~Currently based on #143~

## Related issue(s)

Resolves #144

  can be successfully built.


docker/package/packages.py
```diff
@@ -89,6 +110,28 @@
     node_units.append(SystemdUnit(suffix=network,
                                   service_file=service_file,
                                   startup_script="tezos-node-start"))
+
+# TODO: refactor once #108 is merged
+node_units.append(SystemdUnit(
+    suffix="edo2net",
+    service_file=ServiceFile(Unit(after=["network.target"], requires=[],
+                                  description=f"Tezos node {network}"),
```
pasqu4le, line 118: where is this `network` coming from? :thinking:

pasqu4le (commented):
mostly LGTM, will review again after all the dependencies are merged

docker/package/packages.py
rvem: I assume the last value from the `for` loop above is used here :see_no_evil:

rvem:
@pasqu4le, all dependencies were merged, PR was rebased, please rereview

protocols.json
```diff
@@ -14,10 +14,10 @@
     "005-PsBABY5H",
     "005-PsBabyM1",
     "006-PsCARTHA",
-    "008-PtEdoTez"
+    "008-PtEdoTez",
+    "007-PsDELPH1"
```
pasqu4le, line 18: Nitpick, but I suppose they should be in the opposite order.

docker/package/packages.py
```diff
@@ -85,19 +85,53 @@ def mk_node_unit(suffix, env, desc):
...
+rm -f /var/lib/tezos/node-edo2net/config.json
+cat > /var/lib/tezos/node-edo2net/config.json <<- EOM
+{edo2net_config}
+EOM
+chown tezos:tezos /var/lib/tezos/node-edo2net/config.json
+'''
+
 packages.append(OpamBasedPackage("tezos-node",
                                  "Entry point for initializing, configuring and running a Tezos node",
                                  node_units,
```
pasqu4le, line 137: Shouldn't we embed `007` now as well (in the `optional_opam_deps` one line below)

pasqu4le (commented):
It looks like there are some tests on `007` that are now failing.

docker/package/packages.py
rvem: For now, 007 is not an optional dependency, see [here](https://opam.ocaml.org/packages/tezos-node/tezos-node.8.2/)

rvem:
Tests should be fixed now

pasqu4le (approved):
LGTM, one last little comment is to replace `Deactive` with `Deactivate` in the third commit message.

PR #149, opened 2021-02-26
[#130] Brew services for Tezos binaries

rvem:
## Description
It's convenient to run some of the Tezos binaries in the background (`tezos-{accuser, baker, endorser}-<proto>`, `tezos-node`
and `tezos-signer`). We already provide systemd services for them, however, we don't provide anything similar for macOS.

This PR updates existing and adds new brew formulas to provide background services for the same subset of Tezos binaries.

Additionally, this PR adds formula with `tezos-signer-params` which includes sapling params that are required in the runtime
by some of the Tezos binaries.

## Related issue(s)

Resolves part of #130, post-install stuff is left for another PR.

#### Related changes (conditional)

- [x] I checked whether I should update the [README](../tree/master/README.md)

- [x] I checked whether native packaging works, i.e. native binary packages
  can be successfully built.

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).

PR #150, opened 2021-03-02
[#130] brew formulas postinstalls

### LINKED ISSUE #130: Provide background services in brew formulas
We provide systemd services configurations in both Ubuntu and Fedora `tezos-node` and daemons packages.
Unfortunately, there is no such thing as systemd on macOS, however there is `launchd`.
Also `brew` itself provides an ability to configure background services. 

# Acceptance criteria
Brew formulas provide an easy way to run `tezos-node` and all the daemons in the background as a `launchd` or `brew` service.

Formulas should also create all default data directories during the post-install, similar to #133.

rvem:
## Description
Problem: We want to create default data directories and init node
configs during the post-install stage.

Solution: Add corresponding post_install to the formulas that provide
brew services.

Currently based and targets #149 

## Related issue(s)

Resolves #130 

  can be successfully built.

### DIFF
--- a/Formula/tezos-endorser-008-PtEdo2Zk.rb
+++ b/Formula/tezos-endorser-008-PtEdo2Zk.rb
@@ -122,4 +122,7 @@ class TezosEndorser008Ptedo2zk < Formula
       </plist>
     EOS
   end
+  def post_install
+    mkdir "#{var}/lib/tezos/client"
+  end
 end
--- a/Formula/tezos-endorser-008-PtEdo2Zk.rb
+++ b/Formula/tezos-endorser-008-PtEdo2Zk.rb
@@ -107,7 +107,7 @@ class TezosEndorser008Ptedo2zk < Formula
           <key>EnvironmentVariables</key>
             <dict>
               <key>DATA_DIR</key>
-              <string>#{ENV["HOME"]}/tezos/client</string>
+              <string>#{var}/lib/tezos/client</string>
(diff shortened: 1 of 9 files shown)


Formula/tezos-endorser-008-PtEdo2Zk.rb
```diff
@@ -122,4 +122,7 @@ def plist
       </plist>
     EOS
   end
+  def post_install
+    mkdir "#{var}/tezos/client"
```
pasqu4le, line 126: Missing a `lib` here

Formula/tezos-signer-tcp.rb
```diff
@@ -82,4 +82,7 @@ def plist
       </plist>
     EOS
   end
+  def post_install
+    mkdir "#{var}/lib/tezos/signer-https"
```
pasqu4le, line 86: Shouldn't it be `tcp` here?

rvem: Indeed, thanks

Formula/tezos-endorser-008-PtEdo2Zk.rb
rvem: Fixed

Formula/tezos-signer-tcp.rb
rvem: Fixed

pasqu4le (approved):

PR #153, opened 2021-03-04
Don't depend on rust brew formula

rvem:
## Description
Problem: Tezos requires rust version to be specified explicitly in case
no default suggested version (1.44.0) is used. Previously, we
depended on rust brew formulas, but the rust version in it can be
changed at any time, which means that our formulas are quite unstable.

Solution: Depend on rustup-init and fetch rust compiler via
`rustup-init --default-toolchain 1.44.0`.

## Related issue(s)

Resolves None

  can be successfully built.


pasqu4le (approved):
Only a nitpick in the commit message:
> explicitely in case no the default suggested version (1.44.0)

should probably be changed to something like:
```
explicitely in case a version different from the default one (1.44.0)
```

PR #154, opened 2021-03-11
[#148] tezos-node-<network> aliases

### LINKED ISSUE #148: Consider adding aliases for commands on Ubuntu

As part of the explanation for #125 it is apparent that there is the need to call several `tezos-*` commands wuth the `tezos` user in order to set up baking.

However in the case of `tezos-node`, many of these calls require the user to specify the `data-dir` to be used.

To improve UX I propose to add `tezos-node-<network>` during installation (to be used by the `tezos` user), which are just aliases to `tezos-node` with the correct data directory set.
These should make its use more straightforward as their name would match the one of the corresponding `systemd` services.
 
# Acceptance criteria

There should be aliases available to the `tezos` user to call `tezos-node` without having to specify the `--data-dir`.

rvem:
## Description
Problem: It would be convenient to have such aliases when it's required
to run node with the data directory specified in the systemd service
environment.

Solution: Add tezos-node-<network> scripts that are generated during
postinst stage. They are also cleaned up during the postrm.

## Related issue(s)

Resolves part of #148 (it'll be closed once the updated packages are published and the baking article is updated)

  can be successfully built.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,17 +205,22 @@ file, they can be found in [`scripts`](./docker/package/scripts) and
 It's possible to run multiple same services, e.g. two `tezos-node`s that run different
 networks.
 
-`tezos-node` packages provide three services out of the box:
-`tezos-node-delphinet` and `tezos-node-mainnet` that run
-`delphinet` and `mainnet` networks respectively.
+`tezos-node` packages provide multiple services out of the box:
+`tezos-node-edo2net` and `tezos-node-mainnet` that run
+`edo2net` and `mainnet` networks respectively.
 
 In order to start it run:
 ```
 systemctl start tezos-node-<network>
 ```
 
+Also, there are `tezos-node-<network>` binary aliases that are equivalent to
+```
+TEZOS_NODE_DIR="<DATA_DIR from tezos-node-<network>.service>" tezos-node
+```
+
 In addition to node services where the config is predefined to a specific network
-(e.g. `tezos-node-mainnet` or `tezos-node-delphinet`), it's possible to run `tezos-node-custom`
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -206,16 +206,21 @@ It's possible to run multiple same services, e.g. two `tezos-node`s that run dif
 networks.
 
 `tezos-node` packages provide three services out of the box:
```
pasqu4le, line 208: Well these are only two services, perhaps this is (and already was) outdated?

Unless you are also counting `tezos-node-custom` here as well?

docker/package/model.py
```diff
@@ -284,6 +290,17 @@ def gen_postinst(self, out):
...
+        postinst_contents = f'''#!/bin/sh
+
+set -e
+
+#DEBHELPER#
+
+{self.postrm_steps}
+'''
+        with open(out, 'w') as f:
+            f.write(postinst_contents)
```
pasqu4le, line 303: I guess it doesn't really make a difference, but shouldn't this be `postrm_contents` instead (same just above)?

pasqu4le (commented):
Only two nitpicks, but one more thing: these should be mentioned in the baking article.

rvem:
> these should be mentioned in the baking article

I propose to update it only after the updated packages are uploaded because otherwise, readers can be confused

pasqu4le:
> only after the updated packages are uploaded

Right, goof point. In this case I'd leave the issue open tho, so we don't forget that this still needs to be done.

README.md
pasqu4le: > provide multiple services
:smart:

pasqu4le (approved):

PR #159, opened 2021-03-15
[#107] Add nixos modules for tezos binaries

### LINKED ISSUE #107: Add nixos modules for tezos binaries

We're already using tezos-node nixos module for our infrastructure. However, it's currently located in a private repo,
it would be nice to share it.

Also, we're providing systemd services for `tezos-node` and daemons (accuser, baker, endorser) in the native packages. It'd be nice to have nixos modules for them as well.

# Acceptance criteria
This repo provides an easy way to run nixos modules with `tezos-node`, `tezos-{accuser, baker, endorser}-<proto>` and `tezos-signer`.

ixahedron:
## Description

Problem: we're providing systemd services for `tezos-node` and tezos daemons in the native packages, but there are no nixos modules for them.

Solution: added nixos modules for `tezos-node`, `tezos-signer`, `tezos-baker-<proto>`, `tezos-accuser-<proto>`, and `tezos-endorser-<proto>`.

## Related issue(s)

Resolves #107 

  can be successfully built.

### DIFF
--- a/nix/modules/common.nix
+++ b/nix/modules/common.nix
@@ -1,0 +1,98 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+{ lib, pkgs, ... }:
+
+with lib;
+rec {
+  sharedOptions = {
+
+    logVerbosity = mkOption {
+      type = types.enum [ "fatal" "error" "warning" "notice" "info" "debug" ];
+      default = "warning";
+      description = ''
+        Level of logs verbosity. Possible values are:
+        fatal, error, warn, notice, info or debug.
+      '';
+    };
+
+  };
+
+  daemonOptions = sharedOptions // {
+
+    baseProtocol = mkOption {
+      type = types.enum [ "009-PsFLoren"];
+      description = ''
+        Base protocol version,
+        only '009-PsFLoren' is supported.
+      '';
+      example = "009-PsFLoren";
+    };
+
+    rpcPort = mkOption {
+      type = types.int;
+      default = 8732;
+      example = 8732;
+      description = ''
+        Tezos node RPC port.
+      '';
+    };
+
+  };
+
+  genDaemonConfig = instancesCfg: service-name: service-pkgs: service-script:
+    mkIf (instancesCfg != {}) {
+      users = mkMerge (flip mapAttrsToList instancesCfg (node-name: node-cfg: genUsers node-name ));
+      systemd = mkMerge (flip mapAttrsToList instancesCfg (node-name: node-cfg:
+        let tezos-service = service-pkgs."${node-cfg.baseProtocol}";
+        in {
+          services."tezos-${node-name}-tezos-${service-name}" = genSystemdService node-name node-cfg service-name // rec {
+            bindsTo = [ "network.target" "tezos-${node-name}-tezos-node.service" ];
+            after = bindsTo;
+            path = with pkgs; [ curl ];
+            preStart =
+              ''
+                while ! _="$(curl --silent http://localhost:${toString node-cfg.rpcPort}/chains/main/blocks/head/)"; do
+                  echo "Trying to connect to tezos-node"
+                  sleep 1s
+                done
+
+                service_data_dir="$STATE_DIRECTORY/client/data"
+                mkdir -p "$service_data_dir"
+
+                # Generate or update service config file
+                if [[ ! -f "$service_data_dir/config" ]]; then
+                  ${tezos-service} -d "$service_data_dir" -E "http://localhost:${toString node-cfg.rpcPort}" \
+                  config init --output "$service_data_dir/config" >/dev/null 2>&1
+                else
+                  ${tezos-service} -d "$service_data_dir" -E "http://localhost:${toString node-cfg.rpcPort}" \
+                  config update >/dev/null 2>&1
+                fi
+              '';
+            script = service-script node-cfg;
+          };
+      }));
+    };
+
+  genUsers = node-name: {
+    groups."tezos-${node-name}" = { };
+    users."tezos-${node-name}" = { group = "tezos-${node-name}"; };
+  };
+
+  genSystemdService = node-name: node-cfg: service-name: {
+    wantedBy = [ "multi-user.target" ];
+    description = "Tezos ${service-name}";
+    environment = {
+      TEZOS_LOG = "* -> ${node-cfg.logVerbosity}";
+    };
+    serviceConfig = {
+      User = "tezos-${node-name}";
+      Group = "tezos-${node-name}";
+      StateDirectory = "tezos-${node-name}";
+      Restart = "always";
+      RestartSec = "10";
+    };
+  };
+
+}
--- a/nix/modules/tezos-accuser.nix
+++ b/nix/modules/tezos-accuser.nix
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+{config, lib, pkgs, ...}:
+
+with lib;
+
+let
+  tezos-accuser-pkgs = {
+    "009-PsFLoren" =
+      "${pkgs.ocamlPackages.tezos-baker-009-PsFLoren}/bin/tezos-baker-009-PsFLoren";
+  };
+  cfg = config.services.tezos-accuser;
+  common = import ./common.nix { inherit lib; inherit pkgs; };
+  instanceOptions = types.submodule ( {...} : {
+    options = common.daemonOptions // {
+
+      enable = mkEnableOption "Tezos accuser service";
+
+    };
+  });
+
+in {
+  options.services.tezos-accuser = {
+    instances = mkOption {
+      type = types.attrsOf instanceOptions;
+      description = "Configuration options";
+      default = {};
+    };
+  };
+  config =
+    let accuser-script = node-cfg: ''
+        ${tezos-accuser-pkgs.${node-cfg.baseProtocol}} -d "$STATE_DIRECTORY/client/data" \
+        -E "http://localhost:${toString node-cfg.rpcPort}" \
+        run with local node "$STATE_DIRECTORY/node/data" "$@"
+      '';
+    in common.genDaemonConfig cfg.instances "accuser" tezos-accuser-pkgs accuser-script;
+}
--- a/nix/modules/tezos-baker.nix
+++ b/nix/modules/tezos-baker.nix
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+{config, lib, pkgs, ...}:
+
+with lib;
+
+let
+  tezos-baker-pkgs = {
+    "009-PsFLoren" =
+      "${pkgs.ocamlPackages.tezos-baker-009-PsFLoren}/bin/tezos-baker-009-PsFLoren";
+  };
+  cfg = config.services.tezos-baker;
+  common = import ./common.nix { inherit lib; inherit pkgs; };
+  instanceOptions = types.submodule ( {...} : {
+    options = common.daemonOptions // {
+
+      enable = mkEnableOption "Tezos baker service";
+
+      endorserAccountAlias = mkOption {
+        type = types.str;
+        default = "";
+        description = ''
+          Alias for the tezos-endorser account.
+        '';
+      };
+
+    };
+  });
+
+in {
+  options.services.tezos-baker = {
+    instances = mkOption {
+      type = types.attrsOf instanceOptions;
+      description = "Configuration options";
+      default = {};
+    };
+  };
+  config =
+    let baker-script = node-cfg: ''
+        ${tezos-baker-pkgs.${node-cfg.baseProtocol}} -d "$STATE_DIRECTORY/client/data" \
+        -E "http://localhost:${toString node-cfg.rpcPort}" \
+        run with local node "$STATE_DIRECTORY/node/data" ${node-cfg.endorserAccountAlias}
+      '';
+    in common.genDaemonConfig cfg.instances "baker" tezos-baker-pkgs baker-script;
+}
--- a/nix/modules/tezos-endorser.nix
+++ b/nix/modules/tezos-endorser.nix
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+{config, lib, pkgs, ...}:
+
+with lib;
+
+let
+  tezos-endorser-pkgs = {
+    "009-PsFLoren" =
+      "${pkgs.ocamlPackages.tezos-baker-009-PsFLoren}/bin/tezos-baker-009-PsFLoren";
+  };
+  common = import ./common.nix { inherit lib; inherit pkgs; };
+  cfg = config.services.tezos-endorser;
+  instanceOptions = types.submodule ( {...} : {
+    options = common.daemonOptions // {
+
+      enable = mkEnableOption "Tezos endorser service";
+
+      bakerAccountAlias = mkOption {
+        type = types.str;
+        default = "";
+        description = ''
+          Alias for the tezos-baker account.
+        '';
+      };
+
+    };
+  });
+
+in {
+  options.services.tezos-endorser = {
+    instances = mkOption {
+      type = types.attrsOf instanceOptions;
+      description = "Configuration options";
+      default = {};
+    };
+  };
+  config =
+    let endorser-script = node-cfg: ''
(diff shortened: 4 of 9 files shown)


nix/modules/tezos-accuser.nix
```diff
@@ -0,0 +1,70 @@
...
+        Level of logs verbosity. Possible values are:
+        fatal, error, warn, notice, info or debug.
+      '';
+    };
+
+    baseProtocol = mkOption {
+      type = types.str;
+      description = ''
+        Base protocol version,
+        only '007-PsDELPH1' and '008-PtEdo2Zk' are supported.
```
kirelagin, line 42: `types.oneOf`?

nix/modules/tezos-baker.nix
```diff
@@ -0,0 +1,70 @@
...
+        Local name of the node, to support running multiple local instances.
+      '';
+    };
+
+    logVerbosity = mkOption {
+      type = types.str;
+      default = "warning";
+      description = ''
+        Level of logs verbosity. Possible values are:
+        fatal, error, warn, notice, info or debug.
```
kirelagin, line 34: `types.oneOf` as well

nix/modules/tezos-accuser.nix
```diff
@@ -0,0 +1,70 @@
...
+let
+  tezos-accuser-pkgs = {
+    "007-PsDELPH1" =
+      "${pkgs.tezos.tezos-accuser-007-PsDELPH1}/bin/tezos-accuser-007-PsDELPH1";
+    "008-PtEdo2Zk" =
+      "${pkgs.tezos.tezos-accuser-008-PtEdo2Zk}/bin/tezos-accuser-008-PtEdo2Zk";
+  };
+  cfg = config.services.tezos-accuser;
+in {
+  options.services.tezos-accuser = rec {
```
kirelagin, line 18: Does this really need to be `rec`? (Asking because recursive definitions of options are always sus.)

nix/modules/tezos-accuser.nix
```diff
@@ -0,0 +1,70 @@
...
+        only '007-PsDELPH1' and '008-PtEdo2Zk' are supported.
+      '';
+      example = "008-PtEdo2Zk";
+    };
+
+  };
+  config = mkIf cfg.enable {
+    users.groups."tezos-${cfg.node-name}" = { };
+    users.users."tezos-${cfg.node-name}" = { group = "tezos-${cfg.node-name}"; };
+    systemd.services."tezos-${cfg.node-name}-tezos-accuser" = rec {
```
kirelagin, line 51: Also not sure why this one is `rec`.

nix/modules/tezos-node.nix
```diff
@@ -0,0 +1,134 @@
...
+let
+  tezos-node-pkg = pkgs.tezos.binaries.tezos-node;
+  cfg = config.services.tezos-node;
+  genConfigCommand = historyMode: rpcPort: netPort: network: ''
+                      --data-dir "$node_data_dir" \
+                      --history-mode "${cfg.historyMode}" \
+                      --rpc-addr ":${toString cfg.rpcPort}" \
+                      --net-addr ":${toString cfg.netPort}" \
+                      --network "${cfg.network}"
+  '';
```
kirelagin, line 18: Kinda weird formatting.

nix/modules/tezos-node.nix
```diff
@@ -0,0 +1,134 @@
...
+      environment = {
+        TEZOS_LOG = "* -> ${cfg.logVerbosity}";
+      };
+      serviceConfig = {
+        User = "tezos-${cfg.node-name}";
+        StateDirectory = "tezos-${cfg.node-name}";
+        Restart = "always";
+        RestartSec = "10";
+      };
+      preStart =
```
kirelagin, line 106: Not sure why so much copy-paste in this option. You can do things like:

```nix
{

x = ''
    hello world
    blabla
  '' + lib.optionalString (cfg.nodeConfig != null) ''
    foo bar
  '';

}
```

Also, this whole custom `nodeConfig` story is slightly suspicious. The standard pattern is to use the module options to generate a config file, and then append something like `extraOptions` to the end. (In the case of JSON you can just make `extraOptions` an attrset, then generate an attrset from options, and finally merge the two attrsets, and then use `builtins.toJSON` to write your config file.)

kirelagin (changes requested):
Looks like some of the services have the same options. Why don’t you just share them? The easiest thing to do would be something like:

```nix
let
  sharedOptions = {
    # ...
  };
in {

options.services.foo = sharedOptions // {
  # ... service-specific options
};

}

File: nix/modules/tezos-accuser.nix [hunk 7]
```diff
@@ -0,0 +1,70 @@
...
+  cfg = config.services.tezos-accuser;
+in {
+  options.services.tezos-accuser = rec {
+    enable = mkEnableOption "Tezos accuser service";
+
+    node-name = mkOption {
+      type = types.str;
+      default = "edonet";
+      description = ''
+        Local name of the node, to support running multiple local instances.
```
balsoft, line 25: Umm, how is this supposed to work when you only have one optionset? (same with other modules)

E.g. someone imports the module, sets
```
services.tezos-accuser.node-name = "foobar";
```
How would they then add a second instance called `goodaz`? They can't just do
```
services.tezos-accuser.node-name = "goodaz";
```
because that would override the option and not magically create a second accuser instance.

What you probably want to do is move this out into a submodule and then the make the interface something like `services.tezos-accuser.instances.foobar = { <...> }` and `services.tezos-accuser.instances.goodaz = { <...> }` or drop this altogether and let the user handle this as for many other NixOS modules (e.g. using nixos-containers)

balsoft (changes requested):
Definitely needs a lot of copy-paste extraction. Probably move the common options out of modules into a separate file imported everywhere, and make systemd service generation a function in that same file.

nix/modules/tezos-node.nix
ixahedron: I'm not sure if what you suggest if a great idea: `tezos-node` does some magical config construction that I don't really want to touch directly for fear of overwriting something important. E.g. look at the default config:

```
{ "p2p":
    { "bootstrap-peers":
        [ "boot.tzbeta.net", "116.202.172.21", "95.216.45.62" ],
      "listen-addr": "[::]:9732" } }
```

Updating it through cli while leaving the option for overwriting with a full ready-made config seems a bit more reasonable. Am I wrong?

tests/tezos-modules.nix
```diff
@@ -0,0 +1,51 @@
...
+
+  testScript = ''
+    start_all()
+
+    machine.wait_for_unit("tezos-edonet-tezos-node.service")
+    machine.wait_for_unit("tezos-edonet-tezos-signer.service")
+    machine.wait_for_unit("tezos-edonet-tezos-accuser.service")
+    machine.wait_for_unit("tezos-edonet-tezos-baker.service")
+    machine.wait_for_unit("tezos-edonet-tezos-endorser.service")
+  '';
```
balsoft, line 50: Needs moar test IMO. This will pass even if the services break after a second of running. Would be nice to see some sanity checks here.

ixahedron: I'm not sure how to test more precisely than what I do in the latest version, as no service other than `tezos-node` really has an rpc interface. Hopefully just checking status is enough.

ixahedron:
@rvem I guess you're the person to discuss this with. AFAIU, in the current version of this PR, the various daemon systemd services aren't very useful, right? They just run on their own without really binding to `tezos-node`. From what I've read about the tezos "ecosystem", it would be better to e.g. bind `tezos-baker` to `tezos-node` and run it with `tezos-baker run with local node "$STATE_DIRECTORY" bakerkey` and so on, but that, as I understand, would require us to use `tezos-client` too. Is that what needs to be done, plus adding an option for `bakerkey` to `tezos-node` options config?
Or is the current way of running it what we need? If so, then I guess I don't really understand how all the services tie together :/

tests/tezos-modules.nix
```diff
@@ -0,0 +1,67 @@
...
+    for s in services:
+        machine.wait_for_unit(f"tezos-edonet-{s}.service")
+
+    machine.wait_for_open_port(8732)
+    machine.wait_until_succeeds(
+        "curl --silent http://localhost:8732/chains/main/blocks/head/header | grep level"
+    )
+
+    for s in services:
+        output = machine.succeed(f"systemctl status tezos-edonet-{s}.service | grep Active")
```
balsoft, line 65: I think this can just run `systemctl status` service, the exit code should be non-zero if the service fails

ixahedron: Actually, just `machine.succeed(f"systemctl status tezos-edonet-{s}.service")` inevitably fails with an outcome dependent on the order of names in the `services` list: usually with `Exception: command `systemctl status tezos-edonet-tezos-baker.service` failed (exit code 3)`, but sometimes it's this or any other service with exit code 2. The thing is, the trace also shows a perfectly normal `systemctl status` output for the offending service, and then fails for no discernible (to me) reason. E.g.:

```
machine: must succeed: systemctl status tezos-edonet-tezos-baker.service
machine: output: ● tezos-edonet-tezos-baker.service - Tezos baker
     Loaded: loaded (/nix/store/3lkpyl7bldfqrs6q0342xd764h7g538n-unit-tezos-edonet-tezos-baker.service/tezos-edonet-tezos-baker.service; enabled; vendor preset: enabled)
     Active: activating (auto-restart) since Mon 2021-03-29 05:04:32 UTC; 7s ago
    Process: 931 ExecStart=/nix/store/d5d9hrzv8ampx4c4rw6zpfvkyb2rjf35-unit-script-tezos-edonet-tezos-baker-start/bin/tezos-edonet-tezos-baker-start (code=exited, status=0/SUCCESS)
   Main PID: 931 (code=exited, status=0/SUCCESS)
         IP: 0B in, 0B out
        CPU: 373ms

Mar 29 05:04:32 machine systemd[1]: tezos-edonet-tezos-baker.service: Succeeded.
Hint: Some lines were ellipsized, use -l to show in full.

Test "service status sanity check" failed with error: "command `systemctl status tezos-edonet-tezos-baker.service` failed (exit code 3)"
```

Any systemctl status run on its own (e.g. just `machine.succeed("systemctl status tezos-edonet-tezos-baker.service")`) never fails, as far as I can tell. Any idea what the problem could be?

kirelagin: I don’t know why it fails, but nevertheless I want to share a piece of knowledge that I happen to possess: there is `systemctl is-active <unit>` 🤯.

kirelagin: I think this actually makes sense. In your output the unit is `activating`, i.e. it is not actually running yet, thus `systemctl status` exits with a non-zero exit code.

If you want to make sure that it has actually successfully started, you’ll have to wait for it.

Also see this: https://unix.stackexchange.com/questions/385454/systemd-check-that-a-service-is-not-running. There is a difference between `systemctl status` and `systemctl is-active`, and they also mention `systemctl show -p ActiveState` – make sure you pick the right one.

kirelagin: Also, as a matter of fact, `Active: activating (auto-restart) ` looks pretty sus 🤨. Is the service crashing and gets auto-restarted? :)

kirelagin: _Oh, ok, I actually read the discussion above. Looks like the tests caught an actual bug? :)_

balsoft:
Oh, just in case it isn't clear: internet is not accessible from inside NixOS tests, may this be an issue?

nix/binaries
```diff
@@ -0,0 +1 @@
+/nix/store/fza34lq5gv96gpya68ddf19w89w1jkgm-binaries
```
balsoft, line 1: u sure?

tests/tezos-modules.nix
ixahedron: Less of a bug than something I didn't manage to get an answer to in time from @rvem. But all good now, I think, found the culprit :tada:

nix/binaries
ixahedron: Opsolya

nix/modules/tezos-accuser.nix
```diff
@@ -33,5 +33,10 @@ in {
       default = {};
     };
   };
-  config = common.genDaemonConfig cfg.instances "accuser" tezos-accuser-pkgs;
+  config =
```
balsoft, line 36: So much boilerplate!

ixahedron: OK, how to improve? :thinking:

rvem:
> From what I've read about the tezos "ecosystem", it would be better to e.g. bind tezos-baker to tezos-node and run it with tezos-baker run with local node "$STATE_DIRECTORY" bakerkey and so on, but that, as I understand, would require us to use tezos-client too. Is that what needs to be done, plus adding an option for bakerkey to tezos-node options config?

Ideally, `tezos-baker` service should require `tezos-node` service as a part of `Requires` property. Though, there is a situation in which baker can be used within two different networks (mainnet and corresponding testnet). Since `tezos-node` services have more-or-less fixed name `tezos-${node-name}-tezos-node`, I propose adding `node-name` to the service options to define a `Requires` dependency and also AFAIU it should be possible to automatically infer node data directory that needs to be used in the `tezos-baker run with local node` command

ixahedron:
> Ideally, `tezos-baker` service should require `tezos-node` service as a part of `Requires` property. Though, there is a situation in which baker can be used within two different networks (mainnet and corresponding testnet). Since `tezos-node` services have more-or-less fixed name `tezos-${node-name}-tezos-node`, I propose adding `node-name` to the service options to define a `Requires` dependency

Wait, but can't the end user just coordinate what `node-name` they define for `tezos-baker` and `tezos-node`? So they would just write e.g. 

```
services.tezos-node.instances.edonet.enable = true;
services.tezos-baker.instances.edonet = ...
```

Or do you mean that just one `tezos-baker` instance, e.g. `edonet`, might be used with both `services.tezos-node.instances.mainnet` and `services.tezos-node.instances.testnet`? So your proposed `node-name` should be a list?

rvem:
> services.tezos-node.instances.edonet.enable = true;
services.tezos-baker.instances.edonet = ...

oh, that sounds good as well

nix/modules/common.nix
```diff
@@ -0,0 +1,93 @@
...
+      users = mkMerge (flip mapAttrsToList instancesCfg (node-name: node-cfg: genUsers node-name ));
+      systemd = mkMerge (flip mapAttrsToList instancesCfg (node-name: node-cfg:
+        let tezos-service = service-pkgs."${node-cfg.baseProtocol}";
+        in {
+          services."tezos-${node-name}-tezos-${service-name}" = genSystemdService node-name node-cfg service-name // rec {
+            bindsTo = [ "network.target" "tezos-${node-name}-tezos-node.service" ];
+            after = bindsTo;
+            preStart =
+              ''
+                sleep 10 # to allow tezos-node.service to successfully start
```
rvem, line 55: Looks suspicious, 10s can be not enough in case the node is generating an identity, let's use curl to check whether the node is up, see [this](https://gitlab.com/morley-framework/local-chain/-/blob/master/run-local-chain.sh#L54)

nix/modules/tezos-accuser.nix
```diff
@@ -0,0 +1,42 @@
...
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+{config, lib, pkgs, ...}:
+
+with lib;
+
+let
+  tezos-accuser-pkgs = {
+    "007-PsDELPH1" =
```
rvem, line 11: Btw, we've dropped 007 support some time ago, let's drop it in these services as well :wink:

nix/modules/tezos-endorser.nix
```diff
@@ -0,0 +1,42 @@
...
+      "${pkgs.ocamlPackages.tezos-endorser-007-PsDELPH1}/bin/tezos-endorser-007-PsDELPH1";
+    "008-PtEdo2Zk" =
+      "${pkgs.ocamlPackages.tezos-endorser-008-PtEdo2Zk}/bin/tezos-endorser-008-PtEdo2Zk";
+    "009-PsFLoren" =
+      "${pkgs.ocamlPackages.tezos-baker-009-PsFLoren}/bin/tezos-baker-009-PsFLoren";
+  };
+  common = import ./common.nix { inherit lib; };
+  cfg = config.services.tezos-endorser;
+  instanceOptions = types.submodule ( {...} : {
+    options = common.daemonOptions // {
```
rvem, line 21: `tezos-endorser` and `tezos-baker` also should have an option for setting up baker/endorser account alias and use this alias in `tezos-endorser ... run ${cfg.bakerAccountAlias}` or `tezos-baker ... run with local node "$STATE_DIRECTORY/node/data"  ${cfg.endorserAccountAlias}`

rvem (commented):
Sorry for such a long response.
Looks good, have a bunch of minor comments about services functionality

rvem (approved):
LGTM now

balsoft (approved):

nix/modules/tezos-node.nix
```diff
@@ -0,0 +1,119 @@
...
+          Network which node will be running on.
+        '';
+      };
+
+      historyMode = mkOption {
+        type = types.str;
+        default = "full";
+        description = ''
+          Node history mode. Possible values are:
+          full, experimental-rolling or arcive.
```
pasqu4le, line 61: Should this use `types.enum` too?

Also `arcive` looks like a typo, should be `archive`.

pasqu4le (approved):
Only a minor comment, the rest LGTM

PR #160, opened 2021-03-17
[#148] Use new aliases in baking article

### LINKED ISSUE #148: Consider adding aliases for commands on Ubuntu

rvem:
## Description
Problem: Providing data-dir manually to the `tezos-node` commands is a
bit inconvenient.

Solution: Suggest using tezos-node-<network> aliases instead of
providing data directory option manually.

## Related issue(s)

Resolves #148 

  can be successfully built.

### DIFF
--- a/docs/baking.md
+++ b/docs/baking.md
@@ -96,31 +96,14 @@ the smallest and the fastest mode that is sufficient for baking (you can read mo
 
 All commands within the service are run under the `tezos` user.
 
+The `tezos-node` package provides `tezos-node-<network>` aliases that are equivalent to
+`tezos-node --data-dir <DATA_DIR from the tezos-node-<network>.service>`.
+These aliases can be used instead of providing `--data-dir` option to the `tezos-node`
+invocations manually.
+
 In order to import the snapshot, run the following command:
 ```
-sudo -u tezos tezos-node snapshot import --data-dir /var/lib/tezos/node-<network>/ <path to the snapshot file>
-```
-
-#### Troubleshooting node snapshot import
-
-Snapshot import requires that the `tezos-node` data directory doesn't contain the `context` and `store` folders.
-Remove them if you're getting an error about their presence.
-
-In case you're getting an error similar to:
-```
-tezos-node: Error:
-              Invalid block BLjutMj47caB
-                Failed to validate the economic-protocol content of the block: Error:
-                                                                                Invalid signature for block BLjutMj47caB. Expected: tz1Na5QB98cDA.
-```
-
-you should init/update config in the data directory to match the desired network
-before importing the snapshot. To do that, run one of the following commands
-(depending on whether the config in the given data directory was initialized previously):
-```
-sudo -u tezos tezos-node config init --data-dir /var/lib/tezos/node-<network> --network <network>
-#or
-sudo -u tezos tezos-node config update --data-dir /var/lib/tezos/node-<network> --network <network>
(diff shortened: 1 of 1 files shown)


docs/baking.md
```diff
@@ -72,9 +71,14 @@ the smallest and the fastest mode that is sufficient for baking (you can read mo
 
 All commands within the service are run under the `tezos` user.
 
+`tezos-node` package provides `tezos-node-<network>` aliases that are equivalent to
+`tezos-node --data-dir <DATA_DIR from the tezos-node-network.service>`.
```
rvem, line 75:
```suggestion
`tezos-node --data-dir <DATA_DIR from the tezos-node-<network>.service>`.
```

docs/baking.md
```diff
@@ -94,9 +98,9 @@ you should init/update config in the data directory to match the desired network
 before importing the snapshot. To do that, run one of the following commands
 (depending on whether the config in the given data directory was initialized previously):
 ```
-sudo -u tezos tezos-node config init --data-dir /var/lib/tezos/node-<network> --network <network>
+sudo -u tezos tezos-node-<network> config init --network <network>
```
itkach, line 101: from user's perspective --netwrok is redundant here... can alias also include `--network`? alternatively, perhaps tezos-node-<network> should be a script?

rvem: Indeed, including `--network` to the `tezos-node-<networks>` scripts sound like a good way to go

pasqu4le: Actually I think we should rather remove this snapshot troubleshooting section entirely.

We now call `config init` as part of the package installation process, so the problem that could have caused this is no longer there if one follows the guide.

It can still happen if a user decides to manually recreate/alter the node data dir, but those are not the kind of problems that packaging should solve IMO.

So, the `--network` option is not needed by other `tezos-node` commands as the `network` is already specified in the config file **and** we already generate the correct config for each network by default.
I think it's fair to say that we can avoid complicating the scripts to support solving a problem that we now prevent.

docs/baking.md
```diff
@@ -96,31 +95,14 @@ the smallest and the fastest mode that is sufficient for baking (you can read mo
...
-In order to import the snapshot, run the following command:
-```
-sudo -u tezos tezos-node snapshot import --data-dir /var/lib/tezos/node-<network>/ <path to the snapshot file>
-```
-
-#### Troubleshooting node snapshot import
-
-Snapshot import requires that the `tezos-node` data directory doesn't contain the `context` and `store` folders.
-Remove them if you're getting an error about their presence.
+`tezos-node` package provides `tezos-node-<network>` aliases that are equivalent to
```
pasqu4le, line 98:
```suggestion
The `tezos-node` package provides `tezos-node-<network>` aliases that are equivalent to
```

pasqu4le (approved):
A minor nitpick, but otherwise LGTM

Issue #156, opened 2021-03-11
Provide network specific systemd pseudo-services for baking


itkach:
# Clarification and motivation

Tezos users may want to set up baking on a particular network, such as mainnet or a test network for current or upcoming protocol. This involves configuring and starting a set of daemons - baker, endorser and, optionally, accuser,  - for the network's current protocol, and possibly, in case of protocol upgradeable network such as mainnet,  the same set of daemons for the upcoming protocol. Providing individual baker, endorser and accuser services per protocol allows to achieve this, but is not convenient. Also, there are some configuration and inter-service hard and soft dependencies that are not codified: baker service needs to be configured to point to node's data directory, baker and endorser services should be running (or not running) at the same time, node service should be running when baker and endorser are running etc.

A systemd service can be provided to control these several cooperating daemons together, along the lines of what is described in https://alesnosek.com/blog/2016/12/04/controlling-a-multi-service-application-with-systemd/. Modifying this service definition and rolling out these modifications as a regular package update will allow to automatically prepare running bakers for an upcoming protocol upgrade.

# Acceptance criteria

User can start baking on a known network with a single command like so:

```
$ sudo systemctl start tezos-baking-mainnet
```

provided user's private key has been imported and designated for baking.

If system is missing pre-requisits, the command may fail and provide further user instructions, or perhaps even launch configuration script that would guide user through private key import and baker registration, wrapping corresponding tezos-client comands and modifying config files as necessary.

pasqu4le:
Fixed in v9.1-1 native packages

PR #164, opened 2021-03-24
[#156] Baking services

rvem:
## Description
This PR provides a package with `tezos-baking-<network>` services so that users will be able to do
```
sudo systemctl start tezos-baking-{mainnet, edo2net}
```
This service will trigger `tezos-node`, `tezos-baker`, and `tezos-endorser` start.

To test locally on Ubuntu 20.04:
```
sudo add-apt-repository ppa:rvem/tezos-experiments-1
sudo apt-get update
sudo apt-get install tezos-baking
```

[Import baking key](https://github.com/serokell/tezos-packaging/blob/master/docs/baking.md#importing-the-baker-key) with `baker` alias.

[Bootstrap the node](https://github.com/serokell/tezos-packaging/blob/master/docs/baking.md#bootstrapping-the-node).

Once everything is done run:
```
systemctl start tezos-baking-edo2net.service
```

Check that everything else is running now:
```
systemctl status tezos-node-edo2net.service
systemctl status tezos-baker-008-ptedo2zk@edo2net.service
systemctl status tezos-endorser-008-ptedo2zk@edo2net.service
```
Or:
```
journalctl -f _UID=$(id tezos -u)
```
To stop the baking instance run:
```
systemctl stop tezos-baking-edo2net.service
```

## Related issue(s)

Resolves part of #156

  can be successfully built.


itkach:
Looks like baker and endorser fail to start because node is not immediately up and they don't handle this gracefully:

```
root@tez4:~# journalctl -f -u tezos-baker-008-ptedo2zk@edo2net.service
-- Logs begin at Fri 2021-03-26 13:21:03 EDT. --
Mar 26 14:20:49 tez4 systemd[1]: Started Instantiated tezos baker daemon service.
Mar 26 14:20:50 tez4 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Main process exited, code=exited, status=1/FAILURE
Mar 26 14:20:50 tez4 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Failed with result 'exit-code'.
```

```
root@tez4:~# journalctl -f -u tezos-endorser-008-ptedo2zk@edo2net.service
-- Logs begin at Fri 2021-03-26 13:21:03 EDT. --
Mar 26 14:20:53 tez4 tezos-endorser-start[3128]: Connection refused, retrying in 2.25 seconds...
Mar 26 14:20:55 tez4 tezos-endorser-start[3128]: Connection refused, retrying in 3.38 seconds...
Mar 26 14:20:58 tez4 tezos-endorser-start[3128]: Connection refused, retrying in 5.06 seconds...
Mar 26 14:21:03 tez4 tezos-endorser-start[3128]: Error:
Mar 26 14:21:03 tez4 tezos-endorser-start[3128]:   Rpc request failed:
Mar 26 14:21:03 tez4 tezos-endorser-start[3128]:      - meth: GET
Mar 26 14:21:03 tez4 tezos-endorser-start[3128]:      - uri: http://localhost:8732/monitor/bootstrapped
Mar 26 14:21:03 tez4 tezos-endorser-start[3128]:      - error: Unable to connect to the node: "Unix.Unix_error(Unix.ECONNREFUSED, "connect", "")"
Mar 26 14:21:03 tez4 systemd[1]: tezos-endorser-008-ptedo2zk@edo2net.service: Main process exited, code=exited, status=1/FAILURE
Mar 26 14:21:03 tez4 systemd[1]: tezos-endorser-008-ptedo2zk@edo2net.service: Failed with result 'exit-code'.
```

Both come up fine after 
```
systemctl restart tezos-baker-008-ptedo2zk@edo2net.service
systemctl restart tezos-endorser-008-ptedo2zk@edo2net.service
```

few minutes later

itkach:
I also wonder if there's a way to move setting `BAKER_ACCOUNT` from protocol-specific configs (like /etc/default/tezos-baker-008-ptedo2zk) up to network-specific definitions? The way this is set up currently, baking address alias needs to appear in 2 to 4 places - baker and endorser configs for current or current and upcoming protocols. Also, it seems likely that users may want to use different address for different networks. And it would make it easier (I think) to update tezos-baking-<network> services for upcoming protocol switchover

itkach:
Hopefully specifying baking account in multiple, protocol-specific places can go away altogether, but maybe still worth pointing out that ENDORSER_ACCOUNT env variable used in /etc/default/tezos-endorser-008-ptedo2zk should probably really be BAKER_ACCOUNT.

rvem:
> Looks like baker and endorser fail to start because node is not immediately up and they don't handle this gracefully

Indeed, I've moved identity generation to the `ExecStartPre` step of the node service, thus identity is generated before the node and daemons services are started.

rvem:
> I also wonder if there's a way to move setting BAKER_ACCOUNT from protocol-specific configs (like /etc/default/tezos-baker-008-ptedo2zk) up to network-specific definitions?

That makes sense, I've moved the baker and endorser aliases definition to the network-specific configs, e.g. `/etc/default/tezos-baking-edo2net`. Also by default baker account alias is used as an endorser account alias as well.

rvem:
@itkach, I've also updated packages in my private PPA, they should become available in ~30 minutes, so you're welcome to play with them once again:)

itkach:
> Also by default baker account alias is used as an endorser account alias as well.
Looks like that didn't quite work, "$BAKER_ACCOUNT" didn't resolve to actual value and endorser failed to start:

```
Apr 12 16:47:44 tez2 systemd[1]: Started Instantiated tezos endorser daemon service.
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]: Error:
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   Erroneous command line argument 2 ($BAKER_ACCOUNT).
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   no public key hash alias named $BAKER_ACCOUNT
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]: Usage:
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   tezos-endorser-008-PtEdo2Zk [global options] command >
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   tezos-endorser-008-PtEdo2Zk --help (for global option>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   tezos-endorser-008-PtEdo2Zk [global options] command >
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]: To browse the documentation:
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   tezos-endorser-008-PtEdo2Zk [global options] man (for>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   tezos-endorser-008-PtEdo2Zk [global options] man -v 3>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]: Global options (must come before the command):
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -d --base-dir <path>: client data directory
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -c --config-file <path>: configuration file
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -t --timings: show RPC request times
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   --chain <hash|tag>: chain on which to apply contextua>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -b --block <hash|tag>: block on which to apply contex>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -w --wait <none|<int>>: how many confirmation blocks >
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -p --protocol <hash>: use commands of a specific prot>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -l --log-requests: log all requests to the node
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -A --addr <IP addr|host>: [DEPRECATED: use --endpoint>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -P --port <number>: [DEPRECATED: use --endpoint inste>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -S --tls: [DEPRECATED: use --endpoint instead] use TL>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -E --endpoint <uri>: HTTP(S) endpoint of the node RPC>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -R --remote-signer <uri>: URI of the remote signer
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -f --password-filename <filename>: path to the passwo>
Apr 12 16:47:45 tez2 tezos-endorser-start[1606]:   -M --mode <client|mockup>: how to interact with the n>
Apr 12 16:47:45 tez2 systemd[1]: tezos-endorser-008-ptedo2zk@edo2net.service: Main process exited, code=>
Apr 12 16:47:45 tez2 systemd[1]: tezos-endorser-008-ptedo2zk@edo2net.service: Failed with result 'exit-c>
```

my `/etc/default/tezos-baking-edo2net`:
```
BAKER_ACCOUNT="mybaker"
ENDORSER_ACCOUNT="$BAKER_ACCOUNT"
```

itkach:
I wonder if its a bad idea to just put in a predefined alias, e.g. `baker` and instruct users to import their baking key under this alias? One less thing for user to configure and commands in HOWTO can become even more copy paste, e.g.:
```
sudo -u tezos tezos-client import secret key baker $SECRET_KEY_OR_LEDGER_URI
sudo -u tezos tezos-client register key baker as delegate
```

itkach:
looks like all systemctl/journalctl commands can take shorter service name without `.service`

itkach:
Not sure what went wrong but, it looks like tezos-node-mainnet started automatically and was running at the same time as tezos-node-edo2net... 
```
ubuntu@tez2:~$ tezos-client list known protocols
Disclaimer:
  The  Tezos  network  is  a  new  blockchain technology.
  Users are  solely responsible  for any risks associated
  with usage of the Tezos network.  Users should do their
  own  research to determine  if Tezos is the appropriate
  platform for their needs and should apply judgement and
  care in their network interactions.

Fatal error: unknown protocol version.
ubuntu@tez2:~$ curl -v http://localhost:8732/version
*   Trying 127.0.0.1:8732...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8732 (#0)
> GET /version HTTP/1.1
> Host: localhost:8732
> User-Agent: curl/7.68.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-length: 269
< content-type: application/json
<
{"version":{"major":8,"minor":2,"additional_info":"release"},"network_version":{"chain_name":"TEZOS_MAINNET","distributed_db_version":1,"p2p_version":1},"commit_info":{"commit_hash":"6102c808a21b32e732ab9bb1825761cd056f3e86","commit_date":"2021-02-10 22:57:06 +0100"}}
* Connection #0 to host localhost left intact
```

but eventuall it crashed:

```
 fetching headers from peer idqfgLUURwxuuemmPtp8xvL1LM5gkG: 77000/132930
Apr 12 17:03:12 tez2 tezos-node-start[922]: Apr 12 17:03:12.574 - node.validator.bootstrap_pipeline: still fetching headers from peer idqfgLUURwxuuemmPtp8xvL1LM5gkG: 78000/132930
Apr 12 17:03:12 tez2 tezos-node-start[922]: Apr 12 17:03:12.582 - node.validator.bootstrap_pipeline: still fetching headers from peer idqfgLUURwxuuemmPtp8xvL1LM5gkG: 79000/132930
Apr 12 17:03:12 tez2 tezos-node-start[922]: Apr 12 17:03:12.590 - node.validator.bootstrap_pipeline: still fetching headers from peer idqfgLUURwxuuemmPtp8xvL1LM5gkG: 80000/132930
Apr 12 17:03:12 tez2 tezos-node-start[922]: Apr 12 17:03:12.597 - node.validator.bootstrap_pipeline: still fetching headers from peer idqfgLUURwxuuemmPtp8xvL1LM5gkG: 81000/132930
Apr 12 17:03:20 tez2 tezos-node-start[922]: Apr 12 17:03:20.788 - node.validator.bootstrap_pipeline: still fetching headers from peer idqeUCDhuqWv4iWuWUZ7eh6HtVBYbn: 71000/174242
Apr 12 17:06:18 tez2 systemd[1]: tezos-node-mainnet.service: Main process exited, code=killed, status=9/KILL
Apr 12 17:06:18 tez2 tezos-node-start[957]: (/usr/bin/tezos-node) Exit: exit because of uncaught exception: Unix.Unix_error(Unix.EPIPE, "write", "")
Apr 12 17:06:18 tez2 systemd[1]: tezos-node-mainnet.service: Failed with result 'signal'.
```

tezos-node-edo2net is running, but no rpc on port 8732. I fixed /etc/default/tezos-baking-edo2net - changed `ENDORSER_ACCOUNT="$BAKER_ACCOUNT"` to `ENDORSER_ACCOUNT="mybaker"` , restarted with `sudo systemctl restart tezos-baking-edo2net` and now everything looks good.

rvem:
> Looks like that didn't quite work, "$BAKER_ACCOUNT" didn't resolve to actual value and endorser failed to start

That's unfortunate :unamused: 

> looks like all systemctl/journalctl commands can take shorter service name without `.service`

That's true, I'm just used to using full services names:)

> Do we actually need two variables? Does it ever make sense for ENDORSER_ACCOUNT to be different from BAKER_ACCOUNT? Why allow such a possibility?

I'm not sure whether it makes sense to have both `ENDORSER_ACCOUNT` and `BAKER_ACCOUNT`. On one hand, it's possible to run an endorser without the specified account and local node (don't know if it can be useful). Perhaps, it's worth renaming both `ENDORSER_ACCOUNT` and `BAKER_ACCOUNT` to something like `USER_ACCOUNT` to avoid confusion that endorser daemon requires baker account, what do you think?

> Not sure what went wrong but, it looks like tezos-node-mainnet started automatically and was running at the same time as tezos-node-edo2net...

Hmm, that sounds weird, services aren't supposed to be started right after the package installation. Also, there is no dependency between `tezos-baking-edo2net` and `tezos-node-mainnet`. Let me know in case you'll reproduce this behaviour

rvem:
> I wonder if its a bad idea to just put in a predefined alias, e.g. baker and instruct users to import their baking key under this alias?

Sounds good

itkach:
Testing the latest version, node failed to come up initially:

```
ubuntu@tez2:~$ sudo journalctl -u tezos-node-edo2net.service 
-- Logs begin at Wed 2021-04-14 07:54:42 EDT, end at Wed 2021-04-14 08:24:37 EDT. --
Apr 14 08:20:32 tez2 systemd[1]: Starting Tezos node edo2net...
Apr 14 08:22:02 tez2 systemd[1]: tezos-node-edo2net.service: start-pre operation timed out. Terminating.
Apr 14 08:22:02 tez2 systemd[1]: tezos-node-edo2net.service: Control process exited, code=killed, status=15/TERM
Apr 14 08:22:02 tez2 tezos-node-prestart[3230]: [12.3K blob data]
Apr 14 08:22:02 tez2 systemd[1]: tezos-node-edo2net.service: Failed with result 'timeout'.
Apr 14 08:22:02 tez2 systemd[1]: Failed to start Tezos node edo2net.
ubuntu@tez2:~$ sudo journalctl -u tezos-baker-008-ptedo2zk@edo2net.service
-- Logs begin at Wed 2021-04-14 07:54:42 EDT, end at Wed 2021-04-14 08:25:20 EDT. --
Apr 14 08:22:02 tez2 systemd[1]: Dependency failed for Instantiated tezos baker daemon service.
Apr 14 08:22:02 tez2 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Job tezos-baker-008-ptedo2zk@edo2net.service/start failed with result 'dependency'.
ubuntu@tez2:~$ sudo journalctl -u tezos-endorser-008-ptedo2zk@edo2net.service
-- Logs begin at Wed 2021-04-14 07:54:42 EDT, end at Wed 2021-04-14 08:25:29 EDT. --
Apr 14 08:22:02 tez2 systemd[1]: Dependency failed for Instantiated tezos endorser daemon service.
Apr 14 08:22:02 tez2 systemd[1]: tezos-endorser-008-ptedo2zk@edo2net.service: Job tezos-endorser-008-ptedo2zk@edo2net.service/start failed with result 'dependency'.
```

After running `sudo systemctl start tezos-baking-edo2net`, node and endorser came up, but baker did not:

```
ubuntu@tez2:~$ sudo systemctl status tezos-baker-008-ptedo2zk@edo2net
● tezos-baker-008-ptedo2zk@edo2net.service - Instantiated tezos baker daemon service
     Loaded: loaded (/lib/systemd/system/tezos-baker-008-ptedo2zk@.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Wed 2021-04-14 08:27:11 EDT; 8min ago
    Process: 3466 ExecStart=/usr/bin/tezos-baker-start (code=exited, status=1/FAILURE)
   Main PID: 3466 (code=exited, status=1/FAILURE)

Apr 14 08:27:10 tez2 systemd[1]: Started Instantiated tezos baker daemon service.
Apr 14 08:27:11 tez2 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Main process exited, code=exited, status=1/FAILURE
Apr 14 08:27:11 tez2 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Failed with result 'exit-code'.
```

and then finally after another `sudo systemctl restart tezos-baking-edo2net` all three services are up and running

itkach:
> it's possible to run an endorser without the specified account and local node (don't know if it can be useful)

as far as I know that's not useful at all, at least for a regular user :)

> something like `USER_ACCOUNT` to avoid confusion 

Word "user" in this context (tezos, but also unix, daemons, systemd) is too overloaded. I like `BAKER_ADDRESS_ALIAS`, I think that's accurate, even if a bit long

> to avoid confusion that endorser daemon requires baker account

but it does! let's not be shy about it :)

rvem:
> Apr 14 08:22:02 tez2 systemd[1]: tezos-node-edo2net.service: Failed with result 'timeout'.

Hmm, looks like identity creation takes more time than default `ExecStartPre` timeout :unamused:

itkach:
I think it's great that now the entire setup becomes just this:

```
tznet="edo2net"
secret_key="..."

sudo add-apt-repository ppa:rvem/tezos-experiments-1
sudo apt-get update
sudo apt-get install tezos-baking

sudo -u tezos tezos-client import secret key baker $secret_key
snapshot_file=/tmp/tezos-${tznet}.rolling
wget https://${tznet}.xtz-shots.io/rolling -O $snapshot_file
sudo -u tezos tezos-node-${tznet} snapshot import $snapshot_file

sudo systemctl start tezos-baking-$tznet
sudo systemctl enable tezos-baking-$tznet

```
(assuming initial startup/dependencies/timing can be made robust)

One small drawback is that there's no obvious way to see/follow all the logs at the same time - ```sudo journalctl -u tezos-baking-edo2net``` is not very useful‚ and 

```sudo journalctl -f -u tezos-endorser-008-ptedo2zk@edo2net -u tezos-baker-008-ptedo2zk@edo2net -u tezos-node-edo2net```

is too tedious to type, requires too much knowledge of implementation and these are not even all the services, generally speaking (there should also be an instance of baker and endorser for the upcoming protocol, and perhaps we'll also run accuser one day)

Maybe listing logs by use is an acceptable workaround,

```
journalctl -f _UID=$(id tezos -u)
```

this (or is there a better one better) is probably worth documenting in the baking howto

**Comment by rvem** [comment]
Okay, I've increased a timeout for tezos-node service prestart script, now it should be enough to avoid timeout failure

**Comment by rvem** [comment]
Updated v8.3-based packages were published to my private PPA once again:)

File: docker/package/scripts/tezos-node-prestart [hunk 1]
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/bash
```
pasqu4le, line 1:
```suggestion
#!/usr/bin/env bash
```

docker/package/defaults/tezos-baking.conf
```diff
@@ -0,0 +1,9 @@
+# vim: ft=sh
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+# shellcheck disable=SC2034
+DATA_DIR="/var/lib/tezos/.tezos-client"
+NODE_RPC_ENDPOINT="http://localhost:8732"
+BAKER_ADDRESS_ALIAS="baker"
```
pasqu4le, line 9: Two questions:

1. how come the `baker` alias is predefined here but not in `tezos-baker.conf` and `tezos-endorser.conf`?
    I'd imagine the same consideration apply for running single services and we can provide a default there too.
2. Is this file necessary? AFAIK the meta-service runs the two services, which in turn use the configs mentioned above, wouldn't it be sufficient to have those?

docker/package/scripts/tezos-node-prestart
rvem: copy-paste is evil :unamused:

docker/package/defaults/tezos-baking.conf
rvem: > Is this file necessary?

Yes, it's used as an `EnvironmentFile` in `tezos-{baker, endorser}-<proto>@<network>` services
This config basically makes it possible to change `BAKER_ADDRESS_ALIAS` in one place and both baker and endorser services will use updated alias

> how come the baker alias is predefined here but not in tezos-baker.conf and tezos-endorser.conf?

Indeed, it probably makes sense to update them as well

rvem: So, basically, now there are two services for each Tezos daemon:
1) Simple `tezos-{baker, endorser, accuser}-<proto>.service` that acts like previously and doesn't know anything about `tezos-baking-<network>.service`
2) Instantiated `tezos-{baker, endorser, accuser}-<proto>@.service` which is used as a part of `tezos-baking-<network>.service`, see [this](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Service%20Templates)

pasqu4le: :ok_hand: I see, thanks.

> Indeed, it probably makes sense to update them as well

Ok, please do that, aside from that point the rest LGTM

pasqu4le (approved):

itkach:
node and endorser come up, but baker still fails

```
ubuntu@tez2:~$ systemctl status tezos-baker-008-ptedo2zk@edo2net.service
● tezos-baker-008-ptedo2zk@edo2net.service - Instantiated tezos baker daemon service
     Loaded: loaded (/lib/systemd/system/tezos-baker-008-ptedo2zk@.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Fri 2021-04-16 10:22:27 EDT; 57s ago
    Process: 3268 ExecStart=/usr/bin/tezos-baker-start (code=exited, status=1/FAILURE)
   Main PID: 3268 (code=exited, status=1/FAILURE)

Apr 16 10:22:26 tez2 systemd[1]: Started Instantiated tezos baker daemon service.
Apr 16 10:22:27 tez2 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Main process exited, code=exited, status=1/FAILURE
Apr 16 10:22:27 tez2 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Failed with result 'exit-code'.
```

itkach:
@rvem  btw, I got 8.2 tezos-client even though tezos-baking is at 8.3 and so are other packages :

```
~$ dpkg -s tezos-baking
...
Version: 2:8.3-0ubuntu2~focal
```

```
ubuntu@tez2:~$ dpkg -s tezos-client
...
Version: 2:8.2-0ubuntu3~focal
```

rvem:
> I got 8.2 tezos-client even though tezos-baking is at 8.3

That's okay, I haven't updated tezos-client in my private PPA

rvem:
> but baker still fails

Could you please check `launchctl` to find the reason of failure?

itkach:
> > but baker still fails
> 
> Could you please check `launchctl` to find the reason of failure?

how do you mean? `launchctl` would be on mac...

itkach:
@rvem it's super quick to try this in a vm with [multipass](https://multipass.run/)

```
multipass launch -c 1 -d 20G -m 4G --name tez1
multipass shell tez1
``` 
and just run [steps from my comment above](https://github.com/serokell/tezos-packaging/pull/164#issuecomment-819502629) - just plug in your test secret key

rvem:
> how do you mean? launchctl would be on mac...

Oh, right, I mean `journalctl` 🤦

itkach:
> > how do you mean? launchctl would be on mac...
> 
> Oh, right, I meant `journalctl` 🤦

there is no additional info in the log, journalctl has literally those three lines that `systemctl  status` snippet shows

rvem:
Okay, looks like I know why the very first start of the baker service fails. When `/var/lib/tezos/.tezos-client` is fresh and doesn't have a `config` file, both baker and endorser startup scripts see that there is no `/var/lib/tezos/.tezos-client/config` and try to init this config, endorser manages to do that faster and the baker fails with the following error:
```
Error:
  Config file already exists at location: /var/lib/.tezos-client/config
```
Apparently, the best solution for this problem is to add a dependency for these two services, e.g. make endorser daemon start after baker (or vice versa).

rvem:
Instead of adding a dependency, I've allowed restarting daemons services instead

itkach:
restarting these daemons on failure is probably the right thing to do in general, but perhaps tezos-baking-{network} service could pre-initialize tezos-client config, doesn't look like that responsibility belongs with either baker or endorser.

itkach:
tried current version - setup went smoothly, everything worked out :) I see that baker initially failed to start but immediately "self-corrected":

```
Apr 19 16:26:51 tez3 systemd[1]: Started Instantiated tezos baker daemon service.
Apr 19 16:26:52 tez3 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Main process exited, code=exited, status=1/FAILURE
Apr 19 16:26:52 tez3 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Failed with result 'exit-code'.
Apr 19 16:26:52 tez3 systemd[1]: tezos-baker-008-ptedo2zk@edo2net.service: Scheduled restart job, restart counter is at 1.
Apr 19 16:26:52 tez3 systemd[1]: Stopped Instantiated tezos baker daemon service.
Apr 19 16:26:52 tez3 systemd[1]: Started Instantiated tezos baker daemon service.
Apr 19 16:26:52 tez3 tezos-baker-start[3104]: Connection refused, retrying in 1.00 seconds...
Apr 19 16:26:53 tez3 tezos-baker-start[3104]: Waiting for the node to be bootstrapped...
Apr 19 16:26:53 tez3 tezos-baker-start[3104]: Connection refused, retrying in 1.50 seconds...
Apr 19 16:26:55 tez3 tezos-baker-start[3104]: Connection refused, retrying in 2.25 seconds...
Apr 19 16:26:57 tez3 tezos-baker-start[3104]: Current head: BMG3Zrd8ce93 (timestamp: 2021-04-19T17:55:35.000-00:00, validation: 2021-04-19T20:26:57.702-00:00)
```

rvem:
> doesn't look like that responsibility belongs with either baker or endorser

That's a good point, client config generation/update was moved to the `tezos-baking-{network}` prestart stage

docker/package/model.py
```diff
@@ -574,3 +577,13 @@ def gen_rules(self, out):
         rules_contents = gen_systemd_rules_contents(self)
         with open(out, 'w') as f:
             f.write(rules_contents)
+
+    def gen_install(self, out):
+        startup_scripts = \
+            list(set(filter(lambda x: x is not None, map(lambda x: x.startup_script, self.systemd_units))))
```
itkach, line 583: a nitpick, but the same spelled out with a list comprehension is both shorter and more idiomatic python: 

```
set([x.startup_script for x in self.systemd_units if x is not None])
```

docker/package/scripts/tezos-baking-prestart
```diff
@@ -0,0 +1,23 @@
...
+mkdir -p "$DATA_DIR"
+
+if [[ ! -f $config ]]; then
+    "$tezos_client" --base-dir "$DATA_DIR" \
+                    --endpoint "$NODE_RPC_ENDPOINT" \
+                    config init
+else
+    "$tezos_client" --base-dir "$DATA_DIR" \
+                    --endpoint "$NODE_RPC_ENDPOINT" \
+                    config update
```
itkach, line 22: the two commands appear to differ only in init vs update, perhaps it makes sense to spell it out in the way that makes this apparent, something like this:

```
"$tezos_client" --base-dir "$DATA_DIR" \
                    --endpoint "$NODE_RPC_ENDPOINT" \
                    config $([[ -f $config ]] && echo "update" || echo "init")
```

docker/package/scripts/tezos-node-prestart
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
```
pasqu4le, line 3:
```suggestion
# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
```

docker/package/defaults/tezos-baking.conf
```diff
@@ -0,0 +1,9 @@
+# vim: ft=sh
+# SPDX-FileCopyrightText: 2020 TQ Tezos <https://tqtezos.com/>
```
pasqu4le, line 2:
```suggestion
# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
```

README.md
```diff
@@ -159,6 +159,10 @@ data directory.
 `tezos-{accuser, endorser}` have configurable node address, so that they can be used with both
 remote and local node.
 
+As an addition, `tezos-baking` package provides `tezos-baking-<network>` services that orchestrate
+systemd units for `tezos-node`, `tezos-baker-<proto>`, and `tezos-endorser-<proto>`.
+Configuration files for these services are located in `/etc/default/tezos-baking-<network>`.
```
pasqu4le, line 164: I think we shouldn't mention this until we create a new native release, just as we don't update the baking doc here either to avoid incorrect informations.

docker/package/defaults/tezos-baking.conf
```diff
@@ -0,0 +1,9 @@
...
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+# shellcheck disable=SC2034
+DATA_DIR="/var/lib/tezos/.tezos-client"
+NODE_RPC_ENDPOINT="http://localhost:8732"
```
pasqu4le, line 8: Since this is common to `baking-<proto>` services of different protocols it is somewhat problematic when running multiple baking instances for different ones.

The issue is not introduced here and the same problem is true for running single `node`/`baker`/`endorser` services, so I suggest merging this PR and solving it somewhere else.
It's not really a breaking problem, since one can change the configurations, but it would be nice to have default values that can support multiple protocols by using different ports.

A similar thing can be said for `tezos-client` that will just "aim" at this port. For that I suggest extending the scope of #148 to also add `tezos-client-mainnet` `tezos-client-edo2net` and `tezos-client-florencenet` aliases to avoid the need to manually set up the RPC port.

pasqu4le (approved):
Tested the latest changes again.

All seems to work fine, but it's not ideal in case one wants to run multiple instances for different networks.

See the comment below. IMO this should be merged now and that problem handled separately.

PR #165, opened 2021-04-02
v9.0-rc1 brew formulas

pasqu4le:
## Description

Updates to the brew formulas for the new `v9.0-rc1` pre-release.

## Related issue(s)

None.

  can be successfully built.

### DIFF
--- a/Formula/tezos-node-edo2net.rb
+++ b/Formula/tezos-node-edo2net.rb
@@ -4,33 +4,13 @@
 
 class TezosNodeEdo2net < Formula
   url "file:///dev/null"
-  version "v8.2-3"
+  version "v9.0-rc1-1"
 
   bottle :unneeded
   depends_on "tezos-node"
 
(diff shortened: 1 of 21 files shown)


Formula/tezos-node-edo2net.rb
```diff
@@ -4,7 +4,7 @@
 
 class TezosNodeEdo2net < Formula
   url "file:///dev/null"
-  version "v8.2-3"
+  version "v9.0-rc1-1"
```
rvem, line 7: v9.0-rc1 now supports `--network edo2net`, so I propose to remove ugly workarounds with `cat`ing the required node config in this PR

rvem (approved):
LGTM, have only one minor suggestion about `tezos-node-edo2net.rb` service configuration

Formula/tezos-node-edo2net.rb
pasqu4le: :ok_hand: good point, done

PR #169, opened 2021-04-13
Mention mirrored taps in the README

rvem:
## Description
Problem: We'd like to have an ability to install both latest stable
version and latest release candidate brew formulas on the macOS.
However, brew tap uses master branch of the repo by default and it
cannot be changed.

Solution: Create two mirrors of the tezos-packaging repo, one mirror for
the latest stable release and one mirror for the latest release
candidate release, thus users may use one of these repos as a brew tap
and install desired stable or release candidate formulas. Mention this
capability in the README.

## Related issue(s)

https://issues.serokell.io/issue/TM-535

  can be successfully built.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,10 +98,22 @@ Run `./tezos-client` or add it to your PATH to be able to run it anywhere.
 <a name="macos"></a>
 ## Brew tap for macOS
 
-If you're using macOS and `brew`, you can install Tezos binaries from the tap
-provided by this repository. In order to do that run the following:
+If you're using macOS and `brew`, you can install Tezos binaries from the taps provided
+by this repository. There are two taps: one for the latest stable release of Tezos and
+one for the latest release candidate of Tezos.
+
+In order to use latest stable version run the following:
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,10 +98,19 @@ Run `./tezos-client` or add it to your PATH to be able to run it anywhere.
...
+```
+brew tap serokell/tezos-packaging-stable https://github.com/serokell/tezos-packaging-stable.git
+brew install tezos-client
+```
+
+In order to use latest release candidate version run the following:
 ```
-brew tap serokell/tezos-packaging https://github.com/serokell/tezos-packaging.git
+brew tap serokell/tezos-packaging-rc https://github.com/serokell/tezos-packaging-rc.git
 brew install tezos-client
```
pasqu4le, line 114: Nitpick, but I think we can mention the `install` command once.

I think you can just remove it from both here and above and add something after like:
"You can then install the chosen package, e.g.:"
and add it there.

pasqu4le (approved):
LGTM, only one minor suggestion.

PR #185, opened 2021-05-03
[#180] Provide ledger udev rules with tezos-client

### LINKED ISSUE #180: [BUG] tezos-client doesn't see Ledger Nano S on Linux
# To Reproduce

- install tezos-client on a Linux machine (not VM)
- plug in Ledger Nano S into a USB port
- start tezos wallet or baking app on the Ledger device
- run `tezos-client list connected ledgers`
- observe no ledgers listed

# Expected behavior
- tezos-client finds and reports connected ledger with output that start with
```
## Ledger `mark-dog-light-monkey`
Found a Tezos Baking 2.2.9 (git-description: "b28c2364") application running
on Ledger Nano S at [0001:0007:00].
...
```
# Environment
- Raspberry Pi OS (64 Bit)
- 9.0-1

# Additional context
Looks like udev rules need to be created for the ledger device. Ledger has a repo for the rules: https://github.com/LedgerHQ/udev-rules. Also note currently proposed rules are reported to not work on Ubuntu 18.04, which also appears to be the case for the 64-bit Raspberry Pi OS: https://github.com/LedgerHQ/udev-rules/issues/5. As mentioned in the issue, older rules do work. I added these udev rules from Kiln: https://gitlab.com/tezos-kiln/kiln/-/blob/develop/linux-distros.nix#L171 and tezos-client started to see the device.

rvem:
## Description
Problem: Users have to manually install udev rules in case they want to
use tezos-client along with the ledger device.

Solution: Install udev rules during post-installation stage of the
tezos-client package.

Note that potentially further changes can be required in case firmware update won't resolve an issue with the
udev rules on the Raspberry Pi.

## Related issue(s)

Relates #180 (issue will be resolved once updated packages are published)

  can be successfully built.

### DIFF
--- a/docker/package/packages.py
+++ b/docker/package/packages.py
@@ -63,18 +63,22 @@ signer_units = [
         config_file="tezos-signer.conf")
 ]
 
+ledger_udev_postinst = open(f"{os.path.dirname( __file__)}/scripts/udev-rules", 'r').read()
+
 packages = [
     OpamBasedPackage("tezos-client",
                      "CLI client for interacting with tezos blockchain",
                      optional_opam_deps=["tls", "ledgerwallet-tezos"],
-                     requires_sapling_params=True),
+                     requires_sapling_params=True,
+                     postinst_steps=ledger_udev_postinst),
     OpamBasedPackage("tezos-admin-client",
                      "Administration tool for the node",
                      optional_opam_deps=["tls"]),
     OpamBasedPackage("tezos-signer",
                      "A client to remotely sign operations or blocks",
                      optional_opam_deps=["tls", "ledgerwallet-tezos"],
-                     systemd_units=signer_units),
+                     systemd_units=signer_units,
+                     postinst_steps=ledger_udev_postinst),
     OpamBasedPackage("tezos-codec",
                      "A client to decode and encode JSON")
 ]
@@ -237,7 +241,8 @@ for proto in active_protocols:
                                      optional_opam_deps=["tls", "ledgerwallet-tezos"],
                                      requires_sapling_params=True,
                                      postinst_steps= \
-                                        daemon_postinst_common + gen_daemon_specific_postinst(f"tezos-baker-{proto}"),
+                                        daemon_postinst_common + gen_daemon_specific_postinst(f"tezos-baker-{proto}") \
+                                            + ledger_udev_postinst,
                                      postrm_steps=gen_daemon_specific_postrm(f"tezos-baker-{proto}")))
     packages.append(OpamBasedPackage(f"tezos-accuser-{proto}", "Daemon for accusing",
                                      [SystemdUnit(service_file=service_file_accuser,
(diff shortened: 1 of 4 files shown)


docker/package/packages.py
```diff
@@ -63,11 +63,34 @@
         config_file="tezos-signer.conf")
 ]
 
+ledger_udev_postinst = '''
+if [ ! -f /etc/udev/rules.d/20-hw1.rules ]; then
+    cat <<EOF > /etc/udev/rules.d/20-hw1.rules
```
pasqu4le, line 68: One question to clean this up a bit: would it not be possible to move this into a separate file to use here and not keep all it content in `packages.py`?

Asking as it is static content and this isn't easy to look at :slightly_smiling_face:

itkach: I agree, separate file with rules included as part of the .deb would be nice. I would go even further and suggest that there's value in making a separate .deb package for udev rules so that all tezos packages that have Ledger Nano S functionality could depend on it: client, signer, baker, endorser and accuser. Otherwise baker, for example, would only work with ledger nano s if tezos-client is also installed.

rvem:
So, I've updated the script that adds udev rules, so that it's supposed to work on both existing Ubuntu versions and Raspberry Pi OS. Also, udev rules are now added for all packages post-installs that can use ledger (specifically: `tezos-client`, `tezos-signer`, `tezos-baker`, `tezos-endorser`)

docker/package/scripts/udev-rules
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: 2018-2021 Ledger <https://www.ledger.com>
+#
+# SPDX-License-Identifier: LicenseRef-Apache-Ledger
+
```
pasqu4le, line 4: I think it would be worth putting in a comment here the repository and commit where this came from, for future reference.

Also I noticed that compared to a previous diff you removed the 
```
if [ ! -f /etc/udev/rules.d/20-hw1.rules ]; then
```
initial check. Did you find it to be unnecessary?

Overriding the udev config file is not an issue, but perhaps we'd want to avoid making the two `udevadm` calls if possible?

docker/package/packages.py
```diff
@@ -237,7 +241,8 @@ def gen_daemon_specific_postrm(daemon_name):
                                      optional_opam_deps=["tls", "ledgerwallet-tezos"],
                                      requires_sapling_params=True,
                                      postinst_steps= \
-                                        daemon_postinst_common + gen_daemon_specific_postinst(f"tezos-baker-{proto}"),
+                                        daemon_postinst_common + gen_daemon_specific_postinst(f"tezos-baker-{proto}") \
+                                            + ledger_udev_postinst,
                                      postrm_steps=gen_daemon_specific_postrm(f"tezos-baker-{proto}")))
     packages.append(OpamBasedPackage(f"tezos-accuser-{proto}", "Daemon for accusing",
```
pasqu4le, line 247: Is the accuser supposed to have this as well?

docker/package/scripts/udev-rules
rvem: I've decided to remove `if [ ! -f /etc/udev/rules.d/20-hw1.rules ]; then` in case we'll have to update provided udev rules in future.

docker/package/packages.py
rvem: hmm, indeed, missed it somehow, thanks for catching

docker/package/scripts/udev-rules
rvem: > we'd want to avoid making the two `udevadm` calls if possible

AFAIU, `udevadm` has to be called in case rules were updated, we can avoid calling it by looking at the diff between existing rules and rules that we're going to write, but the overhead for calling `udevadm` doesn't look significant for me

pasqu4le: :ok_hand: sounds reasonable to me

pasqu4le: Oh but I still think this:

> I think it would be worth putting in a comment here the repository and commit where this came from, for future reference.

pasqu4le (approved):

PR #188, opened 2021-05-04
[#175] Python code refactoring

### LINKED ISSUE #175: Consider cleaning up and formatting python code

There is now quite a bit of python code in the repository, but it isn't it the best shape.

We should consider cleaning it up, mainly for readability in case someone wants to check it out.

# Acceptance criteria

Revisit the python code and consider:
- making changes for readability and polishing in general
- using a formatting tool, [black](https://github.com/psf/black) seems to be a very common one

rvem:
## Description
There is quite a lot of Python code in the repository now and it doesn't look very well.

This PR provides some refactorings to it:
* Formatting via `black`
* Various refactoring to reduce size of existing modules

## Related issue(s)

Resolves #175

  can be successfully built.

### DIFF
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -68,13 +68,11 @@ class AbstractPackage:
 
 def gen_spec_systemd_part(package):
     systemd_units = package.systemd_units
-    startup_scripts = list(
-        set(map(lambda x: x.startup_script, systemd_units))
-    ) + list(set(map(lambda x: x.prestart_script, systemd_units)))
+    startup_scripts = list(set(map(lambda x: x.startup_script, systemd_units))) + list(
+        set(map(lambda x: x.prestart_script, systemd_units))
+    )
     config_files = list(
-        filter(
-            lambda x: x is not None, map(lambda x: x.config_file, systemd_units)
-        )
+        filter(lambda x: x is not None, map(lambda x: x.config_file, systemd_units))
     )
     install_unit_files = ""
     systemd_unit_files = ""
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: LicenseRef-MIT-TQ
 
 import os, subprocess
+import shutil
 from abc import abstractmethod
 from typing import List, Dict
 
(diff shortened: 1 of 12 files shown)


docker/package/systemd.py
```diff
@@ -0,0 +1,141 @@
...
+        instances: List[str] = [],
+    ):
+        self.suffix = suffix
+        self.service_file = service_file
+        self.startup_script = startup_script
+        self.startup_script_source = startup_script_source
+        self.prestart_script = prestart_script
+        self.prestart_script_source = prestart_script_source
+        self.config_file = config_file
+        self.instances = instances
```
ixahedron, line 84: This looks kinda ugly with the whole giant `__init__` duplication :disappointed: I see that there's an option to use [`@dataclass`](https://docs.python.org/3/library/dataclasses.html), maybe we would want to use that style to automatically assign fields?

ixahedron (approved):
LGTM aside from a couple of comments. Feel free to decide if any suggestion should be applied at all.

docker/package/systemd.py
rvem: :pogchamp:

docker/package/model.py
```diff
@@ -128,34 +108,40 @@ def gen_spec_systemd_part(package):
             service_name = f"%{{name}}-{systemd_unit.suffix}"
         if len(systemd_unit.instances) > 0:
             service_name = f"{service_name}@"
-        install_unit_files += f"install -m 644 {service_name}.service %{{buildroot}}/%{{_unitdir}}\n"
+        install_unit_files += (
+            f"install -m 644 {service_name}.service %{{buildroot}}/%{{_unitdir}}\n"
+        )
```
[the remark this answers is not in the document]
rvem: AFAIU, that's one way to separate a long line (the other is to use `\`, but black uses parentheses)

docker/package/systemd.py
rvem: Sadly, Ubuntu 18.04 (whose docker image is used for package building) has python 3.6.9 and dataclasses aren't supported there. But I think it's time to update and use 20.04 instead

PR #189, opened 2021-05-04
Build native packages in CI

rvem:
## Description
Problem: v9.1 was released to the opam-repository, this means we can now
build native packages.

Solution: Enable native packages building in CI.

## Related issue(s)

Resolves None

  can be successfully built.


pasqu4le (approved):

pasqu4le:
I was thinking: can we have a variable in the pipeline definition to activate or deactivate these CI steps?

This way we can just have a one-line diff the next time (in most cases).

rvem:
Apparently, buildkite doesn't provide a way to use environment variables defined in the current `pipeline.yml` via `env:` in the conditionals:(

Issue #193, opened 2021-05-07
[BUG] Node services are enabled by default


rvem:
# Description
After #156, node services became a part of baking services. In order to propagate start actions via `WantedBy` rule,
these services were enabled during packages post-installation. Unfortunately, this means that systemd will try to start these services on the next start-up. 

# To Reproduce

```
sudo apt-get install tezos-node
# reboot
systemctl status tezos-node-edo2net
# service is running, however, we haven't explicitly enabled it
```

# Expected behavior
Node units aren't started on startup by default.

# Environment

- Ubuntu 20.04
- v9.1-1 `tezos-node` package from launchpad

rvem:
Fixed in v9.1-2 native packages

PR #194, opened 2021-05-10
[#193] Don't use WantedBy in parts of baking services

rvem:
## Description
Problem: WantedBy rule requires services to be enabled in order to take
effect. Enabling also makes services start on system start-up, which is
quite unpleasant.

Solution: Don't use WantedBy and don't enable services in the
post-installation stage. Use Requires rule in the baking services to list
dependencies instead.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -113,7 +113,6 @@ def gen_spec_systemd_part(package):
                                                                 package.systemd_units)))
     install_unit_files = ""
     systemd_unit_files = ""
-    enable_units = ""
     systemd_units_post = ""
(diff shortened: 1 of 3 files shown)


rvem:
To test locally on Ubuntu 20.04:
```
sudo add-apt-repository ppa:rvem/tezos-experiments-1
sudo apt-get update
sudo apt-get install tezos-baking
```

pasqu4le (approved):
LGTM and all seems to be working properly even after a restart :+1:

PR #196, opened 2021-05-11
[#195] Dynamically generate pipeline via nix

### LINKED ISSUE #195: Use smarter pipelines on CI

Currently our buildkite CI pipelines do not consider what changes have been made and so as a result no matter what all the steps are executed (provided other conditions, that are not diff-related).

This is not great as the whole pipeline takes a lot to run and it will do so even on a PR that e.g. only updates a markdown doc.

# Acceptance criteria

Our pipeline can make diff-based decision on what to run and what to skip, avoiding unnecessary steps when possible.

rvem:
## Description
Problem: Some of the ci steps take a huge amount of time. We run them on
any changes even when the changes don't affect any of them.

Solution: Make it possible to provide regexes, which will define paths,
whose changes will trigger step to run. In order to do that pipeline was
defined as a nix attrset. This attrset later supposed to be filtered
basing on the output of 'git diff origin/master...HEAD --name-only'.
The files from diff list are matched against the regexes from the
pipeline config. If the step doesn't match any of them, then it isn't
included in the resulting generated pipeline config.

To generate pipeline config locally run:
```
nix-build .buildkite/filter-pipeline.nix --argstr git_diff "$(git diff origin/master...HEAD --name-only)"
```

~Note, that this PR is currently draft and changes from it don't affect the existing CI configuration,
because it requires some changes in the [pipeline configuration on buildkite website](https://buildkite.com/serokell/tezos-packaging/steps), these changes will break CI in all other branches.~

Existing `.buildkite/pipeline.yml` should be removed before merging as well as the old `buildkite-agent pipeline upload` step
from the pipeline configuration on the website.

## Related issue(s)

Resolves #195

  can be successfully built.

### DIFF
--- a/.buildkite/filter-pipeline.nix
+++ b/.buildkite/filter-pipeline.nix
@@ -1,0 +1,34 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+{ git_diff }:
+let
+  pkgs = (import ../. {}).pkgs;
+  pipeline = import ./pipeline.nix;
+  inherit (pkgs) lib;
+  inherit (lib) any crossLists filter splitString;
+
+  splitted_diff = (splitString "\n" git_diff);
+
+  filter_steps = { steps, diff }: filter (step:
+      if ! step ? only_changes
+      then true
+      else
+        let
+          only_changes = step.only_changes;
+          bool_list = crossLists
+            (diff_element: regex: builtins.match regex diff_element != null)
+            [ diff only_changes ];
+        in any (x: x) bool_list
+    ) steps;
+
+  remove_only_changes_attr =
+    step: if builtins.isString step then step else builtins.removeAttrs step ["only_changes"];
+
+  filter_pipeline = { pipeline, diff }:
+    pipeline // { steps = map remove_only_changes_attr (filter_steps { inherit (pipeline) steps; inherit diff; }); };
+
+in pkgs.writeText "pipeline.yml"
+  (builtins.toJSON (filter_pipeline
+    { pipeline = pipeline; diff = splitted_diff; }
+  ))
--- a/.buildkite/pipeline.nix
+++ b/.buildkite/pipeline.nix
@@ -66,7 +66,7 @@ in
         "eval \"$SET_VERSION\""
         "cd docker"
         "./docker-static-build.sh"
-        "nix run -f.. pkgs.upx -c upx tezos-*"
+        "upx tezos-*"
         "for f in ./tezos-*; do mv mv \"\$f\" \"\$f-arm64\"; done"
       ];
       artifact_paths = [
--- a/.buildkite/pipeline.nix
+++ b/.buildkite/pipeline.nix
@@ -1,0 +1,172 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+let
+  native_packaging_changes_regexes = [
+    "docker\/package/\.*"
+    "docker\/docker-tezos-packages\.sh"
+    "meta\.json"
+    "protocols\.json"
+    "nix\/nix\/sources\.json"
+  ];
+  static_binaries_changes_regexes = [
+    "docker\/build/\.*"
+    "docker\/docker-static-build\.sh"
+    "nix\/nix\/sources\.json"
+    "meta\.json"
+    "protocols\.json"
+  ];
+  nix_changes_regexes = [
+    ".*\.nix"
+    "protocols\.json"
+  ];
+  brew_changes_regexes = [
+    "Formula\/.*\.rb"
+  ];
+in
+{
+  env = {
+    SET_VERSION = "export TEZOS_VERSION=\"$(cat nix/nix/sources.json | jq -r '.tezos.ref' | cut -d'/' -f3)\"";
+  };
+  steps = [
+    { label = "reuse lint";
+      command= "nix run -f . pkgs.reuse -c reuse lint";
+    }
+    { label = "check trailing whitespace";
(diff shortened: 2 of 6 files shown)


zhenyavinogradov:
> Note, that this PR is currently draft and changes from it don't affect the existing CI configuration

You can make a conditional step in buildkite, filtering by branch name

rvem:
> You can make a conditional step in buildkite, filtering by branch name

Indeed, thanks

.buildkite/filter-pipeline.nix
```diff
@@ -0,0 +1,34 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+{ git_diff }:
```
zhenyavinogradov, line 4: You calculate diff with `origin/master` in the buidlkite step, but `origin/master` in buildkite's checkout may be stale because it only fetches the branch it needs, you probably need to fetch master explicitly

.buildkite/pipeline.nix
```diff
@@ -0,0 +1,172 @@
...
+    "docker\/build/\.*"
+    "docker\/docker-static-build\.sh"
+    "nix\/nix\/sources\.json"
+    "meta\.json"
+    "protocols\.json"
+  ];
+  nix_changes_regexes = [
+    ".*\.nix"
+    "protocols\.json"
+  ];
```
zhenyavinogradov, line 22: It doesn't seem enough for nix jobs, e.g. there is `sources.json` file and some `.patch` files. But maybe for nix it's simpler to not use any conditions, because it knows itself what to rebuild and what not to rebuild.

.buildkite/pipeline.nix
```diff
@@ -0,0 +1,172 @@
...
+      ];
+      agents = { queue = "x86_64-darwin"; };
+      only_changes = brew_changes_regexes;
+    }
+    "wait"
+    { label = "test docker-built binaries";
+      commands = [
+        "buildkite-agent artifact download \"docker/*\" . --step \"build via docker\""
+        "chmod +x ./docker/*"
+        "nix-build tests/tezos-binaries.nix --no-out-link --arg path-to-binaries ./docker"
```
zhenyavinogradov, line 94: Should `tests/tezos-binaries.nix` be included in `only_changes` for this job?

.buildkite/pipeline.nix
```diff
@@ -0,0 +1,172 @@
...
+        "rm -rf out"
+      ];
+      artifact_paths = [
+        "./out/*"
+      ];
+      # It takes much time to build binary package, so we do it only on master
+      branches = "master";
+      timeout_in_minutes = 90;
+      agents = { queue = "docker"; };
+      only_changes = native_packaging_changes_regexes;
```
zhenyavinogradov, line 127: There will actually be issues with steps that run on master branch, the diff will always be empty :pk:

.buildkite/pipeline.nix
```diff
@@ -0,0 +1,172 @@
...
+    { label = "test deb binary packages via docker";
+      commands = [
+        "eval \"$SET_VERSION\""
+        # Building all binary packages will take significant amount of time, so we build only one
+        # in order to ensure package generation sanity
+        "./docker/docker-tezos-packages.sh --os ubuntu --type binary --package tezos-baker-008-PtEdo2Zk"
+        "rm -rf out"
+      ];
+      artifact_paths = [
+        "./out/*"
```
zhenyavinogradov, line 121: But `out` is removed above?

.buildkite/pipeline.nix
```diff
@@ -0,0 +1,172 @@
...
+    { label = "test r binary packages via docker";
+      commands = [
+        "eval \"$SET_VERSION\""
+        # Building all binary packages will take significant amount of time, so we build only one
+        # in order to ensure package generation sanity
+        "./docker/docker-tezos-packages.sh --os fedora --type binary --package tezos-baker-008-PtEdo2Zk"
+        "rm -rf out"
+      ];
+      artifact_paths = [
+        "./out/*"
```
zhenyavinogradov, line 151: `out` removed in this job too

.buildkite/pipeline.nix
```diff
@@ -0,0 +1,172 @@
...
+      ];
+      branches = "!master";
+      timeout_in_minutes = 60;
+      agents = { queue = "docker"; };
+      only_changes = native_packaging_changes_regexes;
+    }
+    { label = "test r binary packages via docker";
```
zhenyavinogradov, line 142:
```suggestion
    { label = "test rpm binary packages via docker";
```

.buildkite/pipeline.nix
rvem: hmm, indeed, using `git diff` won't work this easy :(

.buildkite/filter-pipeline.nix
rvem: hmm, apparently, fetching `origin/master` is a bad idea as well because we probably want to compare our changes with the base branch and not with the existing master, e.g. in case where a feature branch is outdated and needs to be rebased

.buildkite/pipeline.nix
rvem: looks like `git diff --name-only HEAD HEAD~1` should be used in case `git diff --name-only origin/master...HEAD` is empty

.buildkite/filter-pipeline.nix
zhenyavinogradov: If you use `...` and not `..` in `git diff`, it will compare to the common base, where the branch forked off

rvem: looks like the proper way is to fetch `origin/master` and do `git diff --name-only $(git merge-base --fork-point origin/master)`

rvem: looks like I'm wrong because `git diff origin/master...HEAD` uses `git merge-base` inside

rvem:
Closed in favour of #197

PR #200, opened 2021-05-18
Deactivate protocol 008

ixahedron:
## Description

Problem: edo2net is now obsolete, so we don't need to support it.

Solution: removed 008 support from tezos-packaging.

## Related issue(s)

Resolves part of [TM-546](https://issues.serokell.io/issue/TM-546).

  can be successfully built.


protocols.json
```diff
@@ -18,7 +18,6 @@
     "008-PtEdoTez"
   ],
   "active": [
-    "008-PtEdo2Zk",
```
rvem, line 21: Please move to the `allowed` instead of removing, so that 008 is still compiled in the `tezos-node` binary (it simplifies bootstrap process for the `mainnet`)

rvem (approved):

rvem:
Oh, one more thing, please remove `[TM-546]` prefix from the commit message header during the rebase.
Some time ago we considered that it's a bad idea to reference our private issue tracker in the public repositories

Issue #202, opened 2021-05-19
[BUG] Fail to install tezos-client on Ubuntu subsytem for Windows


GuillaumeGen:
# Description

When I try to install tezos-client, it fails with the following error message:

```
Setting up tezos-client (2:9.1-0ubuntu3~focal) ...
Failed to send reload request: No such file or directory
dpkg: error processing package tezos-client (--configure):
 installed tezos-client package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 tezos-client
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

# To Reproduce

I ran `sudo add-apt-repository ppa:serokell/tezos && sudo apt-get update` which worked fine and then `sudo apt install tezos-client`

# Expected behavior

The package to be installed and no error message to be printed.

# Environment

- Windows 10 Version 20H2 (build 19042.985)
- WSL: Ubuntu 20.04.2 LTS

rvem:
Fixed in `2:9.1-0ubuntu4` `tezos-client`

PR #204, opened 2021-05-20
[#202] Don't install udev rules inside WSL

rvem:
## Description
Problem: udev and libusb aren't fully supported in the WSL, thus an
attempt to update udev rules for a usb device fails, which leads to
inablitity of installation of our native packages.

Solution: Check if we're inside the WSL (by checking the contents of
/proc/sys/kernel/osrelease) and don't install udev rules if we're inside
it.

## Related issue(s)

Relates #202  (issue will be resolved once updated packages are published)

  can be successfully built.


pasqu4le (approved):
Tested as well, all LGTM :ok_hand:

PR #206, opened 2021-05-25
[#201] Add first version of baking wizard

### LINKED ISSUE #201: Create the first version of a packaging wizard

We want to have an interactive way for people to set up baking (or just a node) that doesn't involve inputting many commands in the terminal.

To do so an option is to add a new native package that contains a CLI wizard utility.

For a first version this wizard should be able to guide the user through the steps that are currently described in the [baking article](https://github.com/serokell/tezos-packaging/blob/master/docs/baking.md), provide choices interactively for the options available and execute the commands (mainly `systemctl`) for them.

# Acceptance criteria

A new native package (`tezos-baking-setup` ?) provides a wizard for easier set-up of tezos configuration and `systemd` services.
It should be able to;
- interactively ask for:
   -  the network to use
   - if the user wants to run only a node or also baking
   - what type of baker key import to use
- provide suggestions for the options above as well
- wait for the step that require some time
- manage (at least launch) `systemd` services
- set up what necessary for the `tezos` user

ixahedron:
Problem: we want to have an interactive way to set up baking infrastructure, or just `tezos-node`.

Solution: created a CLI wizard to collect configuration through queries and execute the approppriate steps.

## Related issue(s)

Resolves part of #201 

  can be successfully built.

### DIFF
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -502,6 +502,7 @@ class TezosBakingServicesPackage(AbstractPackage):
 
     def fetch_sources(self, out_dir):
         os.makedirs(out_dir)
+        shutil.copy(f"{os.path.dirname(__file__)}/tezos_baking_wizard.py", out_dir)
 
     def gen_control_file(self, deps, ubuntu_version, out):
         run_deps = ", ".join(
@@ -584,9 +585,17 @@ Maintainer: {self.meta.maintainer}
         file_contents = """
 .PHONY: install
 
+BINDIR=/usr/bin
+
 tezos-baking:
 
-install: tezos-baking
+tezos-baking-wizard:
+	mv $(CURDIR)/tezos_baking_wizard.py $(CURDIR)/tezos-baking-wizard
+	chmod +x $(CURDIR)/tezos-baking-wizard
+
+install: tezos-baking tezos-baking-wizard
+	mkdir -p $(DESTDIR)$(BINDIR)
+	cp $(CURDIR)/tezos-baking-wizard $(DESTDIR)$(BINDIR)/tezos-baking-wizard
 """
         with open(out, "w") as f:
             f.write(file_contents)
--- a/docker/package/tezos_baking_wizard.py
+++ b/docker/package/tezos_baking_wizard.py
@@ -1,0 +1,611 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+"""
+A wizard utility to help set up tezos-baker.
+
+Asks questions, validates answers, and executes the appropriate steps using the final configuration.
+"""
+
+import os, sys, subprocess, shlex
+import readline
+import re, textwrap
+import urllib.request
+from typing import List
+
+# Input validators
+
+
+def enum_range_validator(options):
+    def _validator(input):
+        intrange = list(map(str, range(1, len(options) + 1)))
+        if input not in intrange and input not in options:
+            raise ValueError(
+                "Please choose one of the provided values or use their respective numbers: "
+                + ", ".join(options)
+            )
+        try:
+            opt = int(input) - 1
+        except:
+            return input
+        else:
+            opts = options
+            if isinstance(options, dict):
+                opts = list(options.keys())
+            return opts[opt]
+
+    return _validator
+
+
+def filepath_validator(input):
+    if input and not os.path.isfile(input):
+        raise ValueError("Please input a valid file path.")
+    return input
+
+
+def required_field_validator(input):
+    if not input.strip():
+        raise ValueError("Please provide this required option.")
+    return input
+
+
+# Runs the input through the passed validator, allowing for possible alteration,
+# but doesn't raise an exception if it doesn't validate to allow for custom options, too.
+def or_custom_validator(validator):
+    def _validator(input):
+        result = input
+        try:
+            result = validator(input)
+        except:
+            pass
+        return result
+
+    return _validator
+
+
+class Validator:
+    def __init__(self, validator):
+        self.validator = validator
+
+    def validate(self, input):
+        if self.validator is not None:
+            if isinstance(self.validator, list):
+                for v in self.validator:
+                    input = v(input)
+                return input
+            else:
+                return self.validator(input)
+
+
+# Regexes
+
+ledger_regex = b"ledger:\/\/[\w\-]+\/[\w\-]+\/[\w']+\/[\w']+"
+secret_key_regex = b"(encrypted|unencrypted):\w+"
+address_regex = b"tz1\w{33}"
+
+# Wizard CLI utility
+
+
+def proc_call(cmd):
+    return subprocess.check_call(shlex.split(cmd))
+
+
+def fetch_snapshot(url):
+    print("Downloading the snapshot from ", url)
+    filename = "/tmp/tezos_node.snapshot"
+    urllib.request.urlretrieve(url, filename, progressbar_hook)
+    print()
+    return filename
+
+
+def progressbar_hook(chunk_number, chunk_size, total_size):
+    done = chunk_number * chunk_size
+    percent = min(int(done * 100 / total_size), 100)
+    print("Progress: ", percent, "%, ", int(done / (1024 * 1024)), "MB", end="\r")
+
+
+def color(input, colorcode):
+    return colorcode + input + "\x1b[0m"
+
+
+def yes_or_no(prompt, default=None):
+    valid = False
+    while not valid:
+        answer = input(prompt).strip().lower()
+        if not answer and default is not None:
+            answer = default
+        if answer in ["y", "yes"]:
+            return True
+        elif answer in ["n", "no"]:
+            return False
+        else:
+            print(color("Please provide a 'yes' or 'no' answer.", "\x1b[1;31m"))
+
+
+def list_connected_ledgers():
+    output = subprocess.run(
+        shlex.split("sudo -u tezos tezos-client list connected ledgers"),
+        capture_output=True,
+    )
+    return [name.decode() for name in re.findall(ledger_regex, output.stdout)]
+
+
+class Step:
+    def __init__(
+        self,
+        id: str,
+        prompt: str,
+        help: str,
+        default: str = "1",
+        options=None,
+        validator=None,
+        reqs=lambda _vs: True,
+    ):
+        self.id = id
+        self.prompt = prompt
+        self.help = help
+        self.default = default
+        self.options = options
+        self.validator = validator
+        self.reqs = reqs
+
+    def pprint_options(self):
+        i = 1
+        def_i = None
+        try:
+            def_i = int(self.default)
+        except:
+            pass
+
+        if self.options and isinstance(self.options, list):
+            str_format = "{:1}. {}"
+            for o in self.options:
+                if def_i is not None and i == def_i:
+                    print(str_format.format(i, "(default) " + o))
+                else:
+                    print(str_format.format(i, o))
+                i += 1
+        elif self.options and isinstance(self.options, dict):
+            str_format = "{:1}. {:<26}  {}"
+            for o in self.options:
+                description = textwrap.indent(
+                    textwrap.fill(self.options[o], 60), " " * 31
+                ).lstrip()
+                if def_i is not None and i == def_i:
+                    print(str_format.format(i, o + " (default)", description))
+                else:
+                    print(str_format.format(i, o, description))
+                i += 1
+
+
+def run_wizard():
+    print("Tezos Baking Wizard")
+    print()
+    print(
+        "Welcome, this wizard will help you to set up the infrastructure to interact with the Tezos blockchain."
+    )
+    print(
+        "In order to run a baking instance, you'll need the following Tezos binaries:\n",
+        "tezos-client, tezos-node, tezos-baker-<proto>, tezos-endorser-<proto>.\n",
+        "If you have installed tezos-baking, these binaries are already installed.",
+    )
+    print(
+        "To access help and possible options for each question, type in 'help' or '?'."
+    )
+    print("Type in 'exit' to quit.")
+    print()
+
+    networks = {
+        "mainnet": "Main Tezos network",
+        "florencenet": "Test network using version 009 of Tezos protocol (Florence)",
+        "granadanet": "Test network using version 010 of Tezos protocol (Granada)",
+    }
+
+    modes = {
+        "baker": "Set up and start all services for baking: "
+        "tezos-node, tezos-baker, and tezos-endorser.",
+        "node": "Only bootstrap and run the Tezos node.",
+    }
+
+    snapshot_import_modes = {
+        "download rolling": "Import rolling snapshot from xtz-shots.io (recommended)",
+        "download full": "Import full snapshot from xtz-shots.io",
+        "file": "Import snapshot from a file",
+        "url": "Import snapshot from a url",
+        "skip": "Skip snapshot import and synchronize with the network from scratch",
+    }
+
+    key_import_modes = {
+        "ledger": "From a ledger",
+        "secret-key": "Either the unencrypted or password-encrypted secret key for your address",
+        "json": "Faucet JSON file from https://faucet.tzalpha.net/",
+    }
+
+    systemd_enable = {
+        "yes": "Enable and start the service, running it both now and on every boot",
+        "no": "Start the service this time only",
+    }
+
+    values = {}
+
+    steps = [
+        Step(
+            id="network",
+            prompt="Which Tezos network would you like to use?\nCurrently supported:",
+            help="The selected network will be used to set up all required services.\n"
+            "The currently supported protocols are 009-PsFLoren (used on florencenet and mainnet) \n"
+            "and 010-PtGRANAD (granadanet).\nKeep in mind that you must select the test network "
+            "(either florencenet or granadanet) if you plan on baking with a faucet JSON file.",
+            options=networks,
+            validator=Validator(enum_range_validator(networks)),
+        ),
+        Step(
+            id="mode",
+            prompt="Do you want to set up baking or to run the standalone node?",
+            help="By default, tezos-baking provides predefined services for running baking instances "
+            "on different networks.\nSometimes, however, you might want to only run the Tezos node.\n"
+            "When this option is chosen, this wizard will help you bootstrap tezos-node only.",
+            options=modes,
+            validator=Validator(enum_range_validator(modes)),
+        ),
+        Step(
+            id="systemd_mode",
+            prompt="Would you like your setup to automatically start on boot?",
+            help="Starting the service will make it available just for this session, great\n"
+            "if you want to experiment. Enabling it will make it start on every boot.",
+            options=systemd_enable,
+            validator=Validator(enum_range_validator(systemd_enable)),
+        ),
+        Step(
+            id="snapshot",
+            prompt="The Tezos node can take a significant time to bootstrap from scratch.\n"
+            "Bootstrapping from a snapshot is suggested instead.\n"
+            "How would you like to proceed?",
+            help="A fully-synced local tezos-node is required for running a baking instance.\n"
+            "By default, service with tezos-node will start to bootstrap from scratch,\n"
+            "which will take a significant amount of time.\nIn order to avoid this, we suggest "
+            "bootstrapping from a snapshot instead.\n\n"
+            "Snapshots can be downloaded from the following websites:\n"
+            "Tezos Giganode Snapshots - https://snapshots-tezos.giganode.io/ \n"
+            "XTZ-Shots - https://xtz-shots.io/ \n\n"
+            "We recommend to use rolling snapshots. This is the smallest and the fastest mode\n"
+            "that is sufficient for baking (you can read more about other tezos-node history modes here:\n"
+            "https://tezos.gitlab.io/user/history_modes.html#history-modes ).\n"
+            "All commands within the service are run under the 'tezos' user.",
+            options=snapshot_import_modes,
+            validator=Validator(enum_range_validator(snapshot_import_modes)),
+        ),
+        Step(
+            id="snapshot_file",
+            prompt="Provide the path to the tezos-node snapshot file.",
+            help="You have indicated wanting to import the snapshot from a file.\n"
+            "You can download the snapshot yourself e.g. from XTZ-Shots or Tezos Giganode Snapshots.",
+            validator=Validator(filepath_validator),
+            reqs=lambda vs: vs["snapshot"] == "file",
+        ),
+        Step(
+            id="snapshot_url",
+            prompt="Provide the url of the tezos-node snapshot file.",
+            help="You have indicated wanting to import the snapshot from a custom url.\n"
+            "You can use e.g. links to XTZ-Shots or Tezos Giganode Snapshots resources.",
+            reqs=lambda vs: vs["snapshot"] == "url",
+        ),
+        Step(
+            id="key_import_mode",
+            prompt="How do you want to import the baker key?",
+            help="To register the baker, its secret key needs to be imported to the data "
+            "directory first.\nBy default tezos-baking-<network>.service will use the 'baker' "
+            "alias\nfor the key that will be used for baking and endorsing.\n"
+            "If you want to test baking with a faucet file, make sure you have chosen a test network like "
+            + list(networks.keys())[1],
+            options=key_import_modes,
+            validator=Validator(enum_range_validator(key_import_modes)),
+            reqs=lambda vs: vs["mode"] == "baker",
+        ),
+        Step(
+            id="ledger_url",
+            prompt="Provide the ledger URL for importing the baker secret key stored on a ledger.",
+            help="Your ledger URL should look something like <ledger://<mnemonic>/ed25519/0'/1'>",
+            default=None,
+            options=list_connected_ledgers(),
+            validator=Validator(
+                [
+                    required_field_validator,
+                    or_custom_validator(enum_range_validator(list_connected_ledgers())),
+                ]
+            ),
+            reqs=lambda vs: vs["mode"] == "baker" and vs["key_import_mode"] == "ledger",
+        ),
+        Step(
+            id="secret_key",
+            prompt="Provide either the unencrypted or password-encrypted secret key for your address.",
+            help="The format is 'unencrypted:edsk...' for the unencrypted key, or 'encrypted:edesk...'"
+            "for the encrypted key.",
+            validator=Validator(required_field_validator),
+            reqs=lambda vs: vs["mode"] == "baker"
+            and vs["key_import_mode"] == "secret-key",
+        ),
+        Step(
+            id="json_filepath",
+            prompt="Provide the path to your downloaded faucet JSON file.",
+            help="Download the faucet JSON file from https://faucet.tzalpha.net/.\n"
+            "The file will contain the 'mnemonic' and 'secret' fields.",
+            validator=Validator([required_field_validator, filepath_validator]),
+            reqs=lambda vs: vs["mode"] == "baker" and vs["key_import_mode"] == "json",
+        ),
+    ]
+
+    for step in steps:
+        if not step.reqs(values):
+            continue
+
+        validated = False
+        while not validated:
+            print(step.prompt)
+            step.pprint_options()
+            answer = input("> ").strip()
+
+            if answer.lower() in ["quit", "exit"]:
+                print("Exiting Tezos Baking Wizard")
+                return None
+            elif answer.lower() in ["help", "?"]:
+                print(step.help)
+                print()
+            else:
+                if not answer and step.default is not None:
+                    answer = step.default
+
+                try:
+                    answer = step.validator.validate(answer)
+                except ValueError as e:
+                    print(color("Validation error: " + str(e), "\x1b[1;31m"))
+                else:
+                    validated = True
+                    values[step.id] = answer
+
+    return values
+
+
+class Setup:
+    def __init__(self, config):
+        self.config = config
+
+    # Check if an account with the 'baker' alias already exists, and ask the user
+    # if it can be overwritten.
+    def check_baker_account(self, cmd):
+        output = subprocess.run(shlex.split(cmd), capture_output=True)
+        if output.returncode and re.search(
+            b"The secret_key alias baker already exists.", output.stderr
+        ):
+            value_regex = b"(?:" + ledger_regex + b")|(?:" + secret_key_regex + b")"
+            value = re.search(value_regex, output.stderr).group(0)
+            address = subprocess.run(
+                shlex.split("sudo -u tezos tezos-client show address baker"),
+                capture_output=True,
+            )
+            if address.returncode == 0:
+                address = re.search(address_regex, address.stdout).group(0)
+            print("An account with the 'baker' alias already exists.")
+            print("Its current value is ", value.decode("utf-8"))
+            print("With a corresponding address: ", address.decode("utf-8"))
+
+            if yes_or_no("Should it be replaced with the new import? <y/N> ", "no"):
+                proc_call(cmd + " --force")
+
+    # Check if there is already some blockchain data in the tezos-node data directory,
+    # and ask the user if it can be overwritten.
+    def check_blockchain_data(self):
+        node_dir = "/var/lib/tezos/node-" + self.config["network"]
+        node_dir_contents = os.listdir(node_dir)
+        clean = ["config.json", "version.json"]
+        diff = set(node_dir_contents) - set(clean)
+        if diff:
+            print("The Tezos node data directory already has some blockchain data:")
+            print("\n".join(diff))
+            print()
+            if yes_or_no("Delete this data and bootstrap the node again? <y/N> ", "no"):
+                for path in diff:
+                    if proc_call("sudo rm -r " + os.path.join(node_dir, path)):
+                        print(
+                            "Could not clean the Tezos node data directory. "
+                            "Please do so manually."
+                        )
+                        raise OSError(
+                            "'sudo rm -r " + os.path.join(node_dir, path) + "' failed."
+                        )
+
+                print("Node directory cleaned.")
+                return True
+            return False
+        return True
+
+    # Importing the snapshot for Node bootstrapping
+    def import_snapshot(self):
+        do_import = self.check_blockchain_data()
+
+        if do_import:
+
+            snapshot_file = ""
+            if self.config["snapshot"] == "skip":
+                return
+            elif self.config["snapshot"] == "file":
+                snapshot_file = self.config["snapshot_file"]
+            elif self.config["snapshot"] == "url":
+                url = self.config["snapshot_url"]
+                snapshot_file = fetch_snapshot(url)
+            elif self.config["snapshot"] == "download rolling":
+                url = "https://" + self.config["network"] + ".xtz-shots.io/rolling"
+                snapshot_file = fetch_snapshot(url)
+            elif self.config["snapshot"] == "download full":
+                url = "https://" + self.config["network"] + ".xtz-shots.io/full"
+                snapshot_file = fetch_snapshot(url)
+
+            proc_call(
+                "sudo -u tezos tezos-node-"
+                + self.config["network"]
+                + " snapshot import "
+                + snapshot_file
+            )
+
+            print("Snapshot imported.")
+
+            if (
+                self.config["snapshot"] == "download rolling"
+                or self.config["snapshot"] == "download full"
+            ):
+                try:
+                    os.remove("/tmp/tezos_node.snapshot")
+                except:
+                    pass
+                else:
+                    print("Deleted the downloaded snapshot.")
+
+    # Bootstrapping tezos-node
+    def bootstrap_node(self):
+
+        self.import_snapshot()
+
+        self.systemctl_start_action("node")
+        proc_call("sleep 5")
+
+        proc_call("sudo -u tezos tezos-client bootstrapped")
+
+        print()
+        print("The Tezos node bootstrapped successfully.")
+
+    # Importing the baker key
+    def import_baker_key(self):
+        if self.config["key_import_mode"] == "secret-key":
+            self.check_baker_account(
+                "sudo -u tezos tezos-client import secret key baker "
+                + self.config["secret_key"]
+            )
+        elif self.config["key_import_mode"] == "json":
+            self.check_baker_account(
+                "sudo -u tezos tezos-client activate account baker with "
+                + self.config["json_filepath"]
+            )
+
+        else:
+            print()
+            input("Please open the Tezos Baking app on your ledger, then hit Enter.")
+            self.check_baker_account(
+                "sudo -u tezos tezos-client import secret key baker "
+                + self.config["ledger_url"]
+            )
+            proc_call("sudo -u tezos tezos-client setup ledger to bake for baker")
+
+    def register_baker(self):
+        print()
+        if self.config["key_import_mode"] == "ledger":
+            input(
+                "Please open the Tezos Baking or Tezos Wallet app on your ledger, then hit Enter."
+            )
+
+        proc_call("sudo -u tezos tezos-client register key baker as delegate")
+        print(
+            "You can check a blockchain explorer (e.g. https://tzkt.io/ or https://tzstats.com/) "
+            "to see the baker status and baking rights of your account."
+        )
+
+    def start_baking(self):
+        self.systemctl_start_action("baking")
+
+    def systemctl_start_action(self, service):
+        if self.config["systemd_mode"] == "yes":
+            proc_call(
+                "sudo systemctl enable tezos-"
+                + service
+                + "-"
+                + self.config["network"]
+                + ".service"
+            )
+        proc_call(
+            "sudo systemctl start tezos-"
+            + service
+            + "-"
+            + self.config["network"]
+            + ".service"
+        )
+
+
+def run_setup(cfg):
+
+    if cfg is not None:
+        setup = Setup(cfg)
+
+        print("Trying to bootstrap tezos-node")
+        setup.bootstrap_node()
+
+        if cfg["mode"] == "node":
+            return
+
+        print("Importing the baker key")
+        setup.import_baker_key()
+
+        print("Registering the baker")
+        setup.register_baker()
+
+        print("Start baking instance")
+        setup.start_baking()
+
+        print(
+            "Congratulations! All required Tezos infrastructure services should now be started."
+            " You can show logs for all the services using the 'tezos' user by running: "
+            "journalctl -f _UID=$(id tezos -u)"
+        )
+
+        if cfg["mode"] == "baker":
+            print()
+            print("To stop the baking instance, run:")
+            print("sudo systemctl stop tezos-baking-" + cfg["network"] + ".service")
+
+            print()
+            print(
+                "If you previously enabled the baking service and want to disable it, run:"
+            )
+            print("sudo systemctl disable tezos-baking-" + cfg["network"] + ".service")
+
+
+if __name__ == "__main__":
+    readline.parse_and_bind("tab: complete")
+    readline.set_completer_delims(" ")
+
+    try:
+        cfg = run_wizard()
+    except KeyboardInterrupt:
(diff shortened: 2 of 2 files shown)


docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+                "sudo -u tezos tezos-client import secret key baker"
+                + self.config["ledger_url"]
+            )
+
+            print()
+            input("Please open the Tezos Baking app on your ledger, then hit Enter.")
+            proc_call("sudo -u tezos tezos-client setup ledger to bake for baker")
+
+    def register_baker(self):
+        proc_call("sudo -u tezos tezos-client register key baker as delegate")
```
rvem, line 248: AFAIR, baker registration also requires the Baking app to be open in case the baker key was imported from the ledger

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+class Setup:
+    def __init__(self, config):
+        self.config = config
+
+    # Bootstrapping tezos-node
+    def bootstrap_node(self):
+        if self.config["snapshot"]:
+            proc_call(
```
rvem, line 204: If node directory has some blockchain data, snapshot import will fail with the following error:
```
tezos-node: Error:
              Invalid data directory 'Please provide a clean directory by removing:
            context
            lock
            store'.
```

I think it makes sense to do the following:
1) Check if there is blockchain data in the tezos-node data directory.
2) If there is ask the user whether we should keep existing data or nuke it and import a new snapshot.

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+        return_code = None
+        while return_code != 0:
+            os.system("sleep 5")
+            return_code = subprocess.call(
+                shlex.split("sudo -u tezos tezos-client bootstrapped")
+            )
+
+    # Importing the baker key
+    def import_baker_key(self):
+        if self.config["key_import_mode"] == "secret-key":
```
rvem, line 224: Let's also check if there is already an account with alias `baker`. If there is, let's show its address to the user and ask whether it should be replaced with the new one

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+        setup.import_baker_key()
+
+        print("Registering the baker")
+        setup.register_baker()
+
+        print("Start baking instance")
+        setup.start_baking()
+
+        print("Checking the services status")
+        setup.check_systemd_status()
```
rvem, line 281: Does it actually do anything? :thinking:

rvem (commented):
Looks good, have some comments that are mostly about running the wizard subsequently multiple times in case the user accidentally provided some nonsense and wizard failed or when the user tries to run wizard over already configured baking instance

docker/package/model.py
```diff
@@ -584,9 +585,16 @@ def gen_makefile(self, out):
...
 .PHONY: install
 
+BINDIR=/usr/bin
+
 tezos-baking:
 
-install: tezos-baking
+baking-wizard:
+	mkdir -p $(DESTDIR)$(BINDIR)
+	cp $(CURDIR)/baking_wizard.py $(DESTDIR)$(BINDIR)/baking_wizard
```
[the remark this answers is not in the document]
rvem: Also, let's use snake-case to be consistent with the rest of the tezos-binaries

docker/package/model.py
```diff
@@ -584,9 +585,16 @@ def gen_makefile(self, out):
         file_contents = """
 .PHONY: install
 
+BINDIR=/usr/bin
+
 tezos-baking:
 
-install: tezos-baking
+baking-wizard:
```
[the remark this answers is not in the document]
rvem: Also, maybe it's worth adding `tezos-` prefix to the script name :thinking:

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+        self.prompt = prompt
+        self.help = help
+        self.default = default
+        self.validator = validator
+        self.completer = completer
+        self.reqs = reqs
+
+    def validate(self, input):
+        if self.validator is not None:
+            if isinstance(self.validator, list):
```
rvem, line 81: Looks kinda sus that `Step` should know something about validator structure. Let's move validation logic from the `Step` class, so that it simply calls something like `self.validator.validate(input)`

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+
+        proc_call(
+            "sudo systemctl start tezos-node-" + self.config["network"] + ".service"
+        )
+
+        return_code = None
+        while return_code != 0:
+            os.system("sleep 5")
+            return_code = subprocess.call(
+                shlex.split("sudo -u tezos tezos-client bootstrapped")
```
rvem, line 219: AFAIU, `tezos-client bootstrapped` process will wait until the node is bootstrapped while running the `tezos-client`, so the sleep loop looks redundant here

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+            )
+
+    # Importing the baker key
+    def import_baker_key(self):
+        if self.config["key_import_mode"] == "secret-key":
+            proc_call(
+                "sudo -u tezos tezos-client import secret key baker"
```
rvem, line 226:
```suggestion
                "sudo -u tezos tezos-client import secret key baker "
```
looks like space is missing, the same for the rest of import invocations

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+                + self.config["secret_key"]
+            )
+        elif self.config["key_import_mode"] == "json":
+            proc_call(
+                "sudo -u tezos tezos-client activate account baker with"
```
rvem, line 231:
```suggestion
                "sudo -u tezos tezos-client activate account baker with "
```

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+                + self.config["json_filepath"]
+            )
+
+        else:
+            print()
+            input("Please open the Tezos Wallet app on your ledger, then hit Enter.")
+            proc_call(
+                "sudo -u tezos tezos-client import secret key baker"
```
rvem, line 239:
```suggestion
                "sudo -u tezos tezos-client import secret key baker "
```

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+    def bootstrap_node(self):
+        if self.config["snapshot"]:
+            proc_call(
+                "sudo -u tezos tezos-node-"
+                + self.config["network"]
+                + " snapshot import "
+                + self.config["snapshot"]
+            )
+
+        proc_call(
```
rvem, line 211: Would be nice to inform user that node service was started:)

docker/package/baking_wizard.py
```diff
@@ -0,0 +1,303 @@
...
+            )
+
+
+if __name__ == "__main__":
+    readline.parse_and_bind("tab: complete")
+
+    try:
+        main()
+    except:
+        print("Exiting Tezos Baking Wizard")
```
rvem, line 303: Let's stop all started services in case the wizard exited with an error. My setup failed with baker import and I was confused by the fact that node was running after the failure

itkach:
I think general UX will be better if each step presents choices immediately, without teaching user to type help, along with the help text/explanation, with ability to choose it by typing choice id rather than spelling it out. For example, instead of

```
Which network would you like to use? <mainnet>
```

present something like this

```
Choose Tezos network:

1) mainnet       Main Tezos network

2) florencenet   Test network using version 8 of Tezos protocol
                 (Florence)
                 
3) granadanet    Test network using Granada  9 of Tezos protocol
                 (Granada)

> 1
```

I think using more vertical space, having each choice on a separate line, using vertical alignment (choice id column, choice value column, description column) as well as limiting overall width makes it more readable/ grockable at a glance.

Having tab completion in the input is nice, but not discoverable enough. With no input, it takes to tabs to get the completion, first tab just makes terminal bell sound, so my first impression was that there is no tab completion at all. I don't mind that it's there, but I don't think this is useful to people who don't live and breathe  cli and terminal based UIs.

itkach:
I think general UX will be better if each step presents choices immediately, without teaching user to type help, along with the help text/explanation, with ability to choose it by typing choice id rather than spelling it out. For example, instead of

```
Which network would you like to use? <mainnet>
```

present something like this

```
Choose Tezos network:

1) mainnet       Main Tezos network

2) florencenet   Test network using version 8 of Tezos protocol
                 (Florence)
                 
3) granadanet    Test network using Granada  9 of Tezos protocol
                 (Granada)

> 1
```

I think using more vertical space, having each choice on a separate line, using vertical alignment (choice id column, choice value column, description column) as well as limiting overall width makes it more readable/ grockable at a glance.

Having tab completion in the input is nice, but not discoverable enough. With no input, it takes to tabs to get the completion, first tab just makes terminal bell sound, so my first impression was that there is no tab completion at all. I don't mind that it's there, but I don't think this is useful to people who don't live and breathe  cli and terminal based UIs.

itkach:
I think it's critically important to handle snapshot download for the the user and not just leave them to type path to a file. In the spirit of previous UX suggestion, the choices should be clearly laid out right away:

```
It take a significant time to bootstrap Tezos node from scratch. 
Starting from a  snapshot is recommended. 
What would you like to do?

1) Import rolling snapshot from xtz-shots.io (recommended) 
2) Import full snapshot from xtz-shots.io
3) Import snapshot from a file/url
4) Skip snapshot import and synchronize with the network from scratch

> 1

```

Choice descriptions need to be informative but concise, so it's probably good to have even more information available about each choice. Perhaps typing `help` or `?` should display a full/detailed explanation text for current step.

This is also needed for private key setup - this is where main complexity is and  not everything can be automated (e.g. remote signer option or using freshly generated new key pair require a lot of explanation)

itkach:
Another great UX improvement would be colorizing the output (e.g. using a library such as https://github.com/tartley/colorama).

docker/package/baking_wizard.py
ixahedron: :thinking: It's not mentioned in the baking guide at any rate. If so, then that's something to fix there too

itkach: I am not sure this is true. Registration fee must be paid to register, and for that transaction must be signed (of type "transfer", I think), so if anything I would expect tezos wallet app should be running

ixahedron: I left it commented at this moment, but I'd really like to get confirmation if there should or shouldn't be a prompt for any app open on the ledger.

docker/package/model.py
```diff
@@ -502,6 +502,7 @@ def __init__(
 
     def fetch_sources(self, out_dir):
         os.makedirs(out_dir)
+        shutil.copy(f"{os.path.dirname(__file__)}/baking_wizard.py", out_dir)
```
rvem, line 505:
```suggestion
        shutil.copy(f"{os.path.dirname(__file__)}/tezos_baking_wizard.py", out_dir)
```

docker/package/model.py
```diff
@@ -584,9 +585,16 @@ def gen_makefile(self, out):
         file_contents = """
 .PHONY: install
 
+BINDIR=/usr/bin
+
 tezos-baking:
 
-install: tezos-baking
+tezos-baking-wizard:
+	chmod +x $(DESTDIR)$(BINDIR)/tezos-baking-wizard
```
rvem, line 593: hmm, this doesn't look right here :thinking:
should be something like: 
```suggestion
        mv $(CURDIR)/tezos_baking_wizard.py $(CURDIR)/tezos-baking-wizard
	chmod +x $(CURDIR)/tezos-baking-wizard
```

docker/package/model.py
```diff
@@ -584,9 +585,16 @@ def gen_makefile(self, out):
...
+
+install: tezos-baking tezos-baking-wizard
+	mkdir -p $(DESTDIR)$(BINDIR)
+	cp $(CURDIR)/tezos_baking_wizard.py $(DESTDIR)$(BINDIR)/tezos-baking-wizard
```
rvem, line 597: To complete my comment above
```suggestion
	cp $(CURDIR)/tezos-baking-wizard $(DESTDIR)$(BINDIR)/tezos-baking-wizard
```

rvem:
Something went wrong after snapshot import:(
```
The Tezos node can take a significant time to bootstrap from scratch.
Bootstrapping from a snapshot is suggested instead.
How would you like to proceed?

1. download rolling      Import rolling snapshot from xtz-shots.io (recommended)
2. download full         Import full snapshot from xtz-shots.io
3. file                  Import snapshot from a file
4. url                   Import snapshot from a url
5. skip                  Skip snapshot import and synchronize with the network from
                         scratch

> 1
Trying to bootstrap tezos-node
Downloading the snapshot from  https://florencenet.xtz-shots.io/rolling
Progress:  100 %,  148 MB
The Tezos node data directory already has some blockchain data:
store
context
peers.json
identity.json
lock

Delete this data and import a new snapshot? <y/N> y
Exiting Tezos Baking Wizard
```

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,466 @@
...
+# Wizard CLI utility
+
+
+def proc_call(cmd):
+    return subprocess.check_call(shlex.split(cmd))
+
+
+def fetch_snapshot(url):
+    print("Downloading the snapshot from ", url)
+    filename = "tezos_node.snapshot"
```
rvem, line 68: Looks like it will download the snapshot to the current directory (from which this wizard is called), thus `tezos-node` won't have sufficient rights to use this file, I suggest downloading it to the `/tmp`
```suggestion
    filename = "/tmp/tezos_node.snapshot"
```

rvem:
Looks weird:
```
How do you want to import the baker key?
{'ledger': 'From a ledger', 'secret-key': 'Either the unencrypted or password-encrypted secret key for your address', 'json': 'Faucet JSON file from https://faucet.tzalpha.net/'}
> secret-key
Validation error: Please choose one of the provided values: 1, 2, 3
```

rvem:
Something goes wrong during node directory clean up before snapshot import:
```
Trying to bootstrap tezos-node
Downloading the snapshot from  https://florencenet.xtz-shots.io/rolling
Progress:  100 %,  148 MB
The Tezos node data directory already has some blockchain data:
context
peers.json
identity.json
store

Delete this data and import a new snapshot? <y/N> y
Exiting Tezos Baking Wizard
```
I suppose that the user who runs the script doesn't have sufficient rights to remove files from `/var/lib/tezos/<network>-node`

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,466 @@
...
+
+        proc_call("sudo -u tezos tezos-client bootstrapped")
+
+        print()
+        print("The Tezos node bootstrapped successfully.")
+
+    # Importing the baker key
+    def import_baker_key(self):
+        if self.config["key_import_mode"] == "secret-key":
+            check_baker_account(
```
rvem, line 374: I believe, `self` is missed here
```suggestion
            self.check_baker_account(
```

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,466 @@
...
+        elif self.config["key_import_mode"] == "json":
+            check_baker_account(
+                "sudo -u tezos tezos-client activate account baker with "
+                + self.config["json_filepath"]
+            )
+
+        else:
+            print()
+            input("Please open the Tezos Wallet app on your ledger, then hit Enter.")
+            check_baker_account(
```
rvem, line 387:
```suggestion
            self.check_baker_account(
```

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,466 @@
...
+
+    # Importing the baker key
+    def import_baker_key(self):
+        if self.config["key_import_mode"] == "secret-key":
+            check_baker_account(
+                "sudo -u tezos tezos-client import secret key baker "
+                + self.config["secret_key"]
+            )
+        elif self.config["key_import_mode"] == "json":
+            check_baker_account(
```
rvem, line 379:
```suggestion
            self.check_baker_account(
```

docker/package/model.py
```diff
@@ -584,9 +585,17 @@ def gen_makefile(self, out):
         file_contents = """
 .PHONY: install
 
+BINDIR=/usr/bin
+
 tezos-baking:
 
-install: tezos-baking
+tezos-baking-wizard:
+	chmod +x $(CURDIR)/tezos-baking-wizard
```
rvem, line 593: apparently, this and the line below should be swapped

docker/package/baking_wizard.py
rvem: I confirm that a response on the ledger is required to run `tezos-client register key baker as delegate`. Both Baking and Wallet app can be used to confirm this operation

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,487 @@
...
+    return values
+
+
+class Setup:
+    def __init__(self, config):
+        self.config = config
+
+    # Check if an account with the 'baker' alias already exists, and ask the user
+    # if it can be overwritten.
+    def check_baker_account(cmd):
```
rvem, line 292:
```suggestion
    def check_baker_account(self, cmd):
```

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,487 @@
...
+            self.check_baker_account(
+                "sudo -u tezos tezos-client import secret key baker "
+                + self.config["ledger_url"]
+            )
+
+            print()
+            input("Please open the Tezos Baking app on your ledger, then hit Enter.")
+            proc_call("sudo -u tezos tezos-client setup ledger to bake for baker")
+
+    def register_baker():
```
rvem, line 417:
```suggestion
    def register_baker(self):
```

rvem (commented):
A bit more suggestions:)

With suggested changes, I was able to test faucet and secret-key based bakers, however, ledger interaction currently doesn't work with the systemd services, see #210:(

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,487 @@
...
+        print(
+            "Congratulations! All required Tezos infrastructure services should now be started."
+            " You can show logs for all the services using the 'tezos' user by running: "
+            "journalctl -f _UID=$(id tezos -u)"
+        )
+
+        if cfg["mode"] == "baker":
+            print()
+            print(
+                "To stop the baking instance, run: sudo systemctl stop tezos-baking"
```
rvem, line 462:
```suggestion
                "To stop the baking instance, run: sudo systemctl stop tezos-baking-"
```

README.md
```diff
@@ -288,6 +288,9 @@ This repository provides two distinct ways for building and packaging tezos bina
 Read [the article](./docs/baking.md) to find out an easy way to set up
 baking instance on Ubuntu using packages provided by our launchpad PPA.
 
+You can also use our interactive CLI utility, `baking_wizard`, that comes
```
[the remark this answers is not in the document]
pasqu4le: I was going to comment this same thing.

One more addition is that since we'll have to make this update later it would be nice for this PR not to close the issue.

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,489 @@
...
+    return filename
+
+
+def progressbar_hook(chunk_number, chunk_size, total_size):
+    done = chunk_number * chunk_size
+    percent = min(int(done * 100 / total_size), 100)
+    print("Progress: ", percent, "%, ", int(done / (1024 * 1024)), "MB", end="\r")
+
+
+def pprint_table(options):
```
pasqu4le, line 86: It looks like this is always called before assigning a "table" to the `options` of a `Step`.

Why not move this to be method of the `Step` class, say `print_options` that will:
1. check for `None` and print nothing in that case.
2. use the logic defined here to print the steps otherwise

Additionally I'd suggest to put a `(default)` near the default option to make clear what happens when a user presses Enter.

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,489 @@
...
+            elif answer.lower() in ["help", "?"]:
+                print(step.help)
+            else:
+                if not answer and step.default is not None:
+                    answer = step.default
+
+                try:
+                    answer = step.validator.validate(answer)
+                except ValueError as e:
+                    print(color("Validation error: " + str(e), "\x1b[1;31m"))
```
pasqu4le, line 278: likewise I think that most of this "handling validation" logic could be moved to a `validate` method of `Step`.

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,489 @@
...
+        self.options = options
+        self.validator = validator
+        self.reqs = reqs
+
+
+def run_wizard(networks):
+    print("Tezos Baking Wizard")
+    print()
+    print(
+        "Welcome to a wizard helping you to set up the infrastructure for interacting with the Tezos blockchain."
```
pasqu4le, line 126:
```suggestion
        "Welcome, this wizard will help you to set up the infrastructure to interact with the Tezos blockchain."
```

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,489 @@
...
+    )
+    print(
+        "In order to run a baking instance, you'll need the following Tezos binaries:",
+        "tezos-client, tezos-node, tezos-baker-<proto>, tezos-endorser-<proto>.",
+    )
```
pasqu4le, line 131: Perhaps we should mention that if one has installed `tezos-baking` they are all already present?

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,489 @@
...
+
+    steps = [
+        Step(
+            id="network",
+            prompt="Which Tezos network would you like to use?\nCurrently supported:\n",
+            help="The selected network will be used to set up all required services.\n"
+            "The currently supported protocol is 009-PsFLoren (used on florencenet and mainnet).\n"
+            "Also, note that the corresponding packages have protocol suffix in lowercase, "
+            "e.g. the list of available baker packages can be found on Serokell's launchpad, i.e. here:\n"
+            "https://launchpad.net/~serokell/+archive/ubuntu/tezos/+packages?field.name_filter=tezos-baker&field.status_filter=published"
```
pasqu4le, line 168: I'm not sure that this is useful information here, if one is using this script has at least install some package and the script does not instruct (or run commands) to install more, so... :man_shrugging:

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,489 @@
...
+            "on baking with a faucet JSON file.",
+            options=pprint_table(networks),
+            validator=Validator(enum_range_validator(networks)),
+        ),
+        Step(
+            id="mode",
+            prompt="Do you want to set up baking or to run the standalone node?\n",
+            help="By default, tezos-baking provides predefined services for running baking instances "
+            "on different networks. Sometimes, however, you might want to only run the Tezos node. "
+            "When this option is chosen, this wizard will help you bootstrap tezos-node only.",
```
pasqu4le, line 179: Perhaps some `/n` would looks good here

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,489 @@
...
+            "By default, service with tezos-node will start to bootstrap from scratch, "
+            "which will take a significant amount of time. In order to avoid this, we suggest "
+            "bootstrapping from a snapshot instead.\n\n"
+            "Snapshots can be downloaded from the following websites:\n"
+            "Tezos Giganode Snapshots - https://snapshots-tezos.giganode.io/ \n"
+            "XTZ-Shots - https://xtz-shots.io/ \n"
+            "We recommend to use rolling snapshots. This is the smallest and the fastest mode "
+            "that is sufficient for baking (you can read more about other tezos-node history modes here: "
+            "https://tezos.gitlab.io/user/history_modes.html#history-modes ).\n"
+            "All commands within the service are run under the 'tezos' user.",
```
pasqu4le, line 198: Also, probably not very useful outside of the article, as we run all of the commands ourselves.

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,489 @@
...
+
+                print("Node directory cleaned.")
+
+                return True
+
+        return False
+
+    # Importing the snapshot for Node bootstrapping
+    def import_snapshot(self):
+        do_import = self.check_blockchain_data(snapshot_file)
```
pasqu4le, line 340: :thinking: currently I get an error here, due to the fact that `snapshot_file` is undefined.

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,489 @@
...
+if __name__ == "__main__":
+    readline.parse_and_bind("tab: complete")
+
+    networks = {
+        "mainnet": "Main Tezos network",
+        "florencenet": "Test network using version 9 of Tezos protocol (Florence)",
+    }
+
+    try:
+        cfg = run_wizard(networks)
```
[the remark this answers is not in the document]
ixahedron: Not currently, so I agree with you there. I initially thought it might be possible to share the list of networks from the configuration, e.g. `packages.py`, but that's not true/easy from the way it will be used.

docker/package/tezos_baking_wizard.py
ixahedron: That's how it was initially, see https://github.com/serokell/tezos-packaging/pull/206#pullrequestreview-667534646

ixahedron: So, you know, tell me which one of you you think knows better :smile:

docker/package/tezos_baking_wizard.py
ixahedron: I don't know, the user still has to choose their preferred snapshot mode, and the info might be useful for the (maybe far-fetched) case when the download size matters. So I'd leave this in for now :shrug:

docker/package/model.py
```diff
@@ -584,9 +585,17 @@ def gen_makefile(self, out):
...
 tezos-baking:
 
-install: tezos-baking
+tezos-baking-wizard:
+	mv $(CURDIR)/tezos_baking_wizard.py $(CURDIR)/tezos-baking-wizard
+	chmod +x $(CURDIR)/tezos-baking-wizard
+
+install: tezos-baking tezos-baking-wizard
+	mkdir -p $(DESTDIR)$(BINDIR)
+	cp $(CURDIR)/tezos_baking_wizard $(DESTDIR)$(BINDIR)/tezos-baking-wizard
```
rvem, line 598:
```suggestion
	cp $(CURDIR)/tezos-baking-wizard $(DESTDIR)$(BINDIR)/tezos-baking-wizard
```

docker/package/tezos_baking_wizard.py
rvem: :joy:

rvem: I don't have a really strong opinion about this, but IMO it's better when validator logic is hidden inside the validator, i.e. we don't know whether the validator accepts only a fixed list of options or only filepaths or something different

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,521 @@
...
+                + self.config["ledger_url"]
+            )
+
+            print()
+            input("Please open the Tezos Baking app on your ledger, then hit Enter.")
+            proc_call("sudo -u tezos tezos-client setup ledger to bake for baker")
+
+    def register_baker(self):
+        print()
+        input(
```
rvem, line 431: Looks like, we should only prompt this in case baker secret-key is on the ledger

rvem (approved):
Don't have any other suggestions, LGTM

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,522 @@
...
+    except EOFError:
+        print()
+        print("Exiting Tezos Baking Wizard")
+        sys.exit(1)
+    except Exception as e:
+        print("Error in Tezos Baking Wizard, exiting.")
+        logfile = "tezos_baking_wizard.log"
+        with open(logfile, "a") as f:
+            f.write(str(e) + "\n")
+            print("The error has been logged to", logfile)
```
pasqu4le, line 502: Perhaps this should be out of the `with` block?

Also, I suppose it would be better if it printed the absolute path.

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,522 @@
...
+        proc_call("sudo systemctl stop tezos-baking-" + cfg["network"] + ".service")
+        print("Error in Tezos Baking setup, exiting.")
+        sys.exit(1)
+    except:
+        proc_call("sudo systemctl stop tezos-baking-" + cfg["network"] + ".service")
+        print("Error in Tezos Baking setup, exiting.")
+        logfile = "tezos_baking_wizard.log"
+        with open(logfile, "a") as f:
+            f.write(str(e) + "\n")
+            print("The error has been logged to", logfile)
```
pasqu4le, line 521: Same as the previous comment

docker/package/tezos_baking_wizard.py
pasqu4le: :ok_hand: fair enough

pasqu4le: :ok_hand: fair enough

pasqu4le (commented):
My only other suggestion is to add another option/step, which is to choose between just `start`ing the service or also `enable`-ing them, which I think would be useful.

docker/package/tezos_baking_wizard.py
pasqu4le: Oh, my comment was unclear, I actually agree with that.

My suggestion was only about moving the handling this `try` and the `validated` variable inside the `step`, which isn't related to the validator itself, but to the validation process in general.

Now that I looks at it again, I suppose one way to do that would be to start without a:
```python
validated = False
while not validated
```
and instead use a:
```python
answer = None
while answer is None
```
and instead of the `try` block call:
```python
answer = step.validate(answer)
```
which actually contains this `try` and returns `None` on failure.

It's not much better TBH, it may even be worse to follow, I don't have a strong opinion either and I'll let you decide :slightly_smiling_face:

ixahedron: I tried it out, and I believe it's actually much worse to follow. With validation logic and the _validator_ logic separate, it now comes to calling `step.validate` which in turn calls `step.validator.validate`, and also `values[step.id]` is constantly overwritten until the condition in the `while` clause stops to be met (and the condition is not as immediately understandable, I believe). I don't think we would gain much from this separation, but it seems we do lose some at-a-glance readability :/ So I'll leave it as is for now

itkach:
trying latest version, I see question "Do you want to enable the tezos-node service or just start it?" and the same for `tezos-baking`, with default choice being "start". What are the considerations for these choices? I personally expect all the services to be started for me and run forever, including automatically after reboot, so I wouldn't event ask these questions in the wizard. Also, isn't starting `tezos-baking` also starts `tezos-node`, so asking about `tezos-node` when setting up baking is redundant anyway? One situation I can think of where starting baker and endorser services on boot may be undesirable is if baking key is encrypted. Node should start automatically regardless. Power users that disagree with these choices can easily reconfigure using standard systemd facilities, but most users shouldn't be burdened with these details.

itkach:
when user chooses to import key from ledger, they should be presented with choice from the list of tz addresses, one for each hashing algo corresponding to default derivation, along with their balances (and ability to specify custom derivation path), similar to how Kiln web UI presents it.
Here's, for example, how the relevant info may be extracted from tezos-client output: https://gist.github.com/itkach/dee1aac56c74b20097aad761a28c6f2e

itkach:
first time I tried, snapshot download was silently skipped... maybe I just hit enter and empty input was not interpreted to mean "choose the default"?

itkach:
After starting Tezos Wallet app on the ledger and hitting Enter, I get this error:
```
Error:
  Erroneous command line argument 6 (baker).
  Cannot parse host of URI: baker

Usage:
  tezos-client [global options] command [command options]
  tezos-client --help (for global options)
  tezos-client [global options] command --help (for command options)
  tezos-client --version (for version information)

To browse the documentation:
  tezos-client [global options] man (for a list of commands)
  tezos-client [global options] man -v 3 (for the full manual)

Global options (must come before the command):
  -d --base-dir <path>: client data directory
  -c --config-file <path>: configuration file
  -t --timings: show RPC request times
  --chain <hash|tag>: chain on which to apply contextual commands (possible tags are 'main' and 'test')
  -b --block <hash|tag>: block on which to apply contextual commands (possible tags are 'head' and 'genesis')
  -w --wait <none|<int>>: how many confirmation blocks before to consider an operation as included
  -p --protocol <hash>: use commands of a specific protocol
  -l --log-requests: log all requests to the node
  -A --addr <IP addr|host>: [DEPRECATED: use --endpoint instead] IP address of the node
  -P --port <number>: [DEPRECATED: use --endpoint instead] RPC port of the node
  -S --tls: [DEPRECATED: use --endpoint instead] use TLS to connect to node.
  -E --endpoint <uri>: HTTP(S) endpoint of the node RPC interface; e.g. 'http://localhost:8732'
  -R --remote-signer <uri>: URI of the remote signer
  -f --password-filename <filename>: path to the password filename
  -M --mode <client|mockup|proxy>: how to interact with the node
  File "tezos_baking_wizard.py", line 557, in <module>
    run_setup(cfg)
  File "tezos_baking_wizard.py", line 508, in run_setup
    setup.import_baker_key()
  File "tezos_baking_wizard.py", line 450, in import_baker_key
    proc_call("sudo -u tezos tezos-client setup ledger to bake for baker")
  File "tezos_baking_wizard.py", line 69, in proc_call
    return subprocess.check_call(shlex.split(cmd))
  File "/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['sudo', '-u', 'tezos', 'tezos-client', 'setup', 'ledger', 'to', 'bake', 'for', 'baker']' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

  File "tezos_baking_wizard.py", line 571, in <module>
    f.write(str(e) + "\n")
NameError: name 'e' is not defined
```

itkach:
there should be no need to start wallet app first, it should work just with baker app
with baker app running, hitting enter on first prompt ("Please open the Tezos Wallet app on your ledger, then hit Enter.") moves to the second one, hitting enter in second one give me this error:

```
Node is bootstrapped.

The Tezos node bootstrapped successfully.
Importing the baker key

Please open the Tezos Wallet app on your ledger, then hit Enter.

Please open the Tezos Baking app on your ledger, then hit Enter.
Warning:

                 This is NOT the Tezos Mainnet.

           Do NOT use your fundraiser keys on this network.

Error:
  Erroneous command line argument 6 (baker).
  Cannot parse host of URI: baker

Usage:
  tezos-client [global options] command [command options]
  tezos-client --help (for global options)
  tezos-client [global options] command --help (for command options)
  tezos-client --version (for version information)

To browse the documentation:
  tezos-client [global options] man (for a list of commands)
  tezos-client [global options] man -v 3 (for the full manual)

Global options (must come before the command):
  -d --base-dir <path>: client data directory
  -c --config-file <path>: configuration file
  -t --timings: show RPC request times
  --chain <hash|tag>: chain on which to apply contextual commands (possible tags are 'main' and 'test')
  -b --block <hash|tag>: block on which to apply contextual commands (possible tags are 'head' and 'genesis')
  -w --wait <none|<int>>: how many confirmation blocks before to consider an operation as included
  -p --protocol <hash>: use commands of a specific protocol
  -l --log-requests: log all requests to the node
  -A --addr <IP addr|host>: [DEPRECATED: use --endpoint instead] IP address of the node
  -P --port <number>: [DEPRECATED: use --endpoint instead] RPC port of the node
  -S --tls: [DEPRECATED: use --endpoint instead] use TLS to connect to node.
  -E --endpoint <uri>: HTTP(S) endpoint of the node RPC interface; e.g. 'http://localhost:8732'
  -R --remote-signer <uri>: URI of the remote signer
  -f --password-filename <filename>: path to the password filename
  -M --mode <client|mockup|proxy>: how to interact with the node
  File "tezos_baking_wizard.py", line 557, in <module>
    run_setup(cfg)
  File "tezos_baking_wizard.py", line 508, in run_setup
    setup.import_baker_key()
  File "tezos_baking_wizard.py", line 450, in import_baker_key
    proc_call("sudo -u tezos tezos-client setup ledger to bake for baker")
  File "tezos_baking_wizard.py", line 69, in proc_call
    return subprocess.check_call(shlex.split(cmd))
  File "/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['sudo', '-u', 'tezos', 'tezos-client', 'setup', 'ledger', 'to', 'bake', 'for', 'baker']' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

  File "tezos_baking_wizard.py", line 571, in <module>
    f.write(str(e) + "\n")
NameError: name 'e' is not defined
```

pasqu4le:
> "Do you want to enable the tezos-node service or just start it?"

I agree that this is not the best way to give the choice, but I actually would give the option, just in the off-chance that one is just trying things out and doesn't actually expect resources to be taken on reboot again.

IMO it should be rephrased in a way that's as simple as possible, like:
```
Would you like your setup to automatically start on boot?
```
where the default is "yes".

The `help` menu for it can contain more details about the `enable`-ing vs `start`-ing of systemd services.

pasqu4le:
> first time I tried, snapshot download was silently skipped... 

Confirming: same happened to me on the latest version (id doesn't seem to be caused by the default choice tho), needs to be solved.

itkach:
> > "Do you want to enable the tezos-node service or just start it?"
> 
> I agree that this is not the best way to give the choice, but I actually would give the option, just in the off-chance that one is just trying things out and doesn't actually expect resources to be taken on reboot again.
> 

The wizard is supposed to do the right thing for most people. The right thing for most people is to run the services forever - if your key does bot require user to input password, in which case the right thing is the opposite (and the user must be informed that they are on the hook for manually starting the services and entering password). Why should not this be codified?

> IMO it should be rephrased in a way that's as simple as possible, like:
> 
> ```
> Would you like your setup to automatically start on boot?
> ```
> 
> where the default is "yes".
> 

I agree that this is a better wording if we must have this question at all.

itkach:
looks like I was affected by #210

Even so, error handling and user messages need to be more robust. There may be other reasons why tezos-client doesn't see connected ledger (ledger is actually NOT connected, we are inside in a VM and USB connection is not configured in VM or is acting up, ledger is not unlocked, bad USB port, bad USB cable, etc.).

Also, running the wizard again should handle existing config/previously made choices more gracefully. If node is already running and boostrapped on the network of choice - don't ask me again to import snapshot, if baker key is already imported let me skip that step.

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,590 @@
...
+        print(
+            "You can check a blockchain explorer (e.g. https://tzkt.io/ or https://tzstats.com/) "
+            "to see the baker status and baking rights of your account."
+        )
+
+    def start_baking(self):
+        self.systemctl_start_action("baking")
+
+    def systemctl_start_action(self, service):
+        if self.config["systemd_enable"] == "both":
```
rvem, line 496: Looks like no such key error occurs here

pasqu4le (commented):
Much better, the script seems much more robust (judging from the experience, not the code) from the tests I did

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,592 @@
...
+    # Check if an account with the 'baker' alias already exists, and ask the user
+    # if it can be overwritten.
+    def check_baker_account(self, cmd):
+        output = subprocess.run(shlex.split(cmd), capture_output=True)
+        if output.returncode and re.search(
+            b"The secret_key alias baker already exists.", output.stderr
+        ):
+            address = re.search(b"The current value is (.+)$", output.stderr).group(1)
+            print("An account with the 'baker' alias already exists.")
+            print("Its address is ", address.decode("utf-8"))
```
rvem, line 369: btw, it'll show the secret key instead of address

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,592 @@
...
+        self.config = config
+
+    # Check if an account with the 'baker' alias already exists, and ask the user
+    # if it can be overwritten.
+    def check_baker_account(self, cmd):
+        output = subprocess.run(shlex.split(cmd), capture_output=True)
+        if output.returncode and re.search(
+            b"The secret_key alias baker already exists.", output.stderr
+        ):
+            address = re.search(b"The current value is (.+)$", output.stderr).group(1)
```
rvem, line 367:
```suggestion
            address = re.search(b"The current value is (.+)$", output.stderr, flags=re.MULTILINE).group(1)
```

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,596 @@
...
+        self.config = config
+
+    # Check if an account with the 'baker' alias already exists, and ask the user
+    # if it can be overwritten.
+    def check_baker_account(self, cmd):
+        output = subprocess.run(shlex.split(cmd), capture_output=True)
+        if output.returncode and re.search(
+            b"The secret_key alias baker already exists.", output.stderr
+        ):
+            value = re.search(b"The current value is (.+)\.", output.stderr).group(1)
```
rvem, line 367: :+1:

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,599 @@
...
+            address = subprocess.run(
+                shlex.split("sudo -u tezos tezos-client show address baker"),
+                capture_output=True,
+            )
+            if address.returncode == 0:
+                address = re.search(b"Hash: (.+)\\n", address.stdout).group(1)
```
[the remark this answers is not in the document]
pasqu4le: Note, something similar could be done for all regex parsing places, I think

docker/package/tezos_baking_wizard.py
```diff
@@ -0,0 +1,599 @@
...
+            print("An account with the 'baker' alias already exists.")
+            print("Its current value is ", value.decode("utf-8"))
+            print("With a corresponding address: ", address.decode("utf-8"))
+
+            replace = input("Should it be replaced with the new import? <y/N> ").strip()
+            if replace.lower() in ["y", "yes"]:
+                proc_call(cmd + " --force")
```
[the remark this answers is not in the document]
ixahedron: Agree. idk about the empty choice though, as it makes sense to me to support a default for that. In this case the default is explicitly "no".

docker/package/tezos_baking_wizard.py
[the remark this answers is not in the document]
ixahedron: FWIW I don't think changing this match here is going to bring us much or indeed that the wording is going to change soon. But it does make sense, so sure.

However, what do you mean by other places? There are four regex searches in the wizard code:
1) one already parses ledger urls which aren't likely to change
2) one is this one (already covered)
3) one ("The secret_key alias baker already exists.") is used to distinguish between different failing modes when importing keys, so we're going to need to provide an up-to-date output like that anyway (we could rely on return codes, but I believe that'd be even less permanent than stderr output).
4) And the last, used in `value`, is supposed to parse both ledger urls and secret keys, so it's easier to do this using the surrounding text than to assemble a giant regex (nobody would ever be able to read, probably). Alternatively, we can remove the "current value" output completely and only show the address, although I believe at least when using ledger it's much easier to remember the mnemonic url than a Tezos address.
I'm wondering what you think about all of these and whether we should change any.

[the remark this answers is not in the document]
pasqu4le: Ah you are right, I should have checked better.
1. For a moment I tough that the `ledger:` was part of the url :facepalm: it's ok as-is indeed
2. _covered_
3. I forgot about that, indeed it should be fine as differentiation (or rather alternatives are similar, as you said).
4. Well, admittedly that would be _one_ big regex, but it's only one of those two options, so it could be the combination of two smaller regex: one that you already have (ledger url) and the other one which would be to add.
    I also think that if not now these could be useful (the one for the secret key, in this case) either way to have.

docker/package/tezos_baking_wizard.py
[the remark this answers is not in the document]
pasqu4le: fair enough :ok_hand:

pasqu4le (approved):
LGTM now

PR #208, opened 2021-05-26
[#195] Provide key attributes for dependency steps

### LINKED ISSUE #195: Use smarter pipelines on CI

rvem:
## Description
Problem: Turned out buildkite uses 'key' attribute in order to determine
steps dependencies instead of using 'label'.

Solution: Explicitly add keys to steps that used as a dependency.

## Related issue(s)

Resolves #195

  can be successfully built.

### DIFF
--- a/.buildkite/filter-pipeline.py
+++ b/.buildkite/filter-pipeline.py
@@ -70,7 +70,19 @@ def dfs(name, depends_on_dict, visited):
             dfs(adjacent, depends_on_dict, visited)
 
 
-def build_tree(steps):
+# Buildkite uses 'key' attribute to determine dependencies of steps.
+# Existing code uses 'label's when analyzing dependency tree, so that
+# it's possible to analyze reverse dependencies as well.
+# This function builds a dict that maps 'key's to 'label's
+def build_key_to_label(steps):
+    key_to_label = {}
+    for step in steps:
+        if "key" in step and "label" in step:
+            key_to_label[step["key"]] = step["label"]
+    return key_to_label
+
+
+def build_tree(steps, key_to_label):
     depends_on_dict = {}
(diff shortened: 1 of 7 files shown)


.buildkite/filter-pipeline.py
```diff
@@ -70,24 +70,34 @@ def dfs(name, depends_on_dict, visited):
             dfs(adjacent, depends_on_dict, visited)
 
 
-def build_tree(steps):
+def build_key_to_label(steps):
+    key_to_label = {}
+    for step in steps:
+        if "key" in step and "label" in step:
+            key_to_label[step["key"]] = step["label"]
+    return key_to_label
```
pasqu4le, line 78: I'm not sure I understand the need for this, can you explain?

Given that this is based on buildkite's `depends_on`, which in turn uses `key`s, shouldn't the logic of this script change to use `key`s as well (and not use `label`s at all)?

rvem: only steps that used in `depends_on` have to have `key` attribute, while we still want to visit all reverse dependencies as well, which don't have `key`, thus it's easier to navigate through dependency tree using `label`

pasqu4le: Oh I see, got it.

Can you please add a comment to this function saying why we need it then?

pasqu4le (approved):

PR #209, opened 2021-06-01
v9.2 release

rvem:
## Description
The new Tezos release arrived, we're updating everything as usual.

Note, that native packages are waiting for opam-repository to be updated.

## Related issue(s)

Resolves None

  can be successfully built.

### DIFF
--- a/Formula/tezos-node-granadanet.rb
+++ b/Formula/tezos-node-granadanet.rb
@@ -1,0 +1,89 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+class TezosNodeGranadanet < Formula
+  url "file:///dev/null"
+  version "v9.2-1"
+
+  bottle :unneeded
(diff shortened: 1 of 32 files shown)


Formula/tezos-node-granadanet.rb
```diff
@@ -0,0 +1,89 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+class TezosNodeFlorencenet < Formula
```
pasqu4le, line 5: Looks like this should be changed

pasqu4le (commented):
Only one comment, also please use the `Problem`/`Solution` format in all the commits :slightly_smiling_face:

Formula/tezos-node-granadanet.rb
rvem: right

rvem: Fixed, but I accidentally squashed changes with the latest commit:(

pasqu4le (approved):

PR #223, opened 2021-06-16
[#220] Fix checking node before it's started

### LINKED ISSUE #220: [BUG] tezos-baking-wizard exits with error when setting up mainnet node
# Description

- Run tezos-baking-wizard
- Choose to set up mainnet node only
- Observe wizard exits with error

```
Delete this data and bootstrap the node again? <y/N> y
Node directory cleaned.
Downloading the snapshot from  https://mainnet.xtz-shots.io/rolling
Progress:  100 %,  1383 MB
Jun 15 17:54:04.314 - snapshot: Importing data from snapshot file /tmp/tezos_node.snapshot
Jun 15 17:54:04.314 - snapshot: You may consider using the --block <block_hash> argument to verify that the block imported is the one you expected
Jun 15 17:54:04.315 - snapshot: Retrieving and validating data. This can take a while, please bear with us
Context: 17243K elements, 1383MiB read
Store: 0K elements, 1383MiB read
Computing predecessors table 0K elements
Jun 15 18:36:53.824 - snapshot: Setting current head to block BLhExgpm1tpdqQxoWPjJ9v13Se1JuQk7CNjMEhNjDt9XteyyBFM
Jun 15 18:36:55.832 - snapshot: Setting history-mode to rolling
Jun 15 18:37:36.230 - snapshot: Successful import from file /tmp/tezos_node.snapshot
Snapshot imported.
Deleted the downloaded snapshot.
Created symlink /etc/systemd/system/multi-user.target.wants/tezos-node-mainnet.service → /lib/systemd/system/tezos-node-mainnet.service.
Warning:
  Failed to acquire the protocol version from the node
  Rpc request failed:
     - meth: GET
     - uri: http://localhost:8732/chains/main/blocks/head/protocols
     - error: Unable to connect to the node: "Unix.Unix_error(Unix.ECONNREFUSED, "connect", "")"

Error:
  Rpc request failed:
     - meth: GET
     - uri: http://localhost:8732/monitor/bootstrapped
     - error: Unable to connect to the node: "Unix.Unix_error(Unix.ECONNREFUSED, "connect", "")"
The error has been logged to /home/pi/tezos_baking_wizard.log
pi@raspberrypi:~ $ cat /home/pi/tezos_baking_wizard.log
Command '['sudo', '-u', 'tezos', 'tezos-client', 'bootstrapped']' returned non-zero exit status 1.
```

# Environment
- Raspberry Pi OS

ixahedron:
## Description

Problem: starting up the tezos-node service takes longer on some machines,
presumably mostly Raspberry Pi's. As we just wait for 5 seconds to allow
the service to start, this leads to the setup wizard checking if the node
has been bootstrapped before the service has even started, so the wizard
crashes because of an ECONNREFUSED error from tezos-client.

Solution: switched from waiting for 5 seconds to checking by requesting
via the RPC interface, every second until the node responds.

## Related issue(s)

Resolves #220

  can be successfully built.

### DIFF
--- a/docker/package/tezos_baking_wizard.py
+++ b/docker/package/tezos_baking_wizard.py
@@ -132,6 +132,36 @@ def list_connected_ledgers():
     return [name.decode() for name in re.findall(ledger_regex, output.stdout)]
 
 
+def get_node_rpc_addr(network):
+    output = subprocess.run(
+        shlex.split("systemctl show tezos-node-" + network + ".service"),
+        capture_output=True,
+    ).stdout
+    config = re.search(b"Environment=(.*)(?:$|\n)", output)
+    if config is None:
+        print(
+            "tezos-node-" + network + ".service configuration not found, "
+            "defaulting to 'http://localhost:8732'"
+        )
+        return "http://localhost:8732"
+    config = config.group(1)
+    address = re.search(b"NODE_RPC_ADDR=(.*?)(?: |$|\n)", config)
+    if address is not None:
+        address = address.group(1).decode("utf-8")
+    else:
+        print("NODE_RPC_ADDR is undefined, defaulting to 'localhost:8732'")
+        address = "localhost:8732"
+
+    mode = "https://"
+    if (
+        re.search(b"KEY_PATH=(?: |$|\n)", config) is not None
+        or re.search(b"CERT_PATH=(?: |$|\n)", config) is not None
+    ):
+        mode = "http://"
+
+    return mode + address
+
+
 class Step:
     def __init__(
         self,
@@ -468,7 +498,15 @@ class Setup:
         self.import_snapshot()
 
         self.systemctl_start_action("node")
-        proc_call("sleep 5")
+
+        while True:
+            rpc_address = get_node_rpc_addr(self.config["network"])
+            try:
+                urllib.request.urlopen(rpc_address + "/version")
+                break
+            except urllib.error.URLError:
+                print("...Waiting for the node service to start...")
+                proc_call("sleep 1")
 
         proc_call("sudo -u tezos tezos-client bootstrapped")
 


docker/package/tezos_baking_wizard.py
```diff
@@ -468,7 +468,16 @@ def bootstrap_node(self):
         self.import_snapshot()
 
         self.systemctl_start_action("node")
-        proc_call("sleep 5")
+
+        status_code = None
+        while status_code != 200:
+            try:
+                status_code = urllib.request.urlopen(
+                    "http://localhost:8732/chains/main/blocks/head"
```
rvem, line 476: Perhaps, it's worth getting the node RPC address from the service configuration instead of hardcoding it here. It's defined as `Environment="NODE_RPC_ADDR=127.0.0.1:8732"` in the `tezos-node-<network>` services

docker/package/tezos_baking_wizard.py
```diff
@@ -468,7 +468,16 @@ def bootstrap_node(self):
         self.import_snapshot()
 
         self.systemctl_start_action("node")
-        proc_call("sleep 5")
+
+        status_code = None
+        while status_code != 200:
```
pasqu4le, line 473: I'm not sure this condition is correct: in case a node is available (so no `URLError` is thrown), but the response is not `200` wouldn't this be stuck in an endless loop?

docker/package/tezos_baking_wizard.py
pasqu4le: One more nitpick about this: the RPC endpoint used is protocol-independent and doesn't rely on much (only on the existence of the `main` chain_id`), but why not use a simpler one?

How about [`/version`](http://tezos.gitlab.io/shell/rpc.html#get-version)?

docker/package/tezos_baking_wizard.py
ixahedron: Sure, but I would've thought it's intended, since we need an 'OK' response to proceed using the node. Do you think we should accept any response code as long as there is one and it's not a `URLError`?

pasqu4le: IMO yes, we should accept any successful connection here, regardless of the returned code.

An alternative might be to throw an error when the status code is not `200`, but I think we should limit this loop to only wait for the node RPC to be available for connection.
I think that whatever comes next should decide what to do with the content of the response (including the status code).

docker/package/tezos_baking_wizard.py
```diff
@@ -132,6 +132,33 @@ def list_connected_ledgers():
     return [name.decode() for name in re.findall(ledger_regex, output.stdout)]
 
 
+def get_node_rpc_addr(network):
+    try:
+        with open("/lib/systemd/system/tezos-node-" + network + ".service", "r") as f:
```
rvem, line 137: Let's get the path to the service file via `systemctl`, because the service can be modified manually via `systemctl edit` and the modified version is store in `/etc/systemd/system/`, in order to get path from the `systemctl`, you can use
`systemctl show -p FragmentPath tezos-node-<network>.service | cut -d'=' -f2`, or parse `systemctl show tezos-node-<network>.service` in code

ixahedron: `systemctl show` also readily shows `Environment` with all the needed variables :thinking: Is it not worth checking them straight from that output? Checking them from the file would be marginally easier, sure, but (to the best of my knowledge) `systemctl edit` could be used to override the `Environment` directly, right?

rvem: oh, indeed, it makes sense to parse `systemctl show` instead if `Environment` is presented in it. AFAIU, `systemctl show` will handle all overrides that happened to the service

rvem (approved):
LGTM now

pasqu4le (approved):
LGTM too

PR #224, opened 2021-06-22
[#222] Modify the wizard to query between setup

### LINKED ISSUE #219: [BUG] tezos-baking-wizard fails if snapshot url doesn't exist
# Description

tezos-baking-setup exits with an error if it can't download snapshot:

```
Trying to bootstrap tezos-node
Downloading the snapshot from  https://granadanet.xtz-shots.io/rolling
The error has been logged to /home/ubuntu/tezos_baking_wizard.log
$ cat /home/ubuntu/tezos_baking_wizard.log
<urlopen error [Errno -3] Temporary failure in name resolution>
```

Currently there are no snapshots for granadanet on xtz-shots.io (domain doesn't exist) since granadanet is very new.  
Similarly, the wizard exits with an error if user provides incorrect URL:

```
The Tezos node can take a significant time to bootstrap from scratch.
Bootstrapping from a snapshot is suggested instead.
How would you like to proceed?
1. download rolling (default)  Import rolling snapshot from xtz-shots.io (recommended)
2. download full               Import full snapshot from xtz-shots.io
3. file                        Import snapshot from a file
4. url                         Import snapshot from a url
5. skip                        Skip snapshot import and synchronize with the network from
                               scratch
> 4
Provide the url of the tezos-node snapshot file.
> http://google.com
Error in Tezos Baking Wizard, exiting.
The error has been logged to /home/ubuntu/tezos_baking_wizard.log
ubuntu@tez2:~$ cat /home/ubuntu/tezos_baking_wizard.log
'NoneType' object has no attribute 'validate'
```

the wizard should handle these errors, present user with a meaningful message (e.g. "xxx doesn't appear to point to a valid tezos snapshot: <error details>" and go back to the snapshot step.

### LINKED ISSUE #222: Modify the baking wizard to collect data between setup steps

Currently the baking wizard collects data from the user all at once, building a full configuration, before running the setup.

This is less than ideal mainly for two reasons:
1. unexpected errors always result in a failure of the whole setup, even tho some steps could be retried instead with new values
2. some data/info is requested unnecessarily as it is asked a priori and not conditionally

Examples:
- one is always asked what to do for bootstrapping the node. However if the node dir already contains data and I decide to keep it (maybe I bootstrapped it recently), the previous choice is ignored.
  The question (and consequently the bootstrapping) should not be asked in such case.
- one is always asked how to import a `baker` key, even when the latter is already present.
  If I decide to keep the existing one (maybe I imported it previously) there should be no such question.
- errors may occur for transient conditions or, like a broken internet connection, a choice made by mistake.
  If those can be repeated the choice should be given again, take #219 as an example on bootstrap source

# Acceptance criteria

The baking wizard should prompt the user in between the setup steps, so that:
- choices are given conditionally, only when the conditions make them necessary/useful
- choices should be given again when a permissible failure happened and the step could be repeated

ixahedron:
## Description

Problem: right now the Tezos Baking Wizard queries all configuration
options from the user all at once, before running the setup. This
has several undesirable ramifications, such as the inability to
re-query an option chosen by accident, or to address transient errors,
as well as asking some information unnecessarily.

Solution: updated the wizard to interleave queries and setup steps,
only asking the information when necessary and retrying failed options.

## Related issue(s)

Resolves #222, resolves #219 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_baking_wizard.py
+++ b/docker/package/tezos_baking_wizard.py
@@ -15,6 +15,48 @@ import re, textwrap
 import urllib.request
 from typing import List
 
+
+# Global options
+
+networks = {
+    "mainnet": "Main Tezos network",
+    "florencenet": "Test network using version 009 of Tezos protocol (Florence)",
+    "granadanet": "Test network using version 010 of Tezos protocol (Granada)",
+}
+
+modes = {
+    "baking": "Set up and start all services for baking: "
+    "tezos-node, tezos-baker, and tezos-endorser.",
+    "node": "Only bootstrap and run the Tezos node.",
+}
+
+snapshot_import_modes = {
+    "download rolling": "Import rolling snapshot from xtz-shots.io (recommended)",
+    "download full": "Import full snapshot from xtz-shots.io",
+    "file": "Import snapshot from a file",
+    "url": "Import snapshot from a url",
+    "skip": "Skip snapshot import and synchronize with the network from scratch",
+}
+
+key_import_modes = {
+    "ledger": "From a ledger",
+    "secret-key": "Either the unencrypted or password-encrypted secret key for your address",
+    "json": "Faucet JSON file from https://faucet.tzalpha.net/",
+}
+
+systemd_enable = {
+    "yes": "Enable the services, running them both now and on every boot",
+    "no": "Start the services this time only",
+}
+
+
+# Regexes
+
+ledger_regex = b"ledger:\/\/[\w\-]+\/[\w\-]+\/[\w']+\/[\w']+"
+secret_key_regex = b"(encrypted|unencrypted):(?:\w{54}|\w{88})"
+address_regex = b"tz1\w{33}"
+
+
 # Input validators
 
 
@@ -51,6 +93,17 @@ def required_field_validator(input):
     return input
 
 
+# The input has to be valid to at least one of the two passed validators.
+def or_validator(validator1, validator2):
+    def _validator(input):
+        try:
+            return validator1(input)
+        except:
+            return validator2(input)
+
+    return _validator
+
+
 # Runs the input through the passed validator, allowing for possible alteration,
 # but doesn't raise an exception if it doesn't validate to allow for custom options, too.
 def or_custom_validator(validator):
@@ -63,6 +116,32 @@ def or_custom_validator(validator):
         return result
 
     return _validator
+
+
+# To be validated, the input should adhere to the Tezos secret key format:
+# {encrypted, unencrypted}:<base58 encoded string with length 54 or 88>
+def secret_key_validator(input):
+    match = re.match(secret_key_regex.decode("utf-8"), input.strip())
+    if not bool(match):
+        raise ValueError(
+            "The input doesn't match the format for the Tezos secret key: "
+            "{{encrypted, unencrypted}:<base58 encoded string with length 54 or 88>}"
+            "\nPlease check the input and try again."
+        )
+    return input
+
+
+# To be validated, the input should adhere to the ledger URL format:
+# ledger://<four mnemonic words separated by dashes>/<curve>/[a-z0-9']+/[a-z0-9']+
+def ledger_url_validator(input):
+    match = re.match(ledger_regex.decode("utf-8"), input.strip())
+    if not bool(match):
+        raise ValueError(
+            "The input doesn't match the format for the ledger URL: "
+            "ledger://<four mnemonic words separated by dashes>/<curve>/[a-z0-9']+/[a-z0-9']+"
+            "\nPlease check the input and try again."
+        )
+    return input
 
 
 class Validator:
@@ -79,12 +158,6 @@ class Validator:
                 return self.validator(input)
 
 
-# Regexes
-
-ledger_regex = b"ledger:\/\/[\w\-]+\/[\w\-]+\/[\w']+\/[\w']+"
-secret_key_regex = b"(encrypted|unencrypted):\w+"
-address_regex = b"tz1\w{33}"
-
 # Wizard CLI utility
 
 
@@ -93,7 +166,7 @@ def proc_call(cmd):
 
 
 def fetch_snapshot(url):
-    print("Downloading the snapshot from ", url)
+    print("Downloading the snapshot from", url)
     filename = "/tmp/tezos_node.snapshot"
     urllib.request.urlretrieve(url, filename, progressbar_hook)
     print()
@@ -103,7 +176,7 @@ def fetch_snapshot(url):
 def progressbar_hook(chunk_number, chunk_size, total_size):
     done = chunk_number * chunk_size
     percent = min(int(done * 100 / total_size), 100)
-    print("Progress: ", percent, "%, ", int(done / (1024 * 1024)), "MB", end="\r")
+    print("Progress:", percent, "%,", int(done / (1024 * 1024)), "MB", end="\r")
 
 
 def color(input, colorcode):
@@ -117,8 +190,10 @@ def yes_or_no(prompt, default=None):
         if not answer and default is not None:
             answer = default
         if answer in ["y", "yes"]:
+            print()
             return True
         elif answer in ["n", "no"]:
+            print()
             return False
         else:
             print(color("Please provide a 'yes' or 'no' answer.", "\x1b[1;31m"))
@@ -171,7 +246,6 @@ class Step:
         default: str = "1",
         options=None,
         validator=None,
-        reqs=lambda _vs: True,
     ):
         self.id = id
         self.prompt = prompt
@@ -179,7 +253,6 @@ class Step:
         self.default = default
         self.options = options
         self.validator = validator
-        self.reqs = reqs
 
     def pprint_options(self):
         i = 1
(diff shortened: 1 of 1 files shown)


docker/package/tezos_baking_wizard.py
```diff
@@ -210,166 +245,121 @@ def pprint_options(self):
...
+    "By default, service with tezos-node will start to bootstrap from scratch,\n"
+    "which will take a significant amount of time.\nIn order to avoid this, we suggest "
+    "bootstrapping from a snapshot instead.\n\n"
+    "Snapshots can be downloaded from the following websites:\n"
+    "Tezos Giganode Snapshots - https://snapshots-tezos.giganode.io/ \n"
+    "XTZ-Shots - https://xtz-shots.io/ \n\n"
+    "We recommend to use rolling snapshots. This is the smallest and the fastest mode\n"
+    "that is sufficient for baking (you can read more about other tezos-node history modes here:\n"
+    "https://tezos.gitlab.io/user/history_modes.html#history-modes ).\n"
+    "All commands within the service are run under the 'tezos' user.",
```
rvem, line 295: Not sure if this last sentence is actually related to this step:)
Perhaps, it can be moved to the welcome message

rvem (approved):
LGTM, have only one nitpick

docker/package/tezos_baking_wizard.py
```diff
@@ -210,166 +245,121 @@ def pprint_options(self):
...
+    options=key_import_modes,
+    validator=Validator(enum_range_validator(key_import_modes)),
+)
+
+secret_key_query = Step(
+    id="secret_key",
+    prompt="Provide either the unencrypted or password-encrypted secret key for your address.",
+    help="The format is 'unencrypted:edsk...' for the unencrypted key, or 'encrypted:edesk...'"
+    "for the encrypted key.",
+    validator=Validator(required_field_validator),
```
rvem, line 332: Let's validate that the provided secret key matches the expected form: `{encrypted, unencrypted}:<base58 encoded string with length 54>`

docker/package/tezos_baking_wizard.py
```diff
@@ -210,166 +245,121 @@ def pprint_options(self):
...
+    validator=Validator([required_field_validator, filepath_validator]),
+)
+
+ledger_url_query = Step(
+    id="ledger_url",
+    prompt="Provide the ledger URL for importing the baker secret key stored on a ledger.",
+    help="Your ledger URL should look something like <ledger://<mnemonic>/ed25519/0'/1'>",
+    default=None,
+    options=list_connected_ledgers(),
+    validator=Validator(
```
rvem, line 349: Also, let's validate a custom URL, since we already have a regex for it

docker/package/tezos_baking_wizard.py
```diff
@@ -513,38 +560,79 @@ def bootstrap_node(self):
...
-            input("Please open the Tezos Baking app on your ledger, then hit Enter.")
-            self.check_baker_account(
-                "sudo -u tezos tezos-client import secret key baker "
-                + self.config["ledger_url"]
-            )
-            proc_call("sudo -u tezos tezos-client setup ledger to bake for baker")
+        replace_baker_key = self.check_baker_account()
+        valid_choice = False
+
+        while replace_baker_key and not valid_choice:
```
pasqu4le, line 587: Something seems off here, when there is no known baker account then `replace_baker_key` is `False` and so (regardless of `valid_choice`) this `while` loop never gets executed.

That, clearly, is a problem as what happens next is that `register_baker` is called and
`sudo -u tezos tezos-client register key baker as delegate` executed (which results in an error).

docker/package/tezos_baking_wizard.py
```diff
@@ -513,38 +560,79 @@ def bootstrap_node(self):
...
+        if address.returncode == 0:
+            value_regex = b"(?:" + ledger_regex + b")|(?:" + secret_key_regex + b")"
+            value = re.search(value_regex, address.stdout).group(0)
+            address = re.search(address_regex, address.stdout).group(0)
+            print()
+            print("An account with the 'baker' alias already exists.")
+            print("Its current value is", value.decode("utf-8"))
+            print("With a corresponding address:", address.decode("utf-8"))
+
+            return yes_or_no("Should it be replaced with the new import? <y/N> ", "no")
```
pasqu4le, line 579: I guess the message should be changed now to:
```suggestion
            return yes_or_no("Do you wish to import a new key and replace it? <y/N> ", "no")
```
as there is no "the new import" now (the question hasn't been asked yet)

docker/package/tezos_baking_wizard.py
```diff
@@ -513,38 +560,79 @@ def bootstrap_node(self):
...
+
+            if self.config["key_import_mode"] == "secret-key":
+                self.query_step(secret_key_query)
+                proc_call(
+                    "sudo -u tezos tezos-client import secret key baker "
+                    + self.config["secret_key"]
+                    + " --force"
+                )
+            elif self.config["key_import_mode"] == "json":
+                if self.config["network"] == "mainnet":
```
pasqu4le, line 599: given that we know this from the beginning, can't we instead avoid to even give the option in `key_mode_query`?

docker/package/tezos_baking_wizard.py
```diff
@@ -513,38 +560,79 @@ def bootstrap_node(self):
...
+                    print("Please choose one of the other options.")
+                    print()
+                    continue
+
+                self.query_step(json_filepath_query)
+                proc_call(
+                    "sudo -u tezos tezos-client activate account baker with "
+                    + self.config["json_filepath"]
+                    + " --force"
+                )
```
pasqu4le, line 613: I purposefully give the wizard a wrong path (to a file that exists) and this results in a failure of the whole wizard.

Can't we instead make sure that if `proc_call` like this fails we repeat the choice (I guess in this case the whole `import_baker_key`)?

More in general, I think this should be done for any call we do to `tezos-client` as there are several things that could go wrong and I think it's always better to re-give the choice instead of failing the whole process.

docker/package/tezos_baking_wizard.py
```diff
@@ -568,79 +656,108 @@ def systemctl_start_action(self, service):
...
+            "To access help and possible options for each question, type in 'help' or '?'."
+        )
+        print("Type in 'exit' to quit.")
+        print()
 
-    if cfg is not None:
-        setup = Setup(cfg)
+        self.query_step(network_query)
+        self.query_step(service_mode_query)
+        self.query_step(systemd_mode_query)
```
pasqu4le, line 682: why not move this to the very end?

After all we'll need to `start` the services either way and this will give a chance to the user to make some check (if he wants to) before actually enabling them.

docker/package/tezos_baking_wizard.py
ixahedron: I did what you suggested for importing the key, but what do you suggest to do about other `tezos-client` calls? E.g. one such call is made in `register_baker`, and if that fails after importing the key, that's (most likely) not the user's fault. What choice should be re-given then?

pasqu4le: mmmh good point, in general I would think that the best UX would be given by providing an option to either try again or go to the previous step (importing the baker key again).

However we need to think of each individual case IMO, in the `register_baker` example maybe we should just go back to the import step directly? 
Simply because that will recognize that there is already a `baker` and give the option to skip the import (so jumping to retying the registering)

docker/package/tezos_baking_wizard.py
```diff
@@ -576,57 +564,79 @@ def check_baker_account(self):
...
-        valid_choice = False
-
-        while replace_baker_key and not valid_choice:
 
-            self.query_step(key_mode_query)
+        if replace_baker_key:
+            if self.config["network"] == "mainnet":
+                key_import_modes.pop("json", None)
+
+            # We define the step here to disallow choosing json baking on mainnet
```
pasqu4le, line 603: I mean, I guess you could have made `key_mode_query` a function with an argument too?

Not sure if it's any better, just thinking aloud.

docker/package/tezos_baking_wizard.py
```diff
@@ -370,6 +370,21 @@ def pprint_options(self):
...
+    return Step(
+        id="key_import_mode",
+        prompt="How do you want to import the baker key?",
+        help="To register the baker, its secret key needs to be imported to the data "
+        "directory first.\nBy default tezos-baking-<network>.service will use the 'baker' "
+        "alias\nfor the key that will be used for baking and endorsing.\n"
+        "If you want to test baking with a faucet file, "
+        "make sure you have chosen a test network like " + list(networks.keys())[1],
+        options=modes,
+        validator=Validator(enum_range_validator(key_import_modes)),
```
pasqu4le, line 384:
```suggestion
        validator=Validator(enum_range_validator(modes)),
```

pasqu4le (approved):
Only a nitpick, but LGTM now

pasqu4le (changes requested):
Uh-oh, wait a second, I think I just spotted a problem

the `exit`/`quit` command doesn't seem to always work, I tried in the "importing baker key" step and it seems to print
"Exiting Tezos Baking Wizard" but then resume as if an exception from the tezos-client call was raised, I suppose due to how the exceptions are handled now.

pasqu4le (approved):
All resolved now

PR #225, opened 2021-06-23
[#218] Enhanced ledger interaction

### LINKED ISSUE #218: Enhance ledger interaction in the baking wizard

In #206 the first version of the baking wizard was implemented.

At the moment, it lacks a convenient way to display information about the keys from the connected
ledger device.

# Acceptance criteria

1) Display more information about the ledger URLs: show address and balance for each displayed URL.
2) Make it possible to specify custom URL for the key from the ledger

rvem:
## Description
Problem: Currently we show only ledger URLs for addresses that are store
on the ledgers, which is not very convenient when one has to choose the
desired key.

Solution: Also show the address that corresponds to the given URL and the
balance of this address.

Currently based on #224.

## Related issue(s)

Resolves #218

  can be successfully built.

### DIFF
--- a/docker/package/tezos_baking_wizard.py
+++ b/docker/package/tezos_baking_wizard.py
@@ -258,6 +258,31 @@ def get_data_dir(network):
         return "/var/lib/tezos/node-" + network
 
 
+def ledger_urls_info(ledger_urls):
+    info = []
+    max_url_len = max(map(len, ledger_urls))
+    for ledger_url in ledger_urls:
+        output = subprocess.run(
+            shlex.split("sudo -u tezos tezos-client show ledger {}".format(ledger_url)),
+            capture_output=True,
+        ).stdout
+        addr = re.search(address_regex, output).group(0).decode()
+        balance = (
+            subprocess.run(
+                shlex.split("tezos-client get balance for {}".format(addr)),
+                capture_output=True,
+            )
+            .stdout.decode()
+            .strip()
+        )
+        info.append(
+            ("{:" + str(max_url_len + 1) + "} address: {}, balance: {}").format(
+                ledger_url + ",", addr, balance
+            )
+        )
+    return info
+
+
 class Step:
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


rvem:
How it looks:
image

docker/package/tezos_baking_wizard.py
```diff
@@ -345,7 +370,9 @@ def pprint_options(self):
     prompt="Provide the ledger URL for importing the baker secret key stored on a ledger.",
     help="Your ledger URL should look something like <ledger://<mnemonic>/ed25519/0'/1'>",
     default=None,
-    options=list_connected_ledgers(),
+    # We need a running node to gather all information about ledger-stored addresses balances
+    # so we're using 'None' here and update options later once the node is available
+    options=None,
```
ixahedron, line 375: Good thinking :not-bad: Also keep in mind that the validator a couple lines below also uses `list_connected_ledgers()`, so you might want to re-assign the validator in the same place as `options`, too, once the node is up and running. Alternatively, you can move the entire step declaration to just before running that query, it'll probably be a bit cleaner even if it's not nice to separate step declarations.

ixahedron (approved):
LGTM, just one suggestion

docker/package/tezos_baking_wizard.py
rvem: `list_connected_ledgers()` doesn't require tezos-node running, but probably it'll be better to run it only once to be sure that the list hasn't changed over time for some reason

rvem:
@ixahedron, I've rebased this PR on the latest master. Please, review once again before merging to be sure that I don't break anything during rebase

docker/package/tezos_baking_wizard.py
```diff
@@ -638,6 +646,27 @@ def import_baker_key(self):
                         )
 
                     else:
+                        connected_ledgers = list_connected_ledgers()
+                        # This step requires tezos-node to be running and bootstrapped in order to gather
+                        # the data about the ledger-stored addresses, so it's defined right before invoking
+                        # after the node was boostrapped.
+                        ledger_url_query = Step(
```
ixahedron, line 653: I don't have a strong opinion on this, but I had a similar step with dependent options in the last version (see `get_key_mode_query`), and I followed a suggestion by @pasqu4le to make it into a function with the options argument. This allowed it to be near other steps where it belongs and also improved readability at a glance I guess. Could you change this to be in a similar fashion?

LGTM to me otherwise.

PR #235, opened 2021-07-01
[#233] Get client specific constants from config

### LINKED ISSUE #233: Fetch 'tezos-client' data directory and node endpoint URL from the service information
At the moment, the CLI wizard uses the default data directory and endpoint URL. However, users may change them in
the baking service configuration, which will cause the wizard to work incorrectly.

We should fetch the data directory and endpoint URL from the service configuration and use them in the `tezos-client` calls.

# Acceptance criteria
`tezos-client` calls in the baking wizard use the data directory and endpoint URL from the baking service configuration.

rvem:
## Description
Problem: Currently wizard assumes that tezos-client uses default data
directory and node rpc endpoint. However, users may change the default
configuration, which will lead to error in the wizard.

Solution: Gather the following tezos-client specific service constants
from the baking service config: data-directory filepath, baker account
alias, and node rpc endpoint.

## Related issue(s)

Resolves #233

  can be successfully built.

### DIFF
--- a/docker/package/tezos_baking_wizard.py
+++ b/docker/package/tezos_baking_wizard.py
@@ -207,36 +207,6 @@ def list_connected_ledgers():
     return [name.decode() for name in re.findall(ledger_regex, output.stdout)]
 
 
-def get_node_rpc_addr(network):
-    output = subprocess.run(
-        shlex.split("systemctl show tezos-node-" + network + ".service"),
-        capture_output=True,
-    ).stdout
-    config = re.search(b"Environment=(.*)(?:$|\n)", output)
-    if config is None:
-        print(
-            "tezos-node-" + network + ".service configuration not found, "
-            "defaulting to 'http://localhost:8732'"
-        )
-        return "http://localhost:8732"
-    config = config.group(1)
-    address = re.search(b"NODE_RPC_ADDR=(.*?)(?: |$|\n)", config)
-    if address is not None:
-        address = address.group(1).decode("utf-8")
-    else:
-        print("NODE_RPC_ADDR is undefined, defaulting to 'localhost:8732'")
-        address = "localhost:8732"
-
-    mode = "https://"
-    if (
-        re.search(b"KEY_PATH=(?: |$|\n)", config) is not None
-        or re.search(b"CERT_PATH=(?: |$|\n)", config) is not None
-    ):
-        mode = "http://"
-
-    return mode + address
-
-
 def get_data_dir(network):
     output = subprocess.run(
         shlex.split("systemctl show tezos-node-" + network + ".service"),
@@ -258,8 +228,11 @@ def get_data_dir(network):
         return "/var/lib/tezos/node-" + network
 
 
-def ledger_urls_info(ledger_urls):
+def ledger_urls_info(ledger_urls, node_endpoint):
     info = []
+    # max accepts only non-empty lists
+    if len(ledger_urls) == 0:
+        return info
     max_url_len = max(map(len, ledger_urls))
     for ledger_url in ledger_urls:
         output = subprocess.run(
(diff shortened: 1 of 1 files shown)


docker/package/tezos_baking_wizard.py
```diff
@@ -283,6 +291,57 @@ def ledger_urls_info(ledger_urls):
...
 
+def search_baking_service_config(network, regex, default):
+    output = subprocess.run(
+        shlex.split("systemctl show tezos-baking-" + network + ".service"),
+        capture_output=True,
+    ).stdout
+    config_filepath = re.search(b"EnvironmentFiles=(.*) ", output)
+    if config_filepath is None:
+        print(
+            "EnvionmentFiles not found in tezos-baking-"
```
ixahedron, line 302:
```suggestion
            "EnvironmentFiles not found in tezos-baking-"
```

docker/package/tezos_baking_wizard.py
```diff
@@ -283,6 +291,57 @@ def ledger_urls_info(ledger_urls):
     return info
 
 
+def search_baking_service_config(network, regex, default):
+    output = subprocess.run(
+        shlex.split("systemctl show tezos-baking-" + network + ".service"),
+        capture_output=True,
+    ).stdout
```
ixahedron, line 298: Actually (related to my comment about getting two variables all over again), maybe there's no need to actually re-query `systemctl` at all? Get the stdout once, then just run regex queries on that. I haven't thought about this in depth, but seems like that should work :shrug:

ixahedron (commented):
A couple of comments/suggestions, but mostly looks good

docker/package/tezos_baking_wizard.py
```diff
@@ -283,6 +291,57 @@ def ledger_urls_info(ledger_urls):
...
+def get_client_data_directory(network):
+    return search_baking_service_config(
+        network, 'DATA_DIR="(.*)"', "/var/lib/.tezos-client"
+    )
+
+
+def get_client_node_rpc_endpoint(network):
+    return search_baking_service_config(
+        network, 'NODE_RPC_ENDPOINT="(.*)"', "http://localhost:8732"
+    )
```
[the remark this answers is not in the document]
rvem: oh, I missed that `get_node_rpc_addr` exists, thanks for pointing that out

docker/package/tezos_baking_wizard.py
rvem: You're right

ixahedron: Seems like you didn't do this :thinking: Did you decide it wasn't worth it?

docker/package/tezos_baking_wizard.py
```diff
@@ -323,25 +324,10 @@ def get_client_data_directory(network):
     )
 
 
-def get_client_node_rpc_endpoint(network):
-    return search_baking_service_config(
-        network, 'NODE_RPC_ENDPOINT="(.*)"', "http://localhost:8732"
-    )
```
ixahedron: But :why:? Granted, right now there's not much of a difference, because both ways would query `systemctl` once, `get_node_rpc_addr` for `tezos-node-<network>.service` and `get_client_node_rpc_endpoint` for `tezos-baking-<network>.service`. But if you do end up going through with saving the output from config file referenced in `EnvironmentFiles` (see my other remaining comment), then it's better to use the new way (and consequently remove the `get_node_rpc_addr` function completely) to 1) save an external process call 2) increase readability some

rvem: Oh, looks like I completely misunderstood your previous comments. Sorry, my brain doesn't work very well today :(

rvem: Please, check whether I got you correctly this time

ixahedron (approved):
Nice, LGTM

PR #247, opened 2021-07-07
[#242] Fix baking wizard failing on Python 3.6

### LINKED ISSUE #242: [BUG] tezos-baking-wizard fails on Ubuntu 18.04
# Description
tezos-baking-wizard exits with "Error in Tezos Baking setup, exiting." after Tezos network prompt.

```
$ cat ./tezos_baking_wizard.log
__init__() got an unexpected keyword argument 'capture_output'
```

# Environment

- OS: Ubuntu 18.04
- tezos-baking  2:9.3-0ubuntu1~bionic
- python 3.6.9

# Additional context

Looks like [capture_output parameter has been added in Python 3.7](https://docs.python.org/3/library/subprocess.html#subprocess.run), Ubuntu 18.04 has Python 3.6

ixahedron:
## Description

Problem: in the baking wizard we use the `subrocess` lib for calling
external processes like `systemctl`. The API changed slightly from
Python 3.6 to 3.7, we need to support Ubuntu 18.04 which by default
has the former, and right now the wizard consistently crashes on it.

Solution: added the differentiation for the culprit method
`subprocess.run` for whether or not to use the `capture_output` parameter
based on the Python version running the wizard.

## Related issue(s)

Resolves #242 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_baking_wizard.py
+++ b/docker/package/tezos_baking_wizard.py
@@ -165,6 +165,13 @@ def proc_call(cmd):
     return subprocess.check_call(shlex.split(cmd))
 
 
+def get_proc_output(cmd):
+    if sys.version_info.major == 3 and sys.version_info.minor < 7:
+        return subprocess.run(shlex.split(cmd), stdout=subprocess.PIPE)
+    else:
+        return subprocess.run(shlex.split(cmd), capture_output=True)
+
+
 def fetch_snapshot(url):
     print("Downloading the snapshot from", url)
     filename = "/tmp/tezos_node.snapshot"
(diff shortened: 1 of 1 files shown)


rvem:
Is there any substantial difference between `proc_call` and `get_proc_output` now?
If no, could you please combine them in a single function

ixahedron:
> Is there any substantial difference between `proc_call` and `get_proc_output` now?
> If no, could you please combine them in a single function?

I mean, yeah, there is. `proc_call` runs the command and checks that the return code is 0, throwing an error if it's not (so if the called process itself exited unsuccessfully). That's all it does.
`get_proc_output` on the other hand runs the command and returns the `CompletedProcess` object, allowing us to collect the stdout/stderr from that call and specifically check the return code.
Now, for sure `proc_call` could be emulated through `get_proc_output`, but I believe the drawback of readability suffering with constant if or try-except blocks would be much more than an extra sorta-similar function.

rvem:
Oh, there is `check_call` and `run`, indeed. Sorry for noise then

rvem (approved):
I assume that you've tested it on Ubuntu 18.04 :pled:
(if no, please do)

PR #250, opened 2021-07-07
[#244] Set high watermark explicitly

### LINKED ISSUE #244: [BUG] High Watermark is not updated when setting up Ledger for baking
# Description

# To Reproduce

- Run tezos-baking-wizard
- Choose to import baking key from ledger 
- Use Ledger device that was previously set up to bake on a different network
- Follow prompts
- Notice a message like this:

```
Setting up the ledger:
* Main chain ID: NetXdQprcVkpaWU -> NetXz969SFaFn8k
* Main chain High Watermark: 251762 -> 251762
* Test chain High Watermark: 71662 -> 71662
```
 
Observe that chain id is about to be updated, but HW is to stay the same. In case of moving Ledger device from baking on a chain with large number of blocks to lower number of blocks made so far, baker/endorser will ignore blocks and will not bake or endorse.

# Expected behavior

High Watermark is adjusted to the current chain level.

# Environment

- Ubuntu 20.04.2
- 2:9.3-0ubuntu1~focal

rvem:
## Description
Problem: 'tezos-client setup ledger to baker' doesn't update high
watermark by default and thus bakes and endorsements can be missed in
case we switch from a network with higher number of blocks to the network
with lower number of blocks.

Solution: Provide current level as high watermark explicitly.

## Related issue(s)

Resolves #244 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_baking_wizard.py
+++ b/docker/package/tezos_baking_wizard.py
@@ -13,6 +13,7 @@ import os, sys, subprocess, shlex
 import readline
 import re, textwrap
 import urllib.request
+import json
 from typing import List
 
 
@@ -497,6 +498,12 @@ class Setup:
             + self.config["node_rpc_addr"]
         )
 
+    def get_current_head_level(self):
+        response = urllib.request.urlopen(
+            self.config["node_rpc_addr"] + "/chains/main/blocks/head/header"
+        )
+        return str(json.load(response)["level"])
+
     # Check if there is already some blockchain data in the tezos-node data directory,
     # and ask the user if it can be overwritten.
     def check_blockchain_data(self):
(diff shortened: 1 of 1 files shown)


ixahedron (approved):
LGTM

docker/package/tezos_baking_wizard.py
```diff
@@ -497,6 +497,15 @@ def get_tezos_client_options(self):
...
 
+    def get_current_head_level(self):
+        output = get_proc_output(
+            "sudo -u tezos tezos-client "
+            + self.get_tezos_client_options()
+            + " rpc get /chains/main/blocks/head/header"
+        ).stdout
+
+        return re.search(b'"level": ([0-9]+),', output).group(1).decode()
+
```
itkach, line 508: why not Python to invoke rpc directly (urllib.request is already in use), or at least parse json response properly (not with regex)?

rvem: Indeed, nice suggestion

PR #252, opened 2021-07-08
[#201] Rename the CLI wizard

ixahedron:
## Description

Problem: the CLI wizard we provide can be used to set up a baking
instance or just the Tezos node, but we call it `tezos-baking-wizard`.
This name is a bit confusing.

Solution: renamed the wizard to `tezos-setup-wizard`.

## Related issue(s)

Resolves part of #201 

  can be successfully built.

### DIFF
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -502,7 +502,7 @@ class TezosBakingServicesPackage(AbstractPackage):
 
     def fetch_sources(self, out_dir):
         os.makedirs(out_dir)
-        shutil.copy(f"{os.path.dirname(__file__)}/tezos_baking_wizard.py", out_dir)
+        shutil.copy(f"{os.path.dirname(__file__)}/tezos_setup_wizard.py", out_dir)
 
     def gen_control_file(self, deps, ubuntu_version, out):
         run_deps = ", ".join(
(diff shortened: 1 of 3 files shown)


pasqu4le (approved):

docker/package/model.py
```diff
@@ -17,6 +17,8 @@
     Unit,
 )
 
+letter_version = "a"
```
pasqu4le, line 20: Can you add a comment (or perhaps a clearer name?) just so we clarify what this is for?

I also think we should mention this in our release template now, as we should reset it to `""` whenever we bump the native release version(s)

pasqu4le (commented):
One more tiny nitpick, the PR description says;
```
Resolves part of #
```
(without mentioning the issue number :slightly_smiling_face: )

docker/package/model.py
rvem: Perhaps, it also can be moved to a `TezosBakingServicesPackage` class definition

rvem (approved):
LGTM, have nothing to add:)

docker/package/model.py
```diff
@@ -446,6 +446,13 @@ def gen_license(self, out):
 
 
 class TezosBakingServicesPackage(AbstractPackage):
+
+    # Sometimes we need to update the tezos-baking package inbetween
```
rvem, line 450: :+1:

pasqu4le (approved):

rvem (approved):

PR #257, opened 2021-07-19
[#255] Suppress Tezos testnet warning

### LINKED ISSUE #255: Suppress "This is NOT the Tezos Mainnet" message in tezos-setup-wizard

All interactions with tezos-client while setting up node or baking on a testnet with tezos-setup-wizard produce  "This is NOT the Tezos Mainnet" message, often repeatedly. I think this is not a useful warning in this context and should be suppressed during setup wizard.

# Acceptance criteria

The warning does not appear.

image

ixahedron:
## Description

Problem: currently, if the Tezos setup wizard is used on a testnet,
the "This is NOT the Tezos Mainnet." warning is shown multiple
times, which disrupts the flow of the setup.

Solution: suppressed the warning by temporarily passing the
approppriate environment variable with every tezos-client call.

## Related issue(s)

Resolves #255 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -162,15 +162,22 @@ class Validator:
 # Wizard CLI utility
 
 
+suppress_warning_env = dict(os.environ, TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER="YES")
+
+
 def proc_call(cmd):
-    return subprocess.check_call(shlex.split(cmd))
+    return subprocess.check_call(shlex.split(cmd), env=suppress_warning_env)
 
 
 def get_proc_output(cmd):
     if sys.version_info.major == 3 and sys.version_info.minor < 7:
-        return subprocess.run(shlex.split(cmd), stdout=subprocess.PIPE)
+        return subprocess.run(
+            shlex.split(cmd), stdout=subprocess.PIPE, env=suppress_warning_env
+        )
     else:
-        return subprocess.run(shlex.split(cmd), capture_output=True)
+        return subprocess.run(
+            shlex.split(cmd), capture_output=True, env=suppress_warning_env
+        )
 
 
 def fetch_snapshot(url):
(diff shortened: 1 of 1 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -208,7 +208,10 @@ def yes_or_no(prompt, default=None):
 
 
 def list_connected_ledgers():
-    output = get_proc_output("sudo -u tezos tezos-client list connected ledgers")
+    output = get_proc_output(
+        "sudo -u tezos TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=YES "
```
rvem, line 212: `TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=YES` copypaste makes me sad:(
AFAIR, `subprocess.run` has `env` argument, so you should be able to modify the current `os.environ` with `TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER` and pass updated env dict to the `run`

ixahedron: I know, it makes me sad too. The `env` argument is what I tried first, but the thing is that (for some reason, idk, the way `sudo` handles environments?) it doesn't work with `sudo -u tezos`.

rvem: :cry:

rvem: What do you think about moving `sudo -u tezos TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=YES` to the `get_proc_output`?

rvem: oh, btw, there is `--preserve-env` option for `sudo`

ixahedron: Oh, `--preserve-env` is pretty cool, thanks! On its own it would still require copy-paste though, just less of it.

Moving `get_proc_call` (and combining with `sudo -E`) is the option with the least changes, so I guess we'll do that. I hope it's OK for readability that that function isn't only used for calling `tezos-client`, but also e.g. for `systemctl show tezos-node-<network>.service`.

ixahedron: I lied. There's going to be some amount of copy-paste either way, either additions or deletions, and introducing _two_ dedicated functions (the equivalents of `proc_call` for checking the return code and of `get_proc_output` for getting the output) seems like an overkill. So I just added `-E` flags to `sudo` where necessary.
Please re-review.

rvem (approved):

PR #258, opened 2021-07-21
[#254] Wait for Baking ledger app

### LINKED ISSUE #254: [BUG] Setup wizard gets stuck if baking app is not started before choosing "1. ledger" for key import
# Description
tezos-setup-wizard gets stuck on ledger URI prompt without options to choose from and without a way to back out if user chose ledger option before starting baking app

# To Reproduce
- Make sure ledger device is not plugged in
- Follow tezos-setup-wizard prompts
- For "How do you want to import the baker key?" choose "1. ledger"
- Observe tezos-client output indicating no ledger is connected
- Observer prompt expects URI or choice  of a  option (but none exist)

# Expected behavior
- User will be instructed to connect Ledger device and start Tezos Baking App
- If ledger device can't be detected setup wizard will say so and will return to "How do you want to import the baker key?" prompt

# Environment

- Ubuntu 18.04
- 2:9.4a-0ubuntu1~bionic

rvem:
## Description
Problem: Currently the connected ledger is needed to get the list of the
ledger-stored keys. However, we ask the user to open the baking app after
calling 'tezos-client list connected ledgers'. As a result, the user can
be stuck during setup by being unable to choose the key because the
provided list of ledger-stored keys will be empty.

Solution: Ask the user to open the Baking app and wait until it's opened before
running `tezos-client list connected ledgers`.

## Related issue(s)

Resolves #254

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -214,9 +214,14 @@ def yes_or_no(prompt, default=None):
             print(color("Please provide a 'yes' or 'no' answer.", "\x1b[1;31m"))
 
 
-def list_connected_ledgers():
-    output = get_proc_output("sudo -u tezos -E tezos-client list connected ledgers")
-    return [name.decode() for name in re.findall(ledger_regex, output.stdout)]
+def wait_for_ledger_baking_app():
+    output = b""
+    while re.search(b"Found a Tezos Baking", output) is None:
+        output = get_proc_output(
+            "sudo -u tezos -E tezos-client list connected ledgers"
+        ).stdout
+        proc_call("sleep 1")
+    return [name.decode() for name in re.findall(ledger_regex, output)]
 
 
 def get_data_dir(network):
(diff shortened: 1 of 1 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -214,9 +214,14 @@ def yes_or_no(prompt, default=None):
...
-    output = get_proc_output("sudo -u tezos -E tezos-client list connected ledgers")
-    return [name.decode() for name in re.findall(ledger_regex, output.stdout)]
+def wait_for_ledger_baking_app():
+    output = b""
+    while re.search(b"Found a Tezos Baking", output) is None:
+        output = get_proc_output(
+            "sudo -u tezos -E tezos-client list connected ledgers"
+        ).stdout
+        proc_call("sleep 1")
+    return [name.decode() for name in re.findall(ledger_regex, output)]
```
ixahedron, line 224: Is there a reason why you opted for the `while` loop and not just moved the `input` prompt? Do you think this way will be more resilient? I'm asking because this calls external processes (`tezos-client` and `sleep 1`) potentially a lot and potentially unnecessary, and that feels a bit iffy.

But if you think this is better, then otherwise looks good.

ixahedron (commented):
Just one question

docker/package/tezos_setup_wizard.py
rvem: The main reason was to force the user to open the Baking app before proceeding the setup just to be sure that we'll be able to fetch ledger-stored keys

rvem: I was following the Kiln behaviour, which tells to open the Baking app on the ledger and then waits till its open

ixahedron (approved):

PR #260, opened 2021-07-22
[#256] Improve ledger import prompt

### LINKED ISSUE #256: Improve ledger prompt

Currently ledger prompt shows four choices - URIs for default derivation path along with corresponding addresses and balances as well as suggests free form entry of another ledger URI.

image

This makes it possible to enter a ledger uri that doesn't match connected ledger, only to be followed by an error.

image

A clearer interaction would be to prompt just for derivation path. An explicit option to go back would also be helpful (maybe user decided to use a different key import, not
ledger). So ledger prompt could look something like this:

```
Select a key to import from ledger://graceful-peacock-everlasting-kakapo:
1. bip25519/0h/0h,  address: tz1VfBqDgtGJ9BrTqBgT3hXQqfuWRpPCCrJy, balance: 6128.499021 ꜩ
2. ed25519/0h/0h,   address: tz1NGS7FEGGfEFp5XRVkHBqVqSQ8VuumF9j3, balance: 0 ꜩ
3. secp256k1/0h/0h, address: tz2RBShUqimw1pbJ5mXZurRPUh2HNP8H9RVp, balance: 0 ꜩ
4. P-256/0h/0h,     address: tz3jWMH6Ao5j7FUobxbrA5qnZvWCUQ4pQvn7, balance: 0 ꜩ
5. Specify different derivation path
6. Go back
> 

```

Option `5` would prompt for a new derivation path, e.g. `1h/1h` and then go back to this prompt with new options reflecting the new selected path:

```
Select a key to import from ledger://graceful-peacock-everlasting-kakapo:
1. bip25519/1h/1h,  address: tz1AAA, balance: 6128.499021 ꜩ
2. ed25519/1h/1h,   address: tz1BBB, balance: 0 ꜩ
3. secp256k1/1h/1h, address: tz2CCC, balance: 0 ꜩ
4. P-256/1h/1h,     address: tz3DDD, balance: 0 ꜩ
5. Specify different derivation path
6. Go back
> 

```

# Acceptance criteria

User can provide alternative derivation path to see  different set of addresses. User can go back to previous prompt. User can't and doesn't need to specify full ledger URI.

rvem:
## Description
Problem: Currently it's possible to choose one of the four ledger keys
provided by tezos-client by default or type a ledger url manually.
However, we don't properly validate that ledger url is valid.
Instead, we should provide a way to choose a derivation path that will be
used with the current ledger device.

Solution: Make it possible to provide custom derivation path instead of
the whole ledger url. Once the derivation is provided it appears in the
list of possible ledger urls to choose from. As an addition, an option
to go back to the baker key import process beginning is added.

New prompt:
image

## Related issue(s)

Resolves #256

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -56,7 +56,7 @@ systemd_enable = {
 ledger_regex = b"ledger:\/\/[\w\-]+\/[\w\-]+\/[\w']+\/[\w']+"
 secret_key_regex = b"(encrypted|unencrypted):(?:\w{54}|\w{88})"
 address_regex = b"tz[123]\w{33}"
-
+derivation_path_regex = b"(?:bip25519|ed25519|secp256k1|P-256)\/[0-9]+h\/[0-9]+h"
 
 # Input validators
 
@@ -132,14 +132,14 @@ def secret_key_validator(input):
     return input
 
 
-# To be validated, the input should adhere to the ledger URL format:
-# ledger://<four mnemonic words separated by dashes>/<curve>/[a-z0-9']+/[a-z0-9']+
-def ledger_url_validator(input):
-    match = re.match(ledger_regex.decode("utf-8"), input.strip())
+# To be validated, the input should adhere to the derivation path format:
+# [0-9]+h/[0-9]+h
+def derivation_path_validator(input):
+    match = re.match(b"[0-9]+h\/[0-9]+h".decode("utf-8"), input.strip())
     if not bool(match):
         raise ValueError(
-            "The input doesn't match the format for the ledger URL: "
-            "ledger://<four mnemonic words separated by dashes>/<curve>/[a-z0-9']+/[a-z0-9']+"
+            "The input doesn't match the format for the derivation path: "
+            "'[0-9]+h/[0-9]+h'"
             "\nPlease check the input and try again."
         )
     return input
@@ -221,7 +221,12 @@ def wait_for_ledger_baking_app():
             "sudo -u tezos -E tezos-client list connected ledgers"
         ).stdout
         proc_call("sleep 1")
-    return [name.decode() for name in re.findall(ledger_regex, output)]
+    base_ledger_url = (
+        re.search(b"ledger:\/\/[\w\-]+\/", output).group(0).decode("utf-8")
+    )
+    return base_ledger_url, [
+        name.decode() for name in re.findall(derivation_path_regex, output)
+    ]
 
 
 def get_data_dir(network):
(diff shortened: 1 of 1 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -422,25 +426,32 @@ def get_key_mode_query(modes):
...
+        options=ledger_urls_info(connected_ledgers, base_ledger_url, node_endpoint)
+        + extra_options,
         validator=Validator(
             [
                 required_field_validator,
-                or_validator(
-                    enum_range_validator(connected_ledgers),
-                    ledger_url_validator,
-                ),
+                enum_range_validator(connected_ledgers + extra_options),
```
ixahedron, line 454: I think there might be a problem here with validation: `connected_ledgers` is built up using `ledger_regex`, so it's still just full ledger URLs. `ledger_urls_info` will truncate the common base URL and only print derivation paths. However, the validator here still receives full URLs via `connected_ledgers`, so if the user types in an (already present) derivation instead of, say, `3`, it will refuse to validate.

rvem: Well, the point is to allow only numeric options here, so that a user choose `Specify derivation path` and inputs derivation in the next prompt

docker/package/tezos_setup_wizard.py
```diff
@@ -56,7 +56,7 @@
 ledger_regex = b"ledger:\/\/[\w\-]+\/[\w\-]+\/[\w']+\/[\w']+"
 secret_key_regex = b"(encrypted|unencrypted):(?:\w{54}|\w{88})"
 address_regex = b"tz[123]\w{33}"
-
+derivation_path_regex = b"(bip25519|ed25519|secp256k1|P-256)\/[0-9]+h\/[0-9]+h"
```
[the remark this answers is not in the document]
rvem: This format matches ledger URLs used in `tezos-client`

docker/package/tezos_setup_wizard.py
ixahedron: Yeah, I know, I'm talking about `enum_range_validator` accepting both numeric options and their written out counterparts. E.g. in the `How do you want to import the baker key` step it's possible to answer both `1` and `ledger`, but not `42` or `idk whatever`. Here, the prompt is for derivation paths, so from the expected behaviour it should be possible to choose `ed25519/0h/0h` by either typing it out or entering `2`. But the validator here accepts full ledger URLs, so `ed25519/0h/0h` would not be accepted, only `ledger://live-fossa-eager-walrus/ed25519/0h/0h`.

rvem: Oh, okay, then something smarter is needed indeed

itkach: As I mention in my original issue description, I don't think user should need to type algo name, just derivation path e.g. 0h/1h, (which btw can be empty and produces yet another set of addresses) or even accept just a pair of numbers types like 0/1 (not sure if those `h` and `'` sometimes used for derivation path need to be typed)

rvem: I'm not sure if it's possible to provide an URL without the algo name to the `tezos-client` :thinking:

itkach: to quote myself:

rvem: Ah, okay, that makes sense

docker/package/tezos_setup_wizard.py
```diff
@@ -422,25 +429,32 @@ def get_key_mode_query(modes):
     validator=Validator([required_field_validator, filepath_validator]),
 )
 
+derivation_path_query = Step(
+    id="derivation_path",
+    prompt="Provide derivation path for the key stored on the ledger.",
+    help="The format is '([0-9]+h/[0-9]+h'",
```
ixahedron, line 435:
```suggestion
    help="The format is '[0-9]+h/[0-9]+h'",
```

ixahedron (approved):
LGTM except a typo

pasqu4le (approved):

PR #261, opened 2021-07-23
v10.0-rc2 release

ixahedron:
## Description

This PR introduces a new v10.0-rc2 release in tezos-packaging.

Follows #

### Changes related to the creation of a release (conditional)

- [x] I updated Tezos sources and opam-repository (if needed) revisions in [sources.json](/serokell/tezos-packaging/tree/master/nix/nix/sources.json).
- [x] I removed old bottles hashes from the brew formulas in [Formula directory](/serokell/tezos-packaging/tree/master/Formula).
- [x] I updated `url :tag` and `version`s in brew formulas.
- [x] I updated release number in [meta.json](/serokell/tezos-packaging/tree/master/meta.json).
- [x] If the native release version was updated, I reset the `letter_version` in [model.py](/serokell/tezos-packaging/tree/master/docker/package/model.py).
- [x] I removed native-packaging-related steps from CI in case opam-repository wasn't updated yet.

#### In case the new Tezos release provides a new protocol and corresponding testnet

- [x] I supported new protocol in [protocols.json](/serokell/tezos-packaging/tree/master/protocols.json).
- [x] I supported new protocol and testnet in native packaging.
- [x] I supported new protocol and testnet in brew formulas.
- [x] I added tests for the new protocol and testnet.

#### Create a new release in this repository

- [ ] I created a new release that is based on the latest autorelease created by the CI.
- [x] I enabled native-packaging-related CI steps after opam-repository got updated.

#### Update brew bottles and repository mirrors

- [ ] I compiled brew bottles for all required macOS versions using [`build-bottles.sh`](/serokell/tezos-packaging/tree/scripts/build-bottles.sh)
      script and uploaded them to the created release.
      Note that for this you'll need a macOS machine running each required version.
- [ ] I added new bottles sha256 hashes to the brew formulas.
- [ ] I pushed changes to either [tezos-packaging-rc](https://github.com/serokell/tezos-packaging-rc) or
      [tezos-packaging-stable](https://github.com/serokell/tezos-packaging-stable) mirror repositories.

#### Publish new native packages

Once [opam-repository](https://opam.ocaml.org/packages/) is updated with the new version of Tezos packages.

- [x] I published new Ubuntu packages, see [these instructions](/serokell/tezos-packaging/tree/master/docker#source-packages-and-publishing-them-on-launchpad-ppa).
- [x] I published new Fedora packages, see [these instructions](/serokell/tezos-packaging/tree/master/docker#srcrpm-packages).

#### Update documentation

[...]

PR #262, opened 2021-07-26
Add bottle hashes for v10.0-rc2

ixahedron:
## Description

Problem: v10.0-rc2 bottles have been built and added to the latest release.

Solution: added hashes for Mojave and Catalina to brew formulae.

## Related issue(s)

Follows #261 

  can be successfully built.


pasqu4le (approved):
One minor nitpick: the "problem" in the commit message isn't really a problem :sweat_smile: 

I think it should mention that there are uploaded bottles, but this solves the fact that their hashes are not in the formulae.

PR #265, opened 2021-07-28
Granada mainnet activation

ixahedron:
## Description

Problem: granada will be activated on mainnet very soon. We need to support an automatic switchover.

Solution: added `010-PtGRANAD` to the list of mainnet protocols.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -451,7 +451,7 @@ class TezosBakingServicesPackage(AbstractPackage):
     # native releases, so we append an extra letter to the version of
     # the package.
     # This should be reset to "" whenever the native version is bumped.
-    letter_version = "a"
+    letter_version = "b"
 
     def __init__(
         self,
--- a/meta.json
+++ b/meta.json
@@ -1,4 +1,4 @@
 {
-    "release": "1",
+    "release": "2",
     "maintainer": "Serokell <hi@serokell.io>"
 }
--- a/docker/package/packages.py
+++ b/docker/package/packages.py
@@ -15,7 +15,7 @@ from .systemd import Service, ServiceFile, SystemdUnit, Unit, Install
 
 networks = ["mainnet", "florencenet", "granadanet"]
 networks_protos = {
-    "mainnet": ["009-PsFLoren"],
+    "mainnet": ["009-PsFLoren", "010-PtGRANAD"],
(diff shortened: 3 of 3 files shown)


docker/package/model.py
```diff
@@ -451,7 +451,7 @@ class TezosBakingServicesPackage(AbstractPackage):
     # native releases, so we append an extra letter to the version of
     # the package.
     # This should be reset to "" whenever the native version is bumped.
-    letter_version = "a"
+    letter_version = ""
```
pasqu4le, line 454: Am I wrong or this could be updated to `b` instead of the `release` be bumped to `2`?

ixahedron: What for is the release number at all then? AFAIU, `letter_version` is specifically for updating `tezos_baking` (i.e. package's source e.g. `tezos-setup-wizard`), whereas this will really affect every package, so we should bump release number instead.

However, it's not really documented anywhere :shrug: , so I might be wrong.

rvem: `letter_version` is required only for `tezos_baking` because we tend to updated sources of this package (i.e. `tezos_setup_wizard.py`), while the other _normal_ packages rely on the sources gathered via `opam-bundle` and they shouldn't change in scope of one upstream version.

I see that the latest published version of tezos-baking is `2:9.4a-0ubuntu1`. However, AFAIU, we've changed the sources of the `tezos_setup_wizard.py` since that, so `letter_version` has to be updated for the newly published version of `tezos-baking` package

meta.json
```diff
@@ -1,4 +1,4 @@
 {
-    "release": "1",
+    "release": "2",
```
rvem, line 2: These changes indeed affect only the metadata of `tezos-baking` package (since it only changes the `.services` files). However, we did change the sources of `tezos-baking` package as well (by modifying `tezos_setup_wizard.py`), so I think we should bump `letter_version` and leave the `release` as it is

docker/package/model.py
rvem: > This should be reset to "" whenever the native version is bumped.

Note that upstream version wasn't bumped, so it should be `letter_version = "b"` instead

rvem (approved):

PR #275, opened 2021-08-10
Deactivate protocol 009

ixahedron:
## Description

Problem: florencenet is now obsolete, so we don't need to support it.

Solution: removed 009 support from tezos-packaging.
Also enabled back the native packaging step in CI.

## Related issue(s)

Follows #271 

  can be successfully built.


kirelagin (approved):

kirelagin:
Note that CI fails on `tezos-lmdb` hashes, which sounds pretty bad 🙄. But, I guess, maybe it’s just the way things work in OCaml…

PR #277, opened 2021-08-11
Use Github Actions for checking and building homebrew bottles

zhenyavinogradov:
Replaces buildkite step for checking homebrew formulas with a Github Actions workflow, and adds another workflow to build and export bottles using catalina runners provided by github. Github does not provide mojave runners, so for mojave we'll still have to build bottles ourselves. And the bottles are only built on master branch because it takes a long time, but here's an example of the workflow running: https://github.com/serokell/tezos-packaging/runs/3260140954.

The bottles are exported as a workflow artifact and can be downloaded from the workflow page. It would be nice to attach them to the pre-release automatically, but I don't know if there is a simple way to do it, because releases are created from buildkite, and buildkite attaches some of its artifacts as well, so we can't really synchronize it with github actions.

Also, brew has a special `brew test-bot` command which we could use, it is intended for setting up a clean environment in CI and running all checks automatically, but I tried using it and it's not really convenient, because it does not give a clear way to configure what you want it to do. It's simpler to just run `brew install` in CI.

Related issue: https://issues.serokell.io/issue/TM-563.

### DIFF
--- a/.github/workflows/build-bottles.yml
+++ b/.github/workflows/build-bottles.yml
@@ -1,0 +1,22 @@
+name: Build homebrew bottles
+on:
+  # Run when a release is tagged
+  push:
+    tags:
+      - "v*"
+jobs:
+  build-bottles:
+    runs-on: macos-10.15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Build bottles
+        run: ./scripts/build-bottles.sh
+
+      - name: Upload bottles to Github Actions
+        uses: actions/upload-artifact@v2
+        with:
+          name: homebrew-bottles
+          path: '*.bottle.*'
+
--- a/.buildkite/pipeline-raw.yml
+++ b/.buildkite/pipeline-raw.yml
@@ -70,18 +70,6 @@ steps:
      queue: "arm64-build"
    only_changes: *static_binaries_changes_regexes
 
- - label: check brew formulas
-   commands:
-     # Check all formulas except 'tezos.rb' because it's a base class for all other formulas
-     - find ./Formula -type f -name "*.rb" -exec ruby -c {} +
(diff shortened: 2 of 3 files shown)


.github/workflows/build-bottles.yml
```diff
@@ -0,0 +1,25 @@
+name: Build homebrew bottles
+on:
+  # Run on master
+  push:
+    branches: [master]
```
ixahedron, line 5: From the Github Actions docs it seems it's possible to just tie this action to `release` directly? Even though it's conditional on relevant paths, we sometimes go a long way committing to master without making a release (and, consequently, without need to build bottles), and also we touch the formulae _after_ each release too, to put in the hashes. Duplicating the runs seems unnecessary.

I'm not sure which `release` activity type would be needed here, but we need the bottles to build on both full releases and prereleases _created from_ the auto-release, but not on _each_ auto-release. If that's not possible, then I guess these paths are the next best alternative.

zhenyavinogradov: Github has draft releases which are only visible to repository maintainers, we can change `autorelease.sh` script to create a draft release and set up the action to run when the release is published (either as a pre-release or as a full release), so that it will only run for releases which you publish manually. Maybe we can even attach the bottles as assets automatically in this case. The only issue is that you'll have to wait 4 hours for the bottles to build after you make a release.

Another options are probably to run the workflow on tags of the form `v*`, or to trigger the workflow manually when you need it.

ixahedron: Running on `v*` tags sounds good to me for now.

zhenyavinogradov: Ok, I updated it to run on tags

ixahedron (approved):
LGTM

Issue #291, opened 2021-08-26
Handle v10.0 storage upgrade in tezos-node services


rvem:
# Clarification and motivation
It's required to upgrade the `tezos-node` storage in case the binary was upgraded from `v9.*` to `v10.0`.
In case the upgrade wasn't done, `tezos-node run` gracefully fails.

It's not very convenient to do the upgrade manually for the node running in the background service,
so this process should be semi-automated.
There are two ways to handle the storage upgrade:
1) Upgrade existing storage.
2) Nuke the existing storage and import a fresh snapshot.

Node services should ask a user about the preferable way of the upgrade interactively in case
the `tezos-node` start fails with an error that requires a storage upgrade.

# Acceptance criteria
Node systemd services can handle storage upgrade by asking the user about the preferable way of upgrading.

rvem:
v10.1 tezos-node native packages include this fix

PR #293, opened 2021-08-27
[#291] Automatic update node storage

rvem:
## Description
Problem: 9.* to 10.* upgrade requires a storage upgrade.
We want to automate this upgrade process.

Solution: Check the current storage version in the node service pre-start
script and perform an upgrade in case it's needed.

## Related issue(s)

Relates #291 (will be resolved, once corresponding updated package is published)

  can be successfully built.

### DIFF
--- a/docker/package/packages.py
+++ b/docker/package/packages.py
@@ -168,7 +168,7 @@ def mk_node_unit(suffix, env, desc):
             environment=env,
             exec_start="/usr/bin/tezos-node-start",
             exec_start_pre=["/usr/bin/tezos-node-prestart"],
-            timeout_start_sec="450s",
+            timeout_start_sec="2400s",
             state_directory="tezos",
             user="tezos",
         ),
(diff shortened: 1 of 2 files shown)


rvem:
Looks like there are more cases in which the upgrade is going to fail, e.g. some issues with the write permissions on the files in the node data directory (see https://gitlab.com/tezos/tezos/-/issues/1687).

So I agree, that we should omit our own checks and let `tezos-node upgrade` fail (well, currently, it won't return non-zero exit code for some reason, but still `tezos-node run` is going to fail)

rvem:
Sadly, `tezos-node` doesn't report when there is not enough space to perform a backup before the storage upgrade:(
It's quite nontrivial to calculate the _proper_ amount of space required for the upgrade, so it's quite nontrivial to support
such a check on our side

pasqu4le (approved):

PR #295, opened 2021-08-31
[#284] Add script automating bottle hash insertion

### LINKED ISSUE #284: Automate inserting bottle hashes into brew formulae

Right now, after each release we build brew bottles, and then manually compute hashes and insert each one into the appropriate `Formula/*.rb`.

This is slow, tedious, and error-prone. It would be nice to automate it.

A bash script would be a good place to start.

# Acceptance criteria

A script is provided to insert bottle hashes into brew formular automatically instead of going through the motions manually.

ixahedron:
## Description

Problem: every time after a new release, we build brew bottles,
and their SHA256 hashes should be added to the brew formulae.
Right now we do this manually, but it's a tedious and error-prone
process that would be nice to automate.

Solution: added a bash script that can do this. It takes a directory
where the bottle.tar.gz files are placed as its argument and inserts
the computed hash for the respective OS into the right formula.
It is supposed to be run from the base directory (tezos-packaging).

## Related issue(s)

Resolves #284 

  can be successfully built.

### DIFF
--- a/scripts/bottle-hashes.sh
+++ b/scripts/bottle-hashes.sh
@@ -1,0 +1,31 @@
+#! /usr/bin/env bash
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+# This script takes a directory where the bottles are stored as its argument.
+# Run it from the base directory (tezos-packaging).
+
+if [[ -d ./Formula ]]
+then
+    if [[ -d "$1" ]]
+    then
+        regex="(tezos-.*)-v.*\.(catalina|mojave)\.bottle\.tar\.gz"
+        for bottle in "$1"/tezos-*.bottle.tar.gz; do
+            if [[ $bottle =~ $regex ]]; then
+                bottle_hash=`sha256sum "$bottle" | cut -d " " -f 1`
+                formula_name="${BASH_REMATCH[1]}"
+                os="${BASH_REMATCH[2]}"
+
+                if [[ -f "./Formula/$formula_name.rb" ]]; then
+                    line="\    sha256 cellar: :any, $os: \"$bottle_hash\""
+                    sed -i "/root_url.*/a $line" ./Formula/$formula_name.rb
+                fi
+            fi
+        done
+    else
+        echo "The passed directory $1 does not exist."
+    fi
+else
+    echo "Please run this script from the base directory (tezos-packaging)."
+fi
--- a/tests/bottle-hashes/test-hash-insert.sh
+++ b/tests/bottle-hashes/test-hash-insert.sh
@@ -1,0 +1,82 @@
+#! /usr/bin/env bash
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+# This script tests ../../scripts/bottle-hashes.sh by generating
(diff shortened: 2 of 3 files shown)


scripts/bottle-hashes.sh
```diff
@@ -0,0 +1,31 @@
...
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+# This script takes a directory where the bottles are stored as its argument.
+# Run it from the base directory (tezos-packaging).
+
+if [[ -d ./Formula ]]
+then
+    if [[ -d "$1" ]]
+    then
+        regex="(tezos-.*)-v[0-9]+\.[0-9]+-[0-9]+\.(catalina|mojave)\.bottle\.tar\.gz"
```
rvem, line 13: will `rc`s match rhis regex? :thinking:

rvem:
The script looks relatively complex, let's add some tests to it that will generate some dummy bottles (e.g. using `dd if=/dev/urandom of=<bottle-arcvhive> bs=1M`) and assert that information about them was successfully inserted to the formulas

scripts/bottle-hashes.sh
pasqu4le: to add to the comment above: we know for sure (according to the current tezos naming scheme) that the `v` is also used at the beginning of every version name.

Consequently what's between the last `-v` before the `.$os`  and the latter iself is all part of the version, which isn't something we are actually interested in, for this script.

I'd suggest a more general expression to gloss over it, I think
```suggestion
        regex="(tezos-.*)-v.*\.(catalina|mojave)\.bottle\.tar\.gz"
```
should work.

tests/bottle-hashes/test-hash-insert.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# Generate some dummy bottles
+dd if=/dev/urandom of=$bottle_dir/$catalina_bottle count=2000 status=none
+dd if=/dev/urandom of=$bottle_dir/$mojave_bottle count=2000 status=none
+
+# Run the hash inserting script
+../../scripts/bottle-hashes.sh $bottle_dir
+
+# Assert the info was inserted correctly
+catalina_hash=`sha256sum $bottle_dir/$catalina_bottle | cut -d " " -f 1`
```
rvem, line 48:
```suggestion
catalina_hash="$(sha256sum "$bottle_dir/$catalina_bottle" | cut -d " " -f 1)"
```

tests/bottle-hashes/test-hash-insert.sh
```diff
@@ -0,0 +1,82 @@
...
+mojave_hash=`sha256sum $bottle_dir/$mojave_bottle | cut -d " " -f 1`
```
rvem, line 49:
```suggestion
mojave_hash="$(sha256sum "$bottle_dir/$mojave_bottle" | cut -d " " -f 1)"
```

tests/bottle-hashes/test-hash-insert.sh
```diff
@@ -0,0 +1,82 @@
...
+    sha256 cellar: :any, catalina: "$catalina_hash"
+  end
+
+  def install
+  end
+end
+EOF
+)
+
+if [[ $(< ./Formula/$formula_name.rb) != $expected_formula ]]; then
```
rvem, line 73:
```suggestion
if [[ $(< ./Formula/$formula_name.rb) != "$expected_formula" ]]; then
```

rvem (approved):
Some suggestions from the shellcheck

pasqu4le (approved):

PR #297, opened 2021-09-01
[#285] Build mojave bottles and attach bottles to releases in CI

### LINKED ISSUE #285: Automate post-release brew bottles generation and upload

After #277 we now have catalina bottles generated by CI (GH Actions in this case) after a `v*` tag is pushed.

It should be possible to do something similar for mojave bottle (maybe using buildkite, as GH does not support it).

Moreover, as part of our [releasing process](https://github.com/serokell/tezos-packaging/blob/master/.github/PULL_REQUEST_TEMPLATE/release_pull_request_template.md#update-brew-bottles-and-repository-mirrors) we tend to upload bottles after making a release (to avoid long waiting times) so we should also change to this trigger and make the upload part of this steps.

# Acceptance criteria

After a pre-release gets published as a release, CI will generate the brew bottles (for both mojiave and catalina) and push the resulting files in the release itself.

zhenyavinogradov:
Problem: we build homebrew bottles for catalina in CI, but not for mojave. And the bottles have to be attached to the releases manually.

Solution: add buildkite pipeline which builds mojave bottles when a tag is pushed, and add automatic attachment for both catalina and mojave.

In buildkite it's a separate pipeline `tezos-packaging-tags`, which runs when a tag is pushed rather than a branch, it takes jobs from `.buildkite/pipeline-for-tags.yml` file.

Resolves #285

### DIFF
--- a/.buildkite/pipeline-for-tags.yml
+++ b/.buildkite/pipeline-for-tags.yml
@@ -1,0 +1,23 @@
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+steps:
+ - label: Build bottles
+   key: build-bottles
+   agents:
+     queue: "x86_64-darwin"
+   if: build.tag =~ /^v.*/
+   commands:
+   - ./scripts/build-bottles.sh
+   artifact_paths:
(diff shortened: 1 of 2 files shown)


ixahedron (approved):
LGTM

rvem (approved):

pasqu4le (approved):

PR #303, opened 2021-09-21
[morley#664] Allow alpha protocol

rvem:
## Description
Problem: Alpha protocol is now required to run 'tezos-client' in mockup
mode.

Solution: Allow it and adjust hacks.nix.

## Related issue(s)

https://gitlab.com/morley-framework/morley/-/issues/664

  can be successfully built.

### DIFF
--- a/nix/build/hacks.nix
+++ b/nix/build/hacks.nix
@@ -229,6 +229,9 @@ rec {
     propagatedBuildInputs = buildInputs;
   });
 
+  tezos-protocol-alpha = osuper.tezos-protocol-alpha.overrideAttrs (o : {
+    buildInputs = o.buildInputs ++ [ tezos-protocol-environment ];
+  });
   tezos-protocol-000-Ps9mPmXa = osuper.tezos-protocol-000-Ps9mPmXa.overrideAttrs (o : {
     buildInputs = o.buildInputs ++ [ tezos-protocol-environment ];
   });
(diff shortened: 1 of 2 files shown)


nix/build/hacks.nix
```diff
@@ -332,7 +343,7 @@ rec {
 
   tezos-client = osuper.tezos-client.overrideAttrs
     (o: {
-      buildInputs = o.buildInputs ++ [ librustzcash self.makeWrapper ];
+      buildInputs = o.buildInputs ++ [ librustzcash self.makeWrapper tezos-client-alpha-commands-registration ];
```
ixahedron, line 346: Wait, where is `tezos-client-alpha-commands-registration` defined?

rvem: It's somewhere in the `ocamlPackages`, that are fulfilled by the `opam-nix` while traversing the context of the `opam-repository` :ssc:

ixahedron: So you don't need to access it through `osuper` or `self.ocamlPackages.<...>` or something? :thinking: Sure would be nice to understand more about how `opam-nix` works

rvem: there is `with oself` on the top of that file, and `self.ocamlPackages` is basically `oself`

ixahedron (approved):

Issue #288, opened 2021-08-25
Automate generation of new releases PRs


pasqu4le:
# Clarification and motivation

After #286 and #287 (I suggest starting working on this issue only after these two) we should have most on-release and post-release actions automated.

However, we'll still need to manually perform the [initial release PR](https://github.com/serokell/tezos-packaging/blob/master/.github/PULL_REQUEST_TEMPLATE/release_pull_request_template.md#changes-related-to-the-creation-of-a-release-conditional) part of our process.

At this point it should be possible to automate this too and have CI open a PR (to let us review and, if necessary, fix) any time there is a new release on [tezos/tezos](https://gitlab.com/tezos/tezos).

Ideally, at the end of this issue, it should be possible to remove the release-specific PR template and move any useful information from there elsewhere.

# Acceptance criteria

CI is capable of opening PRs with all the necessary changes that are needed for the creation of a new release.
If possible, this CI step is triggered automatically following a Tezos software release.

PR #305, opened 2021-09-22
[#286] Automate bottle hashes PR

### LINKED ISSUE #286: Automate PR generation to update bottle hashes

Following #284 and #285 we should have:
- CI step(s) to create and upload brew bottles
- a script to automate the generation of the hashes of said bottles and put it into the respective formulae

As the next step of automation, we should combine the two and have CI open PRs to update bottle hashes too. 

# Acceptance criteria

After bottles are created and uploaded to releases, CI open us PR(s) to update the bottle hashes in the brew formulae.

Note: it would be nice for these PRs to have a reminder in their descriptions that mirrored repositories should be updated too, or perhaps this could also be automated.

ixahedron:
## Description

Problem: we need to automate brew bottles-related processes further.
Right now we automatically build and attach the bottles to releases, and we have a script that computes and inserts their hashes to the respective formulae.
It's a natural next step to open the PRs from the CI.

Solution: added a step to each pipeline that should sync the hashes in a common branch, and then open a PR automatically.

## Related issue(s)

Resolves #286 

  can be successfully built.

### DIFF
--- a/.buildkite/pipeline-for-tags.yml
+++ b/.buildkite/pipeline-for-tags.yml
@@ -34,6 +34,15 @@ steps:
    - buildkite-agent artifact download "out/*" . --step build-source-packages
    - ./scripts/publish-native-packages.sh out
 
+ - label: Add Mojave bottle hashes to formulae
+   depends_on:
+   - "build-bottles"
+   if: build.tag =~ /^v.*/
+   commands:
+   - buildkite-agent artifact download --step build-bottles "*" .
+   - nix-shell ./scripts/shell.nix
+       --run './scripts/sync-bottle-hashes.sh "$BUILDKITE_TAG" "Mojave"'
+
  - label: Attach bottles to the release
    depends_on:
    - "build-bottles"
--- a/scripts/sync-bottle-hashes.sh
+++ b/scripts/sync-bottle-hashes.sh
@@ -1,0 +1,50 @@
+#! /usr/bin/env bash
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+# This script takes the new tag and the OS name as its arguments
+# and handles git shenanigans related to syncing two pipelines over a common branch.
+
+set -e
+
+if [ -z "$1" ] || [ -z "$2" ]; then
+    echo "Please call this script with the release tag and the OS name."
+    exit 1
+fi
+
+git config user.name "Serokell CI bot" # necessary for pushing
+git config user.email "hi@serokell.io"
+git fetch --all
+
+branch_name="auto/update-brew-formulae-$1"
+
+# Git doesn't have an easy way to check out a branch regardless of whether it exists.
+if ! git switch "$branch_name"; then
+    git switch -c "$branch_name"
+    git push --set-upstream origin "$branch_name"
+fi
+
+# Try to add hashes and push changes upstream. If there is a collision precisely at the time of
+# pushing, that means the other pipeline has raced this one to push. Reset to origin and try again.
+while : ; do
+    git fetch --all
+    git reset --hard origin/"$branch_name"
+    ./scripts/bottle-hashes.sh .
+    git commit -a -m "[Chore] Add $1 hashes to brew formulae"
+    ! git push || break
+done
+
+pr_body="Problem: we have built brew bottles for the new Octez release, but their hashes
+aren't in the formulae yet.
+
+Solution: added the hashes.
+"
+
+set +e
+
+# We create the PR with the first push, when the other pipeline hasn't finished yet.
+# That's why we 'set +e': one of the two times the command will fail.
+gh pr create -B master -t "[Chore] Add bottle hashes for $1" -b "$pr_body"
+
+exit 0
--- a/scripts/update-brew-formulae.sh
+++ b/scripts/update-brew-formulae.sh
@@ -1,0 +1,25 @@
+#! /usr/bin/env bash
+# SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+#
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+# This script takes the new tag as its argument.
+# Run it from the base directory (tezos-packaging).
+
+if [[ -d ./Formula ]]
+then
+    regex="(v.*)-[0-9]*"
+    if [[ $1 =~ $regex ]]; then
+        tag="${BASH_REMATCH[0]}"
+        version="${BASH_REMATCH[1]}"
+        find ./Formula -type f \( -name 'tezos-*.rb' ! -name 'tezos-sapling-params.rb' \) \
+            -exec sed -i "s/version \"v.*\"/version \"$tag\"/g" {} \; \
+            -exec sed -i "s/:tag => \"v.*\"/:tag => \"$version\"/g" {} \; \
+            -exec sed -i "/catalina/d" {} \; \
+            -exec sed -i "/mojave/d" {} \;
+    else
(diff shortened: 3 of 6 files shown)


.buildkite/pipeline-for-tags.yml
```diff
@@ -15,6 +15,10 @@ steps:
 
  - wait
 
+ - label: Add Mojave bottle hashes to formulae
```
rvem, line 18: I assume that this job won't have access to the bottles by default to calculate hashes, you'll have to download them as artifacts from the previous job using
```
buildkite-agent artifact download --step build-bottles "*"
```
See the `Attach bottles to the release` job

scripts/update-brew-formulae.sh
```diff
@@ -0,0 +1,25 @@
+#! /usr/bin/env bash
```
rvem, line 1: Is this script used anywhere? :cant-find:

ixahedron: Not yet, but it should find some use in #288. I wrote the script when I thought it might be used here and now and was told to 'keep it', so I thought I'd commit it anyway 🙈

scripts/sync-bottle-hashes.sh
```diff
@@ -0,0 +1,55 @@
...
+
+   pr_body="Add bottle hashes for $1
+
+Problem: we have built brew bottles for the new Octez release, but their hashes
+aren't in the formulae yet.
+
+Solution: added the hashes.
+"
+
+    # Git doesn't have an easy way to check out a branch regardless of whether it exists.
```
pasqu4le, line 29: I think that there's an easier way to do this syncing using `checkout`.

The flow should be something like this:
1. `git fetch --all`
2. `git checkout <branch_name>`
    Note that by default if a remote `<branch_name>` exists this will both create the local branch and start tracking the remote one. If the remote one doesn't exist it will fail.
3. Iff (2) failed then do `git checkout -b <branch_name>` and `git push -u <remote> <branch_name>`.
    Note that in the remote case in which both push at the same time `git push` won't fail (because they are pushing the same thing).
4. At this point we know that both remote and local branch exist and are synced (for now at least).
5.  `compute_and_commit` here
6. `git push`, or at least try to.
     If this fails it means that someone else pushed on the branch in the meanwhile.
7. Iff (6) fails, run `git fetch --all` and `git reset --hard <remote>/<branch_name>` then jump back to (4)

scripts/update-brew-formulae.sh
pasqu4le: I guess we can leave a TODO for that issue in here then.

scripts/sync-bottle-hashes.sh
ixahedron: Unless I'm missing something, this version loses the notion of whether the PR is already opened. It's not necessary a big deal, because GitHub doesn't allow opening two PRs from one branch, but it is, in my opinion, not as clean: we now have to open the PR from the first pipeline that finishes, meaning it will have only half of the work for an hour or so. It also means opening the PR will fail half of the tries.

But it is shorter, so I did like you suggested :/

rvem (approved):
Looks good now, I assume that you've tested it in some way

scripts/sync-bottle-hashes.sh
```diff
@@ -0,0 +1,55 @@
...
+# SPDX-License-Identifier: LicenseRef-MIT-TQ
+
+# This script takes the new tag and the OS name as its arguments
+# and handles git shenanigans related to syncing two pipelines over a common branch.
+
+set -e
+
+if [ -z "$1" ] || [ -z "$2" ]; then
+    echo "Please call this script with the release tag and the OS name."
+else
```
pasqu4le, line 13: Nitpick, but since there `set -e` above, this should work as well:
```suggestion
if [ -z "$1" ] || [ -z "$2" ]; then
    echo "Please call this script with the release tag and the OS name."
    exit 1
fi
```
and avoid the `else` and the resulting indentation (plus it returns an error code).

scripts/sync-bottle-hashes.sh
```diff
@@ -0,0 +1,55 @@
...
+    if ! git switch "$branch_name"; then
+        git switch -c "$branch_name"
+        git push --set-upstream origin "$branch_name"
+    fi
+
+    compute_and_commit "$2"
+
+    set +e
+    git push
+
```
pasqu4le, line 35: I believe this to be unnecessary (it's called in the `if` just below)?

scripts/sync-bottle-hashes.sh
```diff
@@ -0,0 +1,55 @@
...
+    if ! git push; then
+        # If there is a collision precisely at the time of pushing, that means the other pipeline
+        # has pushed. Reset to origin and try again. Once should always be enough.
+        set -e
+        git fetch --all
+        git reset --hard origin/"$branch_name"
+        compute_and_commit "$2"
+        git push
+    else
```
pasqu4le, line 44: > Once should always be enough

Well, only as long as there are no more than 2 bots, which might not always be the case.

Why not make this `if` a `while` instead? 

I believe that in that case the `fetch` and `reset` might be put in the `compute_and_commit` function as well.

Maybe it's even possible to do a `do ... while` loop and not use a function at all (it seems doable with [some tricks](https://stackoverflow.com/a/27761760/10986307)).

P.s. isn't the `git push` at the end here redundant as well?

scripts/sync-bottle-hashes.sh
```diff
@@ -0,0 +1,55 @@
...
+
+    function compute_and_commit {
+        ./scripts/bottle-hashes.sh .
+        git commit -a -m "[Chore] Add $1 hashes to brew formulae"
+    }
+
+    # Git doesn't have an easy way to check out a branch regardless of whether it exists.
+    if ! git switch "$branch_name"; then
+        git switch -c "$branch_name"
+        git push --set-upstream origin "$branch_name"
```
pasqu4le, line 28: cool, I didn't know about `git switch` :ok_hand:

scripts/sync-bottle-hashes.sh
pasqu4le: Oh I was pretty sure I left a comment about opening the PR, looks like I forgot or it got lost :disappointed: 

Anyway, yes, inevitably we'll rely on one to fail when the PR already exists, and do nothing there.

scripts/sync-bottle-hashes.sh
ixahedron: I disagree with turning it into a do-while loop in a way you linked. I tried it and IMO it's very unreadable if you're not an 80th level bash-fu master.
Is the way I tried in the last commit good enough? I think it should work the same.

pasqu4le: Oh, your solution is perfectly fine IMO :ok_hand: 

I mostly linked to that answer for the explanation :sweat_smile:

scripts/sync-bottle-hashes.sh
```diff
@@ -0,0 +1,48 @@
...
+aren't in the formulae yet.
+
+Solution: added the hashes.
+"
+
+set +e
+
+# We create the PR with the first push, when the other pipeline hasn't finished yet.
+# That's why we 'set +e': one of the two times the command will fail.
+gh pr create -B master -H -t "[Chore] Add bottle hashes for $1" -b "$pr_body"
```
pasqu4le, line 48: IIRC the return code of the whole script in bash is the same as the last command.

So I believe that to avoid one of the runs from failing we should have an `exit 0` at the end here

pasqu4le (commented):
Not entirely sure on how to do the testing, but it should be possible from here.

I suppose you could push a commit that:
- changes the pipelines to run bottle building on this branch
- removes the hashes from the formulae
- changes the script to first checkout this branch

I may be missing something tho

PR #335, opened 2021-11-12
[#334] Auto create release upon release PR merge

### LINKED ISSUE #334: Push versioned pre-release after octez release PR

We currently have a pipeline on `master` branch that will push an autorelease tag and related GitHub pre-release.
This allows us to check and collect the result of building (mostly static binaries) after an MR has been merged.

However, after a PR to support a new octez release is merged we need to manually turn the autorelease into a new one and push a different tag.
_Note: This can be done with a single action using the UI, but isn't possible at all with the `gh` cli utility_

This requires our presence both when merging said PR and after its pipeline is completed, which is going to be in most cases unnecessary, so we should instead consider to only require manual intervention if something goes wrong.

# Acceptance criteria

The `autorelease` CI step
- pushes a new version tag when running on the merge commit of a new octez release PR
- pushes the updated `autorelease` tag otherwise (as it does now)
- creates a pre-release accordingly either way

ixahedron:
Problem: we have automated almost the whole release making process except for actually publishing the release from autoreleases. Currently, after the release PR is merged and a auto pre-release made, we have to manually push a tag and publish a release with a correct name.

This has already lead to maintainers forgetting to publish a release, which led to some CI issues. We would like to mitigate this by automating this process as well.

Solution: reworked the auto-release making process to actually make full releases/pre-releases if we merged a release PR, or still make an `auto-release` if not.

## Related issue(s)

Resolves #334 

  can be successfully built.

### DIFF
--- a/scripts/autorelease.sh
+++ b/scripts/autorelease.sh
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: LicenseRef-MIT-TQ
 
+# This script takes the Buildkite build message, usually the last commit's title.
+
 set -euo pipefail
 
 # Project name, inferred from repository name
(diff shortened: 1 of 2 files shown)


scripts/autorelease.sh
```diff
@@ -27,21 +29,36 @@ shopt -s extglob
...
 
 # Update the tag
 git fetch # So that the script can be run from an arbitrary checkout
-git tag -f auto-release
+git tag -f $tag
 git push --force --tags
 
 # Create release
-gh release create -F "$TEMPDIR"/"$project"/release-notes.md --prerelease auto-release --title auto-release
+gh release create -F "$TEMPDIR"/"$project"/release-notes.md "$mode_flag" $tag --title $tag
```
rvem, line 61: I'm slightly afraid that here `gh` might not like `"$mode_flag"` here when the variable is empty because it will use it as an empty argument :thinking:
E.g.
```
$ dir=""
$ git log "$dir"
fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
```

scripts/autorelease.sh
```diff
@@ -27,21 +29,36 @@ shopt -s extglob
...
+    tag="auto-release"
+
+    # Delete autorelease if it exists
+    gh release delete auto-release --yes || true
+fi
 
 # Update the tag
 git fetch # So that the script can be run from an arbitrary checkout
-git tag -f auto-release
+git tag -f $tag
```
rvem, line 57:
```suggestion
git tag -f "$tag"
```

scripts/autorelease.sh
```diff
@@ -27,21 +29,36 @@ shopt -s extglob
...
 git push --force --tags
 
 # Create release
-gh release create -F "$TEMPDIR"/"$project"/release-notes.md --prerelease auto-release --title auto-release
+gh release create -F "$TEMPDIR"/"$project"/release-notes.md "$mode_flag" $tag --title $tag
 
 # Upload assets
-gh release upload auto-release "$assets_dir"/*
+gh release upload $tag "$assets_dir"/*
```
rvem, line 64:
```suggestion
gh release upload "$tag" "$assets_dir"/*
```

scripts/autorelease.sh
ixahedron: 🤔 You're right, might be a problem. As `$mode_flag` can only ever take one of two values (`""` or `"--prerelease"), I think it's fine to leave out the quotes to let it expand to nothing

rvem: I mean `$mode_flag` expanding to nothing may cause a falure

rvem: Because no argument and nothing argument may be handle differently, see my example with `git log`

rvem: oh, I see now that you've changed

rvem (approved):

PR #338, opened 2021-11-15
[#314] Fix 'not on branch' error on updating mirrors

### LINKED ISSUE #314: Automate updates of mirror repositories

We have two mirror repositories to this one that are used for macOS formulae distribution of stable releases vs release candidates, see the [macOS docs](https://github.com/serokell/tezos-packaging/blob/da3f3bb22e66954330e8654fd46e60463d8e35bd/docs/distros/macos.md).

These are periodically advanced, manually, to track a new commit in `master` when new Octez version come out and formulae are updated.

# Acceptance criteria

CI pipeline on `master` will update both mirror repositories when brew formulae have been updated either after:
- moving to a new version
- including new bottle hashes

ixahedron:
## Description

Problem: in a CI run after merging a release PR, updating the relevant mirror repo fails because we don't use the proper `--mirror` option.

Solution: added the flag to the commands.

## Related issue(s)

Resolves #314

  can be successfully built.

### DIFF
--- a/.buildkite/pipeline-raw.yml
+++ b/.buildkite/pipeline-raw.yml
@@ -145,6 +145,7 @@ steps:
    only_changes: *native_packaging_changes_regexes
 
  - label: create auto release/pre-release
+   key: auto-release
    commands:
    - mkdir binaries
    - mkdir arm-binaries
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline-raw.yml
```diff
@@ -170,14 +170,14 @@ steps:
      build.branch == "master" &&
        ( build.message =~ /^Merge pull request .* from serokell\/auto\/v[0-9]+\.[0-9]+-release/ ||
          build.message =~ /^Merge pull request .* from serokell\/auto\/update-brew-formulae-v[.0-9]+-[0-9]+/ )
-   command: git push git@github.com:serokell/tezos-packaging-stable.git
+   command: git push --mirror git@github.com:serokell/tezos-packaging-stable.git
 
  - label: update RC mirror repository
```
pasqu4le, line 175: I have one additional small suggestion here: make these two steps depend on the `autorelease` one.

`git push --mirror` will push all reference, including tags, which means that if we wait for the `autorelease` step we'll also have new tags pushed in the mirrors too

pasqu4le (approved):

PR #345, opened 2021-11-29
[#323] Add dependencies to tezos-sandbox formula

### LINKED ISSUE #323: [BUG] tezos-sandbox homebrew formula missing dependencies
# Description

Hi there. I've found that the tezos-sandbox formula is missing some dependencies. This is known and here's an extract from https://tezos.gitlab.io/developer/flextesa.html:

> If you are on Mac OS X, you can do brew install coreutils util-linux. Then run the tests with:

```
export PATH="/usr/local/opt/coreutils/libexec/gnubin:/usr/local/opt/util-linux/bin:$PATH"
```

Perhaps we could add coreutils and util-linux as dependencies to the formula. As for adjusting the PATH, perhaps we could have the formula emit a notice or append the given export command above to ~/.bashrc and ~/.zshrc if either exist?

# To Reproduce

1. Install tezos-sandbox via homebrew: `brew install tezos-sandbox`
2. Run `tezos-sandbox mini-net`
3. You should see the following output:
```
Flextesa:
  Genesis-block-hash not set, using:
  `BLdZYwNF8Rn6zrTWkuRRNyrj6bQWPkfBog2YKhWhn5z3ApmpzBf` (corresponding
  chain-id: `NetXKMbjQL2SBox`) (from user choice: “Default”).
Flextesa:
  An error happened
    Error:
        Process-error: Non-zero exit status: Shell command: "mkdir -p
        '/Users/mweichert/tezos-mininet'"; wrong-status: 'exited with 127'..
    * stdout: EMPTY
    * stderr: EMPTY
Flextesa: Killing all processes.
Flextesa: Waiting for processes to all die.
Flextesa: Error: Exiting with 4.
```

# Expected behavior

No error should be thrown.

# Environment

- macos Big Sur 11.3.1 (Darwin MacBook-Pro 20.4.0 Darwin Kernel Version 20.4.0: Thu Apr 22 21:46:47 PDT 2021; root:xnu-7195.101.2~1/RELEASE_X86_64 x86_64 i386 MacBookPro11,5 Darwin)
- tezos-sandbox/v10.3-1

# Additional context

In the $ROOT_DIR, stderr.log and stdout.log are empty but meta.log contains:
```
================================================================================
Date: 2021-11-02T22:47:47.545-00:00
Starting: cmd-00001-200620895bdd43322124ed9a44532695-03423878
Cmd: ["setsid"; "sh"; "-c"; "mkdir -p '/tmp/mininet-test'"]
================================================================================
```

Question: Where can I find the bash script that is failing? Perhaps that could be distributed in the formula to /usr/local/share or something like that, and have it utilized by the formula / binary some how. Thanks!

ixahedron:
## Description

Problem: there is a known issue with using tezos-sandbox on MacOS, which is that it depends on `coreutils` and `util-linux`. These are not installed upon running `brew install tezos-sandbox`.
Additionally, homebrew installs all tools that conflict with MacOS ones in name with the `g` prefix, meaning the user needs to manipulate `$PATH` to use these dependencies.

Solution: added `coreutils` and `util-linux` as dependencies to the `tezos-sandbox` formula. As brew does not allow for post-setup user configuration manipulation, `$PATH` cannot be easily modified to include `gnubin`. We use the paradigmatic homebrew way and output the command in the caveats section for the user to run manually.

## Related issue(s)

Resolves #323 

  can be successfully built.

### DIFF
--- a/Formula/tezos-sandbox.rb
+++ b/Formula/tezos-sandbox.rb
@@ -19,7 +19,7 @@ class TezosSandbox < Formula
     depends_on dependency => :build
   end
 
-  dependencies = %w[gmp hidapi libev libffi]
+  dependencies = %w[gmp hidapi libev libffi coreutils util-linux]
   dependencies.each do |dependency|
     depends_on dependency
   end
@@ -54,5 +54,14 @@ class TezosSandbox < Formula
     install_template "src/bin_sandbox/main.exe",
                      "_build/default/src/bin_sandbox/main.exe",
                      "tezos-sandbox"
+  end
+
+  # homebrew does not allow for post-setup modification of user files,
+  # so we have to provide a caveat to be displayed for the user to adjust $PATH manually
+  def caveats
+    <<~EOS
+      tezos-sandbox depends on 'coreutils' and 'util-linux', which have been installed. Please run the following command to bring them in scope:
+        export PATH=#{HOMEBREW_PREFIX}/opt/coreutils/libexec/gnubin:#{HOMEBREW_PREFIX}/opt/util-linux/bin:$PATH
+    EOS
   end
 end


rvem (approved):
LGTM, don't forget to update the stable mirror

PR #347, opened 2021-12-03
[#237] Create the first version of a voting wizard

### LINKED ISSUE #237: Create a voting wizard

This is an issue related to #201, where the wizard described there is for _setting up_ a node/baking instance, the one here is for a wizard to _perform_ voting.

`tezos-client` has a command to submit votes, see `submit ballot for`, which is fairly straightforward
However there are still part of the command that we can provide as a more convenient interactive choice (mainly the proposal hash to choose from) and we should do so.

# Acceptance criteria

A new interactive CLI wizard is provided in native packages that allows a voter to choose the options and submit a vote.

ixahedron:
## Description

Problem: bakers who want to vote on the Tezos protocol amendment currently have to do it through base `tezos-client` or other unintuitive avenues. As we already have an interactive baking setup wizard, we would like to extend interactive capabilities of our packages and provide a similar wizard for voting.

Solution: added the first version of Tezos Voting Wizard.

## Related issue(s)

Resolves #237 

  can be successfully built.

### DIFF
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -563,6 +563,7 @@ class TezosBakingServicesPackage(AbstractPackage):
     def fetch_sources(self, out_dir):
         os.makedirs(out_dir)
         shutil.copy(f"{os.path.dirname(__file__)}/tezos_setup_wizard.py", out_dir)
+        shutil.copy(f"{os.path.dirname(__file__)}/tezos_voting_wizard.py", out_dir)
 
     def gen_control_file(self, deps, ubuntu_version, out):
         run_deps_list = ["acl", "tezos-client", "tezos-node"]
@@ -648,9 +649,14 @@ tezos-setup-wizard:
 	mv $(CURDIR)/tezos_setup_wizard.py $(CURDIR)/tezos-setup-wizard
 	chmod +x $(CURDIR)/tezos-setup-wizard
 
-install: tezos-baking tezos-setup-wizard
+tezos-voting-wizard:
+	mv $(CURDIR)/tezos_voting_wizard.py $(CURDIR)/tezos-voting-wizard
+	chmod +x $(CURDIR)/tezos-voting-wizard
+
+install: tezos-baking tezos-setup-wizard tezos-voting-wizard
 	mkdir -p $(DESTDIR)$(BINDIR)
 	cp $(CURDIR)/tezos-setup-wizard $(DESTDIR)$(BINDIR)/tezos-setup-wizard
+	cp $(CURDIR)/tezos-voting-wizard $(DESTDIR)$(BINDIR)/tezos-voting-wizard
 """
         with open(out, "w") as f:
             f.write(file_contents)
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -562,6 +562,7 @@ class TezosBakingServicesPackage(AbstractPackage):
 
     def fetch_sources(self, out_dir):
         os.makedirs(out_dir)
+        shutil.copy(f"{os.path.dirname(__file__)}/wizard_structure.py", out_dir)
         shutil.copy(f"{os.path.dirname(__file__)}/tezos_setup_wizard.py", out_dir)
         shutil.copy(f"{os.path.dirname(__file__)}/tezos_voting_wizard.py", out_dir)
 
@@ -655,6 +656,7 @@ tezos-voting-wizard:
 
 install: tezos-baking tezos-setup-wizard tezos-voting-wizard
 	mkdir -p $(DESTDIR)$(BINDIR)
+	cp $(CURDIR)/wizard_structure.py $(DESTDIR)$(BINDIR)/wizard_structure.py
 	cp $(CURDIR)/tezos-setup-wizard $(DESTDIR)$(BINDIR)/tezos-setup-wizard
 	cp $(CURDIR)/tezos-voting-wizard $(DESTDIR)$(BINDIR)/tezos-voting-wizard
 """
--- a/docker/package/tezos_voting_wizard.py
+++ b/docker/package/tezos_voting_wizard.py
@@ -16,6 +16,11 @@ import re, textwrap
 
 # Global options
 
+networks = {
+    "mainnet": "Main Tezos network",
+    "custom": "Custom network, will require its name",
+}
+
 ballot_outcomes = {
     "yay": "Vote for accepting the proposal",
     "nay": "Vote for rejecting the proposal",
@@ -92,11 +97,14 @@ class Validator:
 welcome_text = """Tezos Voting Wizard
 
 Welcome, this wizard will help you to vote in the Tezos protocol amendment process.
-Please note that for this you need to already have set up the baking infrastructure
-on mainnet, as only bakers can submit ballots or proposals.
+Please note that for this you need to already have set up the baking infrastructure,
+normally on mainnet, as only bakers can submit ballots or proposals.
 
 If you have installed tezos-baking, you can run Tezos Setup Wizard to set it up:
 tezos-setup-wizard
+
+Alternatively, you can use this wizard for voting on a custom chain,
+which also needs to be set up already.
 
 All commands within the service are run under the 'tezos' user.
 
@@ -223,7 +231,7 @@ custom_network_query = Step(
     help="The selected network will be used to vote.\n"
     "For example, to use the tezos-baking-custom@voting service, input 'voting'."
     "You need to already have set up the custom network using systemd services.",
-    # TODO: "\nFor that, you can follow a tutorial here:",
+    # TODO: "\nYou can follow a tutorial here:",
     validator=Validator(required_field_validator),
 )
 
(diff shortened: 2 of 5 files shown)


docker/package/tezos_voting_wizard.py
```diff
@@ -0,0 +1,512 @@
...
+
+
+# To be validated, the input should adhere to the protocol hash format:
+# <base58 encoded string with length 51 starting with P>
+def protocol_hash_validator(input):
+    match = re.match(protocol_hash_regex.decode("utf-8"), input.strip())
+    if not bool(match):
+        raise ValueError(
+            "The input doesn't match the format for the protocol hash: "
+            "P[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{50}"
```
rvem, line 69: Is it possible to use `protocol_hash_regex`?

docker/package/tezos_voting_wizard.py
```diff
@@ -0,0 +1,512 @@
...
+        print(f"Data directory: {self.config['client_data_dir']}")
+        print(f"Node RPC address: {self.config['node_rpc_addr']}")
+        print(f"Baker alias: {self.config['baker_alias']}")
+        if not yes_or_no("Does this look correct? (Y/n)", "yes"):
+            print("Try setting up baking again by running:")
+            print("tezos-setup-wizard")
+            sys.exit(1)
+
+    def fill_baking_config(self):
+        output = get_proc_output("systemctl show tezos-baking-mainnet.service").stdout
```
rvem, line 315: Seems like we need #349 in order to test this wizard and a way to choose the network that is going to be used during the voting

docker/package/tezos_voting_wizard.py
```diff
@@ -0,0 +1,559 @@
...
+        ),
+    )
+
+
+ballot_outcome_query = Step(
+    id="ballot_outcome",
+    prompt="Choose the outcome for your ballot.",
+    help="'yay' is for supporting the proposal, 'nay' is for rejecting the proposal, "
+    "'pass' is used to not influence a vote but still contribute to reaching a quorum.",
+    options=ballot_outcomes,
```
rvem, line 277: Let's add `default=None` here just for safety so that users cannot accidentally click 'Enter' and vote `yay`

docker/package/tezos_voting_wizard.py
```diff
@@ -0,0 +1,559 @@
...
+            return True
+        elif answer in ["n", "no"]:
+            print()
+            return False
+        else:
+            print(color("Please provide a 'yes' or 'no' answer.", color_red))
+
+
+# we don't need any data here, just a confirmation Tezos Wallet app is open
+def wait_for_ledger_baking_app():
```
rvem, line 153:
```suggestion
def wait_for_ledger_wallet_app():
```

docker/package/tezos_voting_wizard.py
```diff
@@ -0,0 +1,559 @@
...
+        # check with the user that the baker data sounds correct
+        self.check_data_correctness()
+
+        self.config["tezos_client_options"] = self.get_tezos_client_options()
+
+        # if a ledger is used for baking, ask to open Tezos Wallet app on it before proceeding
+        if self.check_ledger_use():
+            print()
+            print("Please open the Tezos Wallet app on your ledger.")
+            wait_for_ledger_baking_app()
```
rvem, line 510:
```suggestion
            wait_for_ledger_wallet_app()
```

docker/package/tezos_voting_wizard.py
```diff
@@ -0,0 +1,559 @@
...
+                print("This means you are not present in the voting listings.")
+            elif re.search(b"Too many proposals", result.stderr) is not None:
+                print(
+                    color(
+                        "Cannot submit because of too many proposals submitted.",
+                        color_red,
+                    )
+                )
+                print("This means you have already submitted more than 20 proposals.")
+            # No other "legitimate" proposal error ('empty_proposal', 'unexpected_proposal')
```
rvem, line 444: Do I understand correctly that 'unexpected_proposal' happens when the proposal hash is malformed and this is caught by the regex check in our code?

ixahedron: No, that should be the `Invalid proposal` one, and our regex can't catch all invalid hashes, it mostly just checks the length.
I'm not sure what the `unexpected_proposal` is for 😕 The docs mention it's when the voting is not in the 'proposal' period, but in that case, `tezos-client` reports "Not in a proposal period".

docs/tests/voting.md
```diff
@@ -0,0 +1,78 @@
...
+
+1) Generate a pair of keys associated with `baker` alias:
+
+```
+sudo -u tezos tezos-client gen keys baker
+```
+
+2) In a separate terminal start a [voting scenario script](https://gitlab.com/morley-framework/local-chain#voting-scenario) from the local-chain repo.
+
+This script will provide you a path to the node config that will be used by the custom baking service.
```
ixahedron, line 27: Ok, I found where I was having problems. For some reason the node couldn't use/recognise the config that was generated by the `local-chain` scenario in my home directory 🤦 Moving it to `/var/lib/tezos/` helped

rvem: hmm, for me even path in home directory works fine :thinking:

DMozhevitin:
It seems that if the user's node isn't connected with the node that is running by `voting.py`, then `tezos-client show voting period` command's stdout will be empty, and `fill_voting_period_info` in `tezos_setup_wizard.py` will wail with `'NoneType' object has no attribute 'group'` error.

That's indeed is an invalid situation, but maybe it worth adding some explanation to this error for wizard users.

DMozhevitin:
Btw, there is some duplications in `tezos_setup_wizard` and `tezos_voting_wizard`, because they really have the similar interface. I think it's absolutely ok for the case when we only have 2 wizards.

If in the future we need to write some other wizards, maybe it worth to extract their common part to the library to speed up the development in the future? :thinking:

Not sure if we plan to develop more wizards, and if it's worth it at all, but just in case, decided to write about it

ixahedron:
> If in the future we need to write some other wizards, maybe it worth to extract their common part to the library to speed up the development in the future?

Oh yeah, that's a very reasonable suggestion! I actually already did most of that but it's not included in the PR yet because of the WIP status, so thanks for reminding me. I'll try to finish and push it tomorrow, to be reviewed before the merge

docker/package/model.py
```diff
@@ -644,13 +646,22 @@ def gen_makefile(self, out):
 
 tezos-baking:
 
+wizard-structure:
+	mv $(CURDIR)/wizard_structure.py $(CURDIR)/wizard_structure.py
```
rvem, line 650: :monkahmm:

ixahedron: :laughing: Def. What would you suggest?

rvem: Is `wizard-structure` target really needed?
I thnk you can safely omit it

rvem: Moreover, seems like it actually causes `tezos-baking` binary package build to fail:
```
mv: '/tezos-packaging/docker/tezos-baking-12.0-rc2/wizard_structure.py' and '/tezos-packaging/docker/tezos-baking-12.0-rc2/wizard_structure.py' are the same file
make[1]: *** [Makefile:9: wizard-structure] Error 1
make[1]: *** Waiting for unfinished jobs....
```

docker/package/model.py
```diff
@@ -644,13 +646,22 @@ def gen_makefile(self, out):
...
 	chmod +x $(CURDIR)/tezos-setup-wizard
 
-install: tezos-baking tezos-setup-wizard
+tezos-voting-wizard:
+	mv $(CURDIR)/tezos_voting_wizard.py $(CURDIR)/tezos-voting-wizard
+	chmod +x $(CURDIR)/tezos-voting-wizard
+
+install: wizard-structure tezos-baking tezos-setup-wizard tezos-voting-wizard
 	mkdir -p $(DESTDIR)$(BINDIR)
+	cp $(CURDIR)/wizard_structure.py $(DESTDIR)$(BINDIR)/wizard_structure.py
```
rvem, line 662: TBH, I'm not sure that it's a good idea to put `*.py` files to the `/usr/bin`.

Seems like there is a proper approach for packaging python applications for Debian, see [this doc](https://wiki.debian.org/Python/LibraryStyleGuide?action=show&redirect=Python%2FPackaging).

However, I don't really insist on doing everything properly in this PR since it's already open for quite a long time, and we'd like to include it into the next release, but please create a follow-up issue:)

rvem:
Tested updated versions of both wizards, they work fine :relieved:

docker/package/model.py
ixahedron: I agree. I'll create the issue as soon as we merge the PR, to have something to link to

rvem (approved):

DMozhevitin:
Works fine for me too :+1:

PR #350, opened 2021-12-09
[Chore] Disable granada support

rvem:
## Description
Problem: Hangzhou protocol was activated on mainnet and granada is now
obsolete. There is no sense to provide granada-related functionality in
our packages.

Solution: Remove granada support.

## Related issue(s)

Resolves #

  can be successfully built.


pasqu4le (approved):
:ok_hand: nothing seems missing to me

PR #352, opened 2021-12-15
[#346] Document versioning policy

### LINKED ISSUE #346: Document versioning policy and scheme

The versioning policy used by `tezos-packaging` can be a bit confusing and there is no documentation to clarify this.

_Note: this is somewhat related to #307._

For example, version `10.3-1` was mistaken to be `10.3.1` (which doesn't exist) and while the dash makes a difference, there is no explanation in the repo for _what the difference is_, which is that we use an RPM-like versioning scheme.

Moreover, we should provide info for where/how releases are provided, for example:
- octez RCs only become pre-releases in the [release page](https://github.com/serokell/tezos-packaging/releases) differently from releases
- likewise there are different PPAs that distribute RCs and stable releases
- there are two mirror repositories for stable and candidate releases for macOS

most (if not all) of this info is already in the related documentation, but it would be nice to have it all in one place for those who are interested.

# Acceptance criteria

There is a document in the repo describing the versioning policy and scheme

rvem:
## Description
Problem: Our versioning policy isn't documented anywhere, as a result
users may get confused when they see a package with a version that is
different from the upstream release version.

Solution: Write a doc that explains our versioning policy and the way in
which packages for upstream releases and pre-releases are distributed.
Also, refer to this doc in the main README.

## Related issue(s)

Resolves #346

  can be successfully built.

### DIFF
--- a/docs/versioning.md
+++ b/docs/versioning.md
@@ -1,0 +1,52 @@
+<!--
+   - SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+   -
+   - SPDX-License-Identifier: LicenseRef-MIT-TQ
+   -->
+
+# Versioning
+
+`tezos-packaging` follows all Octez releases, both stable and candidates.
+
+Packages from our releases provide addinional functionality (e.g. systemd services
+in Ubuntu and Fedora package or brew formulae with launchd services).
+This additional functionality may change within the same upstream version.
+
+In order to track this, our GitHub releases and packages use
+the following RPM-like versioning scheme: `<name>-<version>-<release>`:
+* `<name>` is used to mention which Octez binary is packaged.
+* `<version>` is used to reference the packaged upstream version.
+* `<release>` is used to reflect changes in additional packages functionality.
+
+## GitHub releases
+
+We provide GitHub releases for stable upstream releases and pre-releases for release candidates.
+
+In our GitHub repository we use tags in which `<name>-` part is ommited.
+
+E.g. `v11.0-1` is the first `tezos-packaging` release within the `v11.0` upstream stable release,
+or `v11.0-rc2-2` for the second `tezos-packaging` release within the `v11.0-rc2` upstream release candidate.
+
+GitHub {pre-}releases contain static binaries and brew bottles compiled from the given
+upstream source version.
+
+## Ubuntu packages
+
+Ubuntu packages use a slightly different versioning scheme, which follows
+the [Debian versioning policy](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version):
+`<name>-<version>-0ubuntu<release>~<ubuntu-version>`.
+
+E.g. `tezos-client-11.0+no-adx-0ubuntu1~focal`, where `focal` is `20.04 LTS`.
+
+We use different PPA repositories for stable and release candidate Ubuntu packages.
+You can read more about this in the [doc about Ubuntu packages](./distros/ubuntu.md).
+
+## Fedora packages
+
+We use different Copr projects for stable and release candidate Fedora packages.
+You can read more about this in the [doc about Fedora packages](./distros/fedora.md).
+
+## Brew formulae
+
+We use two distinct repository mirrors to provide stable and release candidate brew formulae.
+You can read more about this in the [doc about macOS packaging](./distros/macos.md).
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ This repo provides various forms of distribution for [Tezos](http://tezos.gitlab
 See the [official documentation](http://tezos.gitlab.io/introduction/howtouse.html)
 for information about the binaries, their usage, and concepts about the Tezos networks.
 
+See the [versioning doc](./docs/versioning.md) for information about the versioning
+policy for the provided forms of distribution.
+
 ## Set up a node and/or baking on Ubuntu
 
 The simplest procedure to set up a node and/or baking instance is provided for Ubuntu.


docs/versioning.md
```diff
@@ -0,0 +1,48 @@
+<!--
+   - SPDX-FileCopyrightText: 2021 TQ Tezos <https://tqtezos.com/>
+   -
+   - SPDX-License-Identifier: LicenseRef-MIT-TQ
+   -->
+
+# Versioning
+
+`tezos-packaging` provides support for all Octez stable releases and RC candidates.
```
pasqu4le, line 9: :thinking: `support` on it's own seems a bit odd IMO, it kinda sounds like this repository offers technical assistance on releases.
Also, the `C` in `RC` stands for candidate, so there is duplication there :slightly_smiling_face: 

What do you think of:
```suggestion
`tezos-packaging` follows all Octez releases, both stable and candidates.
```

docs/versioning.md
```diff
@@ -0,0 +1,48 @@
...
+
+However, in some cases we provide multiple releases within the same upstream version
+because our packages provide additional functionality (e.g. systemd services in Ubuntu
+and Fedora packages or brew formulae with launchd services) that may need to be updated
+within the single upstream release.
+
+In order to achieve that, our GitHub releases and packages use the following RPM-like versioning scheme:
+`<upstream-version>-<release-number>`, e.g. `v11.0-1` for the first `tezos-packaging` release
+within the `v11.0` upstream stable release, or `v11.0-rc2-2` for the second `tezos-packaging` release
+within the `v11.0-rc2` upstream release candidate.
```
pasqu4le, line 19: from what i read only RPM versioning is more like:
> [name]-[version]-[release]

all of these 3 values apply to our versioning scheme too. So I suggest:
-  to mention that this scheme is applied to binaries too, which have a `name`
- I would use `version` more in the explanations to indicate the upstream releases

docs/versioning.md
```diff
@@ -0,0 +1,48 @@
...
+   -->
+
+# Versioning
+
+`tezos-packaging` provides support for all Octez stable releases and RC candidates.
+
+However, in some cases we provide multiple releases within the same upstream version
+because our packages provide additional functionality (e.g. systemd services in Ubuntu
+and Fedora packages or brew formulae with launchd services) that may need to be updated
+within the single upstream release.
```
pasqu4le, line 14: Minor suggestion, buy I would rephrase this to more have this order instead:
1. tezos-packaging releases can provide additional functionalities
2. give examples
3. (implied) these are tracked in addition top of the upstream version


P.s it sounds more correct to me to say "the same" rather than "the single" at the end there

docs/versioning.md
```diff
@@ -0,0 +1,48 @@
...
+In order to achieve that, our GitHub releases and packages use the following RPM-like versioning scheme:
+`<upstream-version>-<release-number>`, e.g. `v11.0-1` for the first `tezos-packaging` release
+within the `v11.0` upstream stable release, or `v11.0-rc2-2` for the second `tezos-packaging` release
+within the `v11.0-rc2` upstream release candidate.
+
+`<release-number>` is used to reflect the changes in additional packages functionality.
+
+## Github releases
+
+We provide at github releases for stable upstream releases and pre-releases for release candidates.
```
pasqu4le, line 25:
```suggestion
We provide github releases for stable upstream releases and pre-releases for release candidates.
```

docs/versioning.md
```diff
@@ -0,0 +1,48 @@
...
+within the single upstream release.
+
+In order to achieve that, our GitHub releases and packages use the following RPM-like versioning scheme:
+`<upstream-version>-<release-number>`, e.g. `v11.0-1` for the first `tezos-packaging` release
+within the `v11.0` upstream stable release, or `v11.0-rc2-2` for the second `tezos-packaging` release
+within the `v11.0-rc2` upstream release candidate.
+
+`<release-number>` is used to reflect the changes in additional packages functionality.
+
+## Github releases
```
pasqu4le, line 23: I think it would be better to always capitalize GitHub properly (like I just did :sweat_smile: )

docs/versioning.md
```diff
@@ -0,0 +1,48 @@
...
+
+We provide at github releases for stable upstream releases and pre-releases for release candidates.
+
+Github {pre-}releases contain static binaries and brew bottles compiled from the given
+upstream source version.
+
+## Ubuntu packages
+
+Ubuntu packages use slightly different versioning scheme, which is used to follow
```
pasqu4le, line 32:
```suggestion
Ubuntu packages use a slightly different versioning scheme, which follows
```

docs/versioning.md
```diff
@@ -0,0 +1,52 @@
...
+   - SPDX-License-Identifier: LicenseRef-MIT-TQ
+   -->
+
+# Versioning
+
+`tezos-packaging` follows all Octez releases, both stable and candidates.
+
+Packages from our releases provide addinional functionality (e.g. systemd services
+in Ubuntu and Fedora package or brew formulae with launchd services).
+This additional functionality may change within the same upstream version
```
pasqu4le, line 13:
```suggestion
This additional functionality may change within the same upstream version.
```

docs/versioning.md
```diff
@@ -0,0 +1,52 @@
...
+
+In order to track these additional functionality chages, our GitHub releases and packages use
```
pasqu4le, line 15: Small suggestion:
```suggestion
In order to track this, our GitHub releases and packages use
```
for two reasons:
1. "changes" was misspelled
2. "additional functionality" is in both the previous two sentences as well as in one of the points just following, I think it's worth removing _some_ of this repetition :sweat_smile:

pasqu4le (approved):
Two minor suggestions, but otherwise LGTM :ok_hand:

Issue #205, opened 2021-05-23
[BUG] FAILED to install on Mac M1 with Rust 1.44.0


phanngt:
# Description

# To Reproduce

```Run brew install tezos-client on Mac M1```

# Environment

```OS: MacOs BigSur 11.3.1```

# Additional context

```==> rustup-init --default-toolchain 1.44.0 -y
Last 15 lines from /Users/phannguyen/Library/Logs/Homebrew/tezos-client/01.rustup-init:
2021-05-23 15:45:09 +0700

rustup-init
--default-toolchain
1.44.0
-y

error: $HOME differs from euid-obtained home directory: you may be using sudo
error: $HOME directory: /private/tmp/tezos-client-20210523-17574-kimclr/.brew_home
error: euid-obtained home directory: /Users/phannguyen
info: profile set to 'default'
info: default host triple is aarch64-apple-darwin
info: syncing channel updates for '1.44.0-aarch64-apple-darwin'
info: latest update on 2020-06-04, rust version 1.44.0 (49cae5576 2020-06-01)
error: target 'aarch64-apple-darwin' not found in channel.  Perhaps check https://doc.rust-lang.org/nightly/rustc/platform-support.html for available targets

Do not report this issue to Homebrew/brew or Homebrew/core!

Traceback (most recent call last):
	26: from /opt/homebrew/Library/Homebrew/build.rb:229:in `<main>'
	25: from /opt/homebrew/Library/Homebrew/build.rb:133:in `install'
	24: from /opt/homebrew/Library/Homebrew/utils.rb:558:in `with_env'
	23: from /opt/homebrew/Library/Homebrew/build.rb:138:in `block in install'
	22: from /opt/homebrew/Library/Homebrew/formula.rb:1259:in `brew'
	21: from /opt/homebrew/Library/Homebrew/formula.rb:2318:in `stage'
	20: from /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/forwardable.rb:230:in `stage'
	19: from /opt/homebrew/Library/Homebrew/resource.rb:91:in `stage'
	18: from /opt/homebrew/Library/Homebrew/resource.rb:116:in `unpack'
	17: from /opt/homebrew/Library/Homebrew/resource.rb:199:in `mktemp'
	16: from /opt/homebrew/Library/Homebrew/mktemp.rb:63:in `run'
[...]

PR #356, opened 2021-12-17
[#205] Provide M1 support

### LINKED ISSUE #354: Provide instructions for using brew bottles on mac with M1 CPU

Unfortunately, Octez currently doesn't support M1 processors. However, there is a rosetta binary translator that can be used for building and running Octez binaries on M1.

One of the users managed to use our brew formulae under rosetta, see [this comment](https://github.com/serokell/tezos-packaging/issues/205#issuecomment-871438693) under #205.

It would be nice to provide such instructions in our official usage documentation.

# Acceptance criteria

[macOS usage doc](https://github.com/serokell/tezos-packaging/blob/master/docs/distros/macos.md) explains how to use brew formulae with Octez binaries on M1 using rosetta.

rvem:
## Description
Problem: Currently formulae download opam x86_64 binary from the
GitHub release, which causes a build failure on machines with M1 CPU.

Solution: Determine CPU architecture in runtime and download appropriate
opam binary.

## Related issue(s)

Relates #205 (will be resolved once mirror repository is updated)
Resolves #354 (documentation update won't be needed since we're going to support M1 natively)

  can be successfully built.

### DIFF
--- a/Formula/tezos-accuser-011-PtHangz2.rb
+++ b/Formula/tezos-accuser-011-PtHangz2.rb
@@ -12,7 +12,7 @@ class TezosAccuser011Pthangz2 < Formula
 
   url "https://gitlab.com/tezos/tezos.git", :tag => "v11.0+no_adx", :shallow => false
 
-  version "v11.0+no_adx-1"
+  version "v11.0+no_adx-2"
 
   build_dependencies = %w[pkg-config autoconf rsync wget rustup-init]
   build_dependencies.each do |dependency|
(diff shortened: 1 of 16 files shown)


ixahedron (approved):

Formula/tezos-accuser-011-PtHangz2.rb
```diff
@@ -27,15 +27,14 @@ class << self
 
   bottle do
     root_url "https://github.com/serokell/tezos-packaging/releases/download/#{TezosAccuser011Pthangz2.version}/"
```
pasqu4le, line 29: Are bottles independent from architectures?

I see that currently released bottles include in their name the package version and the macOS version used, but no architecture.

rvem: Seems like brew is able to handle this, there are `big_sur` and `arm64_big_sur`, see [this formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/acmetool.rb#L12) as an example

rvem: > Are bottles independent from architectures?

So, the answer is no

pasqu4le: Never mind, I see from [the documentation](https://docs.brew.sh/Bottles) that architecture and OS version are handled together, so this `root_url` still works.

pasqu4le (approved):

Issue #353, opened 2021-12-16
Provide Big Sur brew bottles in releases


rvem:
# Clarification and motivation

Github actions now provide runners with macOS Big Sur, which can be used to build, see [this blogpost](https://github.blog/changelog/2021-08-16-github-actions-macos-11-big-sur-is-generally-available-on-github-hosted-runners/).

We should use this ability for the time we don't have an ability to build bottles for Big Sur for the time being while
we don't have our own infra for building Big Sur bottles.

# Acceptance criteria

CI builds and uploads Big Sure bottles for each release in this repository.

PR #357, opened 2021-12-17
[#353] Big Sur GitHub actions

rvem:
## Description
Problem: At the moment we provide bottles only for Mojave and Catalina,
and both of these versions are quite old at the moment. Service support
for Mojave stopped some time ago.

Solution: Build bottles for x86_64 Big Sur using GitHub actions since
they recently started providing runners with this macOS version.

You can check [this pipeline](https://github.com/serokell/tezos-packaging/runs/4561166171) to ensure that it actually works.

## Related issue(s)

Resolves first part of #353. `arm64` support will be provided in another PR.

  can be successfully built.

### DIFF
--- a/.github/workflows/build-bottles.yml
+++ b/.github/workflows/build-bottles.yml
@@ -8,7 +8,7 @@ permissions:
   # Restrict GITHUB_TOKEN permissions
   contents: write
 jobs:
-  build-bottles:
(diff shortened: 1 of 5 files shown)


ixahedron (approved):
When are we planning to drop support for Mojave? I imagine without service support, there can't be that many users left?

PR #365, opened 2022-01-03
[#364] Support archive nodes from snapshot import

### LINKED ISSUE #364: Support archive node with --reconstruct in wizard
Hi, I'd like to run an archive node.  So far the only way I found to make this is to edit the tezos-setup-wizard file like this:

```
sed -i.bak 's| snapshot import | snapshot import --reconstruct |g' /usr/bin/tezos-setup-wizard
```

It'd be nice to offer this as an option in the wizard.

ixahedron:
## Description

Problem: there is a use-case for being able to reconstruct storage to run an archive node when importing a full snapshot. The Setup Wizard at the moment doesn't support running archive nodes in any capacity except by manually changing the code, which isn't very convenient.

Solution: if the snapshot the user wants to import is in full history mode, inquire whether the storage should be reconstructed to yield an archive node.

## Related issue(s)

Resolves #364 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -51,6 +51,8 @@ systemd_enable = {
 }
 
 
+TMP_SNAPSHOT_LOCATION = "/tmp/tezos_node.snapshot"
+
 # Regexes
 
 ledger_regex = b"ledger:\/\/[\w\-]+\/[\w\-]+\/[\w']+\/[\w']+"
@@ -178,7 +180,7 @@ def get_proc_output(cmd):
 
 def fetch_snapshot(url):
     print("Downloading the snapshot from", url)
-    filename = "/tmp/tezos_node.snapshot"
+    filename = TMP_SNAPSHOT_LOCATION
     urllib.request.urlretrieve(url, filename, progressbar_hook)
     print()
     return filename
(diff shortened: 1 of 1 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -621,18 +635,24 @@ def import_snapshot(self):
 
             valid_choice = True
 
+            import_flag = ""
+            if is_full_snapshot(self.config["snapshot"]):
+                if yes_or_no("Do you want to run an archive node? (y/N) ", "no"):
+                    import_flag = "--reconstruct "
```
pasqu4le, line 641: A small suggestion, something like:
```suggestion
                if yes_or_no("Do you also want to reconstruct all the chain data to run an archive node? (y/N) ", "no"):
                    import_flag = "--reconstruct "
```
might help make more clear that choosing yes here will take a lot more effort :slightly_smiling_face: 

However considering that the default option is sane and anyone confused will _hopefully_ take it, this might not be that important. I'll let you choose.

pasqu4le (approved):
LGTM, I tested it as well and everything works nicely.

docker/package/tezos_setup_wizard.py
ixahedron: I like your wording better 🙂

PR #367, opened 2022-01-11
[#366] Add history mode support in setup wizard

### LINKED ISSUE #366: Add a node history mode option to the setup wizard

Currently, Tezos Setup Wizard doesn't support archive nodes in any capacity, though #365 introduces support when starting a node from a snapshot. We would like to add full support for archive nodes, including starting them from scratch.

Additionally, when the user doesn't want to import a snapshot at all, the node is, by default, started in full history mode. It's possible rolling mode would suffice for the majority of users, as well as be consistent with the recommendations of the setup wizard itself.

It therefore  might be prudent to present an option to ask for history mode needed.

This is possible to do e.g. using the `tezos-node config update` subcommand, which takes possible config parameters as flags.

# Acceptance criteria

Tezos Setup Wizard allows to start a node in any history mode, both from scratch and when importing snapshots.

ixahedron:
## Description

Problem: currently, the only history mode support present in the Tezos Setup Wizard is the ability to reconstruct an archive mode when importing a full snapshot, and to import either a full or a rolling snapshot. This means we have no history mode support for syncing the node from scratch, and the default mode is 'full', which might not be reasonable.

Solution: added a new step in the wizard, asking about the preferred history mode in case the user skips snapshot import.

## Related issue(s)

Resolves #366 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -50,6 +50,12 @@ systemd_enable = {
     "no": "Start the services this time only",
 }
 
+history_modes = {
+    "rolling": "Store a minimal rolling window of chain data, lightest option",
+    "full": "Store enough chain data to reconstruct the complete chain state",
+    "archive": "Store all the chain data, very storage-demanding",
+}
+
 
 TMP_SNAPSHOT_LOCATION = "/tmp/tezos_node.snapshot"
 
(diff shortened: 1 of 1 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -578,6 +595,13 @@ def import_snapshot(self):
             snapshot_file = TMP_SNAPSHOT_LOCATION
 
             if self.config["snapshot"] == "skip":
```
rvem, line 597: I think it makes sense to ask for the desired history mode even if the user decided to download a snapshot or provide one stored locally. What do you think about providing the same choice when the user downloaded a full snapshot (well, except for the fact, that there are going to be only two options: full and archive)?

This way we may combine these changes and changes from the #365 and the history mode choice will be more consistent within various bootstrap options

rvem: Also, there is a possibility to run a rolling node using a full snapshot, which is probably not useful, but in some cases, some snapshot providers provide only full snapshots and thus it still may be useful

ixahedron: Ah, if it's possible to start a rolling node from a full snapshot, it definitely makes sense to me to ask about the mode when importing

docker/package/tezos_setup_wizard.py
```diff
@@ -50,6 +50,12 @@
     "no": "Start the services this time only",
 }
 
+history_modes = {
+    "rolling": "Store only minimal data, deleting everything before a rolling checkpoint",
+    "full": "Store chain data enough to reconstruct the complete chain state",
+    "archive": "Store everything, very storage-demanding",
```
pasqu4le, line 56: Small wording suggestion:
```suggestion
    "rolling": "Store a minimal rolling window of chain data, lightest option",
    "full": "Store enough chain data to reconstruct the complete chain state",
    "archive": "Store all the chain data, very storage-demanding",
```

docker/package/tezos_setup_wizard.py
pasqu4le: Related to this: what do you think about inverting the order of the "syncing" and "history mode" choices?

If we ask for the history mode first then in the syncing question we can only show:
- "download rolling" if the chosen history was rolling
- "download full" otherwise (and use the option from #365 if archive was picked)

in either case this first option would be the default one (followed by the common "file"/"url"/"skip" ones).

ixahedron: :thinking: I see a potential problem in not offering the `download full` snapshots option if a `rolling` history mode was chosen, related to the whole "snapshot providers only providing full snapshots", but I guess it's a smaller subset of users that are going to run into this compared to the potential UX improvement, so I agree

rvem (approved):
Looks good now!

Issue #349, opened 2021-12-06
Support custom networks in the baking wizards and services


rvem:
# Clarification and motivation
Sometimes we want to test some specific scenarios that don't happen on testnets, e.g. we want to test the protocol switchover or the interaction with the voting process (see #237).

However, it's possible to run custom networks in order to test such scenarios.
E.g., there is a script that provides interactive testing of the voting process: https://gitlab.com/morley-framework/local-chain/-/merge_requests/10.

At the moment it's impossible to run a baking service within the custom network using the node with the custom config file.

# Acceptance criteria
It should be possible to set up and run a baking service with a node that uses a custom config network.

Ideally, the setup wizard should ask for the config to use and the set of protocols for which daemons will be started.

rvem:
Resolved in `v12.0-1`

PR #369, opened 2022-01-14
[#349] Add custom chains baking services

rvem:
## Description

This PR adds a convenient way to start a baking service on a network with a custom chain which is defined by the custom node config file.

Testing changes locally:

1) Install updated `tezos-node` and `tezos-baking packages`
2) Clone [`local-chain` repo](https://gitlab.com/morley-framework/local-chain).
3) Create a config file for future custom baking instance:
```
sudo cp /etc/default/tezos-baking-custom@ /etc/default/tezos-baking-custom@voting
```
4) Start `./scenarios/voting.py --base-protocol hangzhou` from `local-chain`.
5) Fill `CUSTOM_NODE_CONFIG` in `/etc/default/tezos-baking-custom@voting` with the path provided by the voting scenario.
6) Start custom baking service:
```
sudo systemctl start tezos-baking-custom@voting
```
7) Check that all parts of it are actually running:
```
systemctl status tezos-node-custom@voting
systemctl status tezos-baker-011-pthangz2@custom@voting.service
systemctl status tezos-baker-012-psithaca@custom@voting.service
systemctl status tezos-endorser-011-pthangz2@custom@voting.service
```
8) Follow the voting scenario using the node from the `tezos-node-custom@voting`, it has `http://localhost:8732` endpoint by default.

9.1 Reset this custom baking service once the scenario is done:
```
sudo systemctl start tezos-baking-custom-reset@voting
```
9.2 Alternative way to reset custom baking service is to set `RESET_ON_STOP=true` env variable in config file and run
```
sudo systemctl stop tezos-baking-custom@voting
```

The only manual steps left are 4) and 5). I think they can be added to the `tezos-setup-wizard`.
Alternatively, we can consider custom baking services as an _advanced_ feature, leave these steps manual, and
document how one may start a custom baking service without the setup wizard. What do you think?

## Related issue(s)

Relates #349

  can be successfully built.


docker/package/model.py
```diff
@@ -530,6 +530,69 @@ def __init__(
...
+        custom_requires = []
+        for network in target_networks:
+            for proto in network_protos[network]:
+                custom_requires.append(f"tezos-baker-{proto.lower()}@custom@%i.service")
+                if proto not in self.noendorser_protos:
+                    custom_requires.append(
+                        f"tezos-endorser-{proto.lower()}@custom@%i.service"
+                    )
+        self.systemd_units.append(
+            SystemdUnit(
```
ixahedron, line 542: This `SystemdUnit` is in a lot of ways copy-paste of the previous non-custom ones. It might be nice to separate it into some function to only pass arguments that differ to it... But maybe you don't think it's worth it, I'll leave it to you to decide.

docker/package/model.py
```diff
@@ -530,6 +530,69 @@ def __init__(
...
+                service_file=ServiceFile(
+                    Unit(
+                        after=[],
+                        description=f"One-shot service to reset custom baking service",
+                    ),
+                    Service(
+                        exec_start="/usr/bin/tezos-baking-custom-reset %i",
+                        user="root",
+                        state_directory="tezos",
+                        environment_file=f"/etc/default/tezos-baking-custom@%i",
```
ixahedron, line 585: Does it make sense that its `environment_file` is the same as for the custom baking service itself?

ixahedron (commented):
4) and 5) could be added to the setup wizard, but I don't think they should. Custom baking services _do_ sound like an advanced feature (and a tutorial couldn't hurt regardless btw), and I don't think we should bloat he wizard with these choices that 1% of users is going to need.

docker/package/packages.py
```diff
@@ -213,12 +223,23 @@ def mk_node_unit(suffix, env, desc):
 node_units.append(
     mk_node_unit(
         suffix="custom",
-        env=["DATA_DIR=/var/lib/tezos/node-custom", "CUSTOM_NODE_CONFIG="]
+        env=["NODE_DATA_DIR=/var/lib/tezos/node-custom", "CUSTOM_NODE_CONFIG="]
```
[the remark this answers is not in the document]
rvem: Hmmm, seems like my commit splitting abilities aren't that great :joy:

docker/package/model.py
rvem: Hmm, seems like I initially wanted to use it in order to determine `NODE_DATA_DIR` from it, but at the moment `tezos-baking-custom-reset` is determining this env variable from `systemctl show`.
Actually, it's questionable what is better. On one hand `environment_file` here is indeed redundant, on the other hand, it's a more reliable way to determine env variable from this file rather than `grep`'ng `systemctl show` output.

What do you think?

docker/package/model.py
rvem: Yeah, they're quite similar indeed. The difference is `requires`, `description`, and `environment_file`. I agree that it's better to avoid copy-paste

docker/package/packages.py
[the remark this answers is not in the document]
rvem: I'll fix it during the next rebase

docker/package/model.py
ixahedron: Reliable is better 🙈 One redundant option doesn't matter that much in terms of performance nor readability, I think, so I'd use the file.

rvem: Oh, wait, I was wrong `NODE_DATA_DIR` isn't defined in this `environment_file`:(
So this option is indeed redundant here

rvem:
@ixahedron, may I ask you for a re-review?

ixahedron (approved):
LGTM, don't forget to move the thing to another commit ^

docker/package/packages.py
```diff
@@ -157,16 +157,25 @@
 """
```
pasqu4le, line 158: This commit (93b7817e7fb332451e8f1d3f7d8c5b4a71759b46) message has "instanted" in it, which seems a typo :slightly_smiling_face:

docker/package/packages.py
```diff
@@ -189,7 +189,7 @@ def mk_node_unit(suffix, env, desc):
 common_node_env = ["NODE_RPC_ADDR=127.0.0.1:8732", "CERT_PATH=", "KEY_PATH="]
 for network in networks:
     env = [
-        f"DATA_DIR=/var/lib/tezos/node-{network}",
+        f"NODE_DATA_DIR=/var/lib/tezos/node-{network}",
```
pasqu4le, line 202: > The only manual steps left are 4) and 5). I think they can be added to the tezos-setup-wizard.
Alternatively, we can consider custom baking services as an advanced feature, leave these steps manual, and
document how one may start a custom baking service without the setup wizard. What do you think?

I agree with @ixahedron that the wizard shouldn't have this option because it would complicate it without being useful for the vast majority of people.

However (as she mentioned as well) this PR should add a section in the baking tutorial describing this option.

pasqu4le (commented):
Overall the approach LGTM.

I haven't tested it yet tho, I'll finish my review back after I do.

docker/package/packages.py
pasqu4le: BTW I don't think that these instructions should rely on the voting scenario nor `local-chain` in general, given that there are other ways to run custom test networks (and that `local-chain` is mostly used only by us).

rvem: > this PR should add a section in the baking tutorial describing this option.

I propose to update the baking tutorial later once updated packages are published

docker/package/packages.py
rvem: Will update during the next rebase

docker/package/scripts/tezos-baking-custom-reset
```diff
@@ -0,0 +1,17 @@
...
+
+if [ -z "$1" ]; then
+    echo "Please call this script with the custom baking instance name."
+    exit 1
+fi
+
+instance_name="$1"
+systemctl stop tezos-baking-custom@"$instance_name"
+node_data_dir="$(systemctl show tezos-node-custom@"$instance_name".service | grep -o "NODE_DATA_DIR=\S*" | cut -d'=' -f2)"
+rm -rf "$node_data_dir"
```
pasqu4le, line 17: This is the only choice that seems a bit odd to me.
AFAIK the purpose of this service is mainly this step: to clean up the node data dir.

I'm not familiar with oneshot units, and it seems that cleanup is a common use-case for them, but isn't the call to `systtemctl stop` of another unit a bit odd?
Perhaps I'm missing some advantages, but if this service needs to be started manually, couldn't a script be used for this same purpose?

As an alternative, what do you think of performing this cleanup conditionally on the stop of the `tezos-baking-custom` service?
I suppose that this could be done by adding a variable in `tezos-baking-custom.conf` (that one would set when filling up the `CUSTOM_NODE_CONFIG`) to choose whether to perform a cleanup and this variable could be looked up in an `ExecStopPost` script (to do the removal of the node data dir or not).
What do you think?

rvem: > As an alternative, what do you think of performing this cleanup conditionally on the stop of the tezos-baking-custom service?

That sounds like a good option, I didn't think about it previously

docker/package/packages.py
rvem: > BTW I don't think that these instructions should rely on the voting scenario nor `local-chain` in general, given that there are other ways to run custom test networks (and that `local-chain` is mostly used only by us).

Sure, instructions in the PR description are based on `local-chain` only because this is the most convenient way to test changes for us

pasqu4le (approved):
Tested this out now and all LGTM :ok_hand:

PR #372, opened 2022-01-17
[Chore] Remove ithaca support from formulae

rvem:
## Description
Problem: v11.1 doesn't support ithaca. However, we have ithaca-related
formulae.

Solution: Remove ithaca-related formulae.

## Related issue(s)

Resolves None

  can be successfully built.


ixahedron (approved):
Oh, I actually deliberately left the formulae in the repo but removed them from the build script 😅 Why do you think they should be deleted, btw? Does it affect anything?

rvem:
> Does it affect anything?

One may try to install it from `tezos-packaging-stable` and get an error

PR #373, opened 2022-01-19
[#361] Rework bottles building CI

### LINKED ISSUE #361: Rework building bottles CI jobs for easier rebuilds

When building brew bottles for Catalina and Big Sur for the `v11.0+no_adx-2` release using GitHub Actions, two bottles [have](https://github.com/serokell/tezos-packaging/runs/4582612341?check_suite_focus=true#step:5:779) [failed](https://github.com/serokell/tezos-packaging/runs/4582612443?check_suite_focus=true#step:5:378) to build.

As we have no infrastructure to build for Catalina and `x86_64` Big Sur, we consequently have no way to manually add the missing pieces.

Additionally, as these jobs utilise [a script](../blob/master/scripts/build-bottles.sh) that builds the whole batch of bottles needed, there is no easy way to retry build failures for individual bottles using the CI. Re-running the whole job becomes necessary, which takes a long time.

@rvem suggested rewriting the CI to have each bottle be built in a separate step, at least when using GA.

# Acceptance criteria

It's easy to retry failed bottle builds.

ixahedron:
## Description

For a number of transient reasons, our bottle-building CIs sometimes fail to build some bottles. This isn't even reported, and even if we catch it, rebuilding is tedious, time-consuming, and doesn't guarantee it won't happen again.

This PR reworks both the two GitHub Actions runners' workflow and our Buildkite's tags pipeline to build each bottle in its own step/job, in a way that achieves both good error reporting and easy rebuilds.

### Buildkite
The tags pipeline is now generated semi-dynamically, though it's still possible to define new steps in `pipeline-for-tags.yml`, as usual. This saves us a ton of copy-paste and maintenance efforts since supporting new formulae for e.g. new protocols is as easy as before; only adding the name of the formula in one place in `generate-tags-pipeline` is needed.
While every re-run of the pipeline will still build all the bottles, Buildkite allows to rerun single steps, so this shouldn't cause problems.
Setting this up will require changes in Buildkite's settings.

### GitHub Actions
Before building the bottle, GA should now check the relevant release to confirm the bottle hasn't been built yet, and skip the rest of the job if it has. As GA only supports rerunning whole workflows instead of jobs, this allows us to quickly skip unnecessary work steps.

## Related issue(s)

Resolves #361 

  can be successfully built.


.buildkite/generate-tags-pipeline.sh
```diff
@@ -0,0 +1,75 @@
+#! /usr/bin/env nix-shell
```
rvem, line 1: Don't forget to update [steps in CI configuration](https://buildkite.com/serokell/tezos-packaging-tags/steps) in order to include the run of this script before pipeline configuration uploading

.buildkite/generate-tags-pipeline.sh
```diff
@@ -0,0 +1,75 @@
...
+     queue: \"arm64-darwin\"
+   if: build.tag =~ /^v.*/
+   commands:
+   - brew install --formula ./Formula/tezos-sapling-params.rb"
+
+n=0
+for f in "${formulae[@]}"; do
+  n=$((n+1))
+  ymlappend "
+ - label: Build $f bottle for Big Sur arm64
```
rvem, line 29:
```suggestion
 - label: Build $f bottle for Big Sur arm64
```
```suggestion
 - label: Build $f bottle for Big Sur arm64
```

.buildkite/generate-tags-pipeline.sh
```diff
@@ -0,0 +1,75 @@
...
+  key: build-bottle-$n
```
rvem, line 30: My internal perfectionist is upset due to lack of space here and below :see_no_evil:  
```suggestion
   key: build-bottle-$n
```

.buildkite/generate-tags-pipeline.sh
```diff
@@ -0,0 +1,75 @@
...
+  agents:
+    queue: \"arm64-darwin\"
+  if: build.tag =~ /^v.*/
+  commands:
+  - brew install --formula --build-bottle \"./Formula/$f.rb\"
```
rvem, line 35: AFAIU, `build-bottles.sh` is no longer used. What do you think about moving the commands from this step to a script, so that we'll be able to avoid copy-paste here and in the GA setup?

rvem:
Let's mention somewhere in the `pipeline-for-tags.yml` that part of this pipeline is generated by the script

.github/workflows/build-bottles.yml
```diff
@@ -9,41 +9,72 @@ permissions:
...
     strategy:
+      fail-fast: false
       matrix:
-        include:
-          - os: macos-10.15
-            name: "Catalina"
-          - os: macos-11
-            name: "Big Sur"
+        os: [ { id: macos-10.15, name: "catalina" }, { id: macos-11, name: "big_sur" } ]
+        formula: [tezos-accuser-011-PtHangz2, tezos-admin-client, tezos-baker-011-PtHangz2, tezos-client, tezos-codec, tezos-endorser-011-PtHangz2, tezos-node, tezos-sandbox, tezos-signer]
```
rvem, line 18: I slightly dislike the fact that list of these formulae is listed in two places :disappointed:

Annoyingly, YAML doesn't support any kind of "includes" of other files

.github/workflows/build-bottles.yml
```diff
@@ -9,41 +9,72 @@ permissions:
...
 
     steps:
+      - id: check-built
+        name: Check if the bottle has already been built
+        continue-on-error: true
+        run: gh release view "${GITHUB_REF#refs/tags/}" | grep "${{ matrix.formula }}.*${{ matrix.os.name }}.bottle.tar.gz"
```
rvem, line 24: :pogchamp:

rvem:
Looks quite good. It turned out to be easier than I initially thought :slightly_smiling_face:

scripts/build-one-bottle.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ -z "$1" ] ; then
+    echo "Please call this script with the name of the binary for which to build the bottle."
+    exit 1
+fi
+
+brew install --formula --build-bottle "./Formula/$1.rb"
+brew bottle --force-core-tap --no-rebuild "./Formula/$1.rb"
+brew uninstall "./Formula/$1.rb"
+# https://github.com/Homebrew/brew/pull/4612#commitcomment-29995084
+mv "$1*.bottle.*" "$(echo $1*.bottle.* | sed s/--/-/)"
```
rvem, line 20: This IMO looks a bit more compact and a bit less scary :pled:
```suggestion
mv "$1*.bottle.*" "${1/--/-}"
```

scripts/sync-bottle-hashes.sh
```diff
@@ -30,7 +30,7 @@ fi
 while : ; do
     git fetch --all
     git reset --hard origin/"$branch_name"
-    ./scripts/bottle-hashes.sh .
+    ./scripts/bottle-hashes.sh "./$2"
```
rvem, line 33: Could you explain why this was changed? Seems like it expects bottles to be in `<macOS name>` directories, but I can't find whether bottles are actually moved to these directories

rvem (approved):
LGTM now, have only one minor nitpick

PR #377, opened 2022-01-25
[Chore] Add bottle hashes for v12.0-rc2-1

ixahedron:
Problem: we have built brew bottles for the new Octez release, but their hashes
aren't in the formulae yet.

Solution: added the hashes.


rvem:
From the buildkite docs:
> You can also change working directory to the intended destination and use a <destination> of '.' to always create a directory hierarchy matching the artifact paths

So, apparently, it only automatically creates directories after the <destination>

rvem (approved):

Issue #379, opened 2022-01-28
[BUG] tezos-setup-wizard offers wrong list of addresses when two Ledger devices are connected


itkach:
# Description

When setting up baking with ledger and 2 ledger devices are plugged in, only one device is listed, but the list of 4 available addresses is duplicated:

```
How do you want to import the baker key?
1. ledger (default)            From a ledger
2. secret-key                  Either the unencrypted or password-encrypted secret key for
                               your address
3. json                        Faucet JSON file from https://faucet.tzalpha.net/
> 1
Please open the Tezos Baking app on your ledger.
Select a key to import from ledger://crouching-tiger-hidden-dragon/.
You can choose one of the suggested derivations or provide your own:
1. bip25519/0h/0h,  address: tz1...A, balance: 0 ꜩ
2. ed25519/0h/0h,   address: tz1...B, balance: 0 ꜩ
3. secp256k1/0h/0h, address: tz2...C, balance: 0 ꜩ
4. P-256/0h/0h,     address: tz3...D, balance: 0 ꜩ
5. bip25519/0h/0h,  address: tz1...A, balance: 0 ꜩ
6. ed25519/0h/0h,   address: tz1...B, balance: 0 ꜩ
7. secp256k1/0h/0h, address: tz2...C, balance: 0 ꜩ
8. P-256/0h/0h,     address: tz3...D, balance: 0 ꜩ
9. Specify derivation path
10. Go back
```

# To Reproduce

- Plug-in 2 ledger devices and start tezos baking app on both
- Follow tezos-setup-wizard
- In the import key step, choose ledger, observe second ledger is not listed, addresses from the first device are duplicated 

# Expected behavior

Both devices are listed, with some addresses from each, like so:
```
You can choose one of the suggested derivations or provide your own.

From ledger://crouching-tiger-hidden-dragon/:

1. bip25519/0h/0h,  address: tz1...A, balance: 0 ꜩ
2. ed25519/0h/0h,   address: tz1...B, balance: 0 ꜩ
3. secp256k1/0h/0h, address: tz2...C, balance: 0 ꜩ
4. P-256/0h/0h,     address: tz3...D, balance: 0 ꜩ
5. Specify derivation path

From: ledger://flat-crocodile-toy-cheburashka:

6. bip25519/0h/0h,  address: tz1...E, balance: 0 ꜩ
7. ed25519/0h/0h,   address: tz1...F, balance: 0 ꜩ
[...]

PR #380, opened 2022-01-28
[#379] Support multiple ledgers connection in setup wizard

rvem:
Problem: Setup shows incorrect data when multiple ledgers are connected
and have the Baking app open.

Solution: Properly support a case when multiple ledgers are connected.
Explicitly show the ledger URL in the list of derivations. Additionally,
allow choosing a ledger when specifying additional derivations.

## Related issue(s)

Relates #379 (the issue will be resolved once updated `tezos-baking` package is published)

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -225,12 +225,16 @@ def wait_for_ledger_baking_app():
             f"sudo -u tezos {suppress_warning_text} tezos-client list connected ledgers"
         ).stdout
         proc_call("sleep 1")
-    base_ledger_url = (
-        re.search(b"ledger:\/\/[\w\-]+\/", output).group(0).decode("utf-8")
-    )
-    return base_ledger_url, [
-        name.decode() for name in re.findall(derivation_path_regex, output)
-    ]
+    ledgers_derivations = {}
+    for ledger_derivation in re.findall(ledger_regex, output):
+        ledger_url = (
+            re.search(b"ledger:\/\/[\w\-]+\/", ledger_derivation).group(0).decode()
+        )
+        derivation_path = (
+            re.search(derivation_path_regex, ledger_derivation).group(0).decode()
+        )
+        ledgers_derivations.setdefault(ledger_url, []).append(derivation_path)
+    return ledgers_derivations
 
 
 def get_data_dir(network):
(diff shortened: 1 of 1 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -815,8 +824,7 @@ def import_baker_key(self):
                         ):
                             self.query_step(
                                 get_ledger_derivation(
-                                    ledger_derivations,
-                                    base_ledger_url,
+                                    ledgers_derivations,
```
ixahedron, line 827: I'm not sure how this works. IIUC, it will still only show all the derivation paths together like right now, except without specifying the accompanying base ledger url. So there is a bit less confusion, but it's still difficult to gather which ledger we are chooding. Can we not ask to choose the ledger url (if there is more than one ledger) in any case before this step and show only the relevant ones?
Maybe I missed something in the workflow, in that case I'd like to see some screenshots as I can't test this myself 😞

docker/package/tezos_setup_wizard.py
```diff
@@ -225,12 +225,10 @@ def wait_for_ledger_baking_app():
...
         proc_call("sleep 1")
-    base_ledger_url = (
-        re.search(b"ledger:\/\/[\w\-]+\/", output).group(0).decode("utf-8")
+    ledgers = list(
+        {url.decode() for url in re.findall(b"ledger:\/\/[\w\-]+\/", output)}
     )
-    return base_ledger_url, [
-        name.decode() for name in re.findall(derivation_path_regex, output)
-    ]
+    return ledgers, [name.decode() for name in re.findall(ledger_regex, output)]
```
ixahedron, line 231: Ahh, I think, if I got what you wanted to do in `ledger_urls_info` right (see my previous comment), you missed gluing every derivation together with its ledger address here, otherwise `ledgers_derivations` contains the same incomplete paths as `ledger_derivation` before, i.e. the ones you had to combine with `base_ledger_url` we no longer have.
🤔

docker/package/tezos_setup_wizard.py
```diff
@@ -225,12 +225,10 @@ def wait_for_ledger_baking_app():
             f"sudo -u tezos {suppress_warning_text} tezos-client list connected ledgers"
         ).stdout
         proc_call("sleep 1")
-    base_ledger_url = (
-        re.search(b"ledger:\/\/[\w\-]+\/", output).group(0).decode("utf-8")
+    ledgers = list(
+        {url.decode() for url in re.findall(b"ledger:\/\/[\w\-]+\/", output)}
     )
```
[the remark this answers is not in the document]
rvem: > can there be repetitions?

Yes, it matches every `ledger://<bla-bla>/bip25519/0h/0h` from the `tezos-client` output

docker/package/tezos_setup_wizard.py
rvem: So, currently, it shows 4 derivations for each ledger:
image
When you choose `Specify derivation path`, it asks for a specific ledger URL before querying the derivation path on it:
image

ixahedron: Ah, OK, I see my mistake :facepalm: Not sure how readable this is though, especially with the lack of column alignment :thinking: I still think it might be better to choose the ledger beforehand, but idk

rvem: I agree, that it isn't very readable :unamused:

Perhaps, we can store a `dict` instead of a list and display derivations for each ledger separately :thinking:

rvem: Unfortunately, I don't have access to the second ledger at the moment, but this is how it looks when we show derivations without showing the full ledger path:
image

pasqu4le: Dividing the options per-ledger doesn't look bad to me.

However, I see that `Specify derivation path` is given separately, which makes me think there will be only one regardless of the number of ledgers (as in the screenshot above as well), but shouldn't it also be given per-ledger instead?

rvem: `Specify derivation path` lets you specify the ledger. See the screenshot in my previous comment

pasqu4le: oh sorry, I completely missed that, never mind my previous comment :ok_hand:

rvem: I've added proper alignment for cases when there are more than 10 options. Now it looks like this:
image

rvem: @ixahedron, @pasqu4le, may I ask you to re-review now?

docker/package/tezos_setup_wizard.py
```diff
@@ -324,15 +328,42 @@ def pprint_options(self):
             pass
 
         if self.options and isinstance(self.options, list):
-            str_format = "{:1}. {}"
+            options_count = 0
             for o in self.options:
-                if def_i is not None and i == def_i:
-                    print(str_format.format(i, "(default) " + o))
+                if isinstance(o, dict):
+                    for k, values in o.items():
```
ixahedron, line 334: I don't see `k` used anywhere, why not
```suggestion
                    for values in o.values():
```

ixahedron (approved):
One nitpick, but LGTM otherwise

PR #382, opened 2022-02-01
[#375] Document 'tezos-packaging' usage on WSL

### LINKED ISSUE #375: Consider adding Windows support through WSL

At the moment Windows is not an OS supported by Tezos, from [its documentation](http://tezos.gitlab.io/introduction/howtoget.html#environment):
> Currently Tezos is being developed for Linux x86_64, mostly for Debian/Ubuntu and Arch Linux. The following OSes are also reported to work: macOS (x86_64), Arch Linux ARM (aarch64), Debian Linux (buster), Ubuntu Linux (focal). A Windows port is feasible and might be developed in the future.

However, it might be possible to run a Tezos setup with [WSL](https://docs.microsoft.com/en-us/windows/wsl/), most likely using [Ubuntu on WSL](https://ubuntu.com/wsl), since Ubuntu is arguably the primary OS target of `tezos-packaging`.

We should check for this possibility and test that everything works correctly in such a setup.

# Acceptance criteria

1. existing packages are tested on WSL and verified to work (including with Ledger devices)
2. if necessary, changes are made to add support for Windows on a WSL setup
3. if successful, the documentation is updated to reflect that and give installation instructions for Windows

DMozhevitin:
## Description

Problem: During testing it was found that `tezos-packaging`
can be used on WSL, but its usage has some limitations.

Solution: Information about WSL support was added to support
policy doc. Limitations are also mentioned in the corresponding
sections.

## Related issue(s)

Resolves #375

  can be successfully built.

### DIFF
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -55,6 +55,13 @@ Apart from the `.service` file you'll need the service startup script and defaul
 configuration file, they can be found in the [`scripts`](../docker/package/scripts) and
 [`defaults`](../docker/package/defaults) folders respectively.
 
+## Systemd units on WSL
+
+Unfortunately, `systemd` is not officially supported on WSL.
+
+However, there are several unofficial workarounds for it and some have been known to work with `tezos-packaging`'s units.
+
+If you are successfully running a distro on WSL with `systemd`, the documentation above should apply to you too.
 
 ## Multiple similar systemd services
 
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ the binaries, and the services used.
 - [**Raspberry Pi OS**](./docs/distros/ubuntu.md#raspberry)
 - [**Fedora**](./docs/distros/fedora.md)
 - [**macOS**](./docs/distros/macos.md)
+- [**Windows using WSL**](./docs/distros/ubuntu.md#wsl)
 
 The information about supported versions of the aforementioned OSes is available in the [support policy doc](./docs/support-policy.md).
 
--- a/docker/package/scripts/udev-rules
+++ b/docker/package/scripts/udev-rules
@@ -6,11 +6,10 @@
 # that are fixing https://github.com/LedgerHQ/udev-rules/issues/5, so that provided rules work on the Raspberry Pi OS
 # Ubuntu 18.04
 
-# Don't add udev rules in case the package is installed inside either docker or podman container
-# or WSL (since it doesn't fully support udev and libusb at the moment).
+# Don't add udev rules in case the package is installed inside either docker or podman container.
 # Otherwise, post-installation script will fail due to inability to non-zero exit code of the
 # 'udevadm control --reload-rules' call
-if [ ! -f /.dockerenv ] && [ ! -f /.containerenv ] && ! grep -qEi "(Microsoft|WSL)" /proc/sys/kernel/osrelease ; then
+if [ ! -f /.dockerenv ] && [ ! -f /.containerenv ] ; then
     cat <<EOF > /etc/udev/rules.d/20-hw1.rules
 # HW.1 / Nano
 SUBSYSTEMS=="usb", ATTRS{idVendor}=="2581", ATTRS{idProduct}=="1b7c|2b7c|3b7c|4b7c", TAG+="uaccess", TAG+="udev-acl", MODE="0660", GROUP="plugdev"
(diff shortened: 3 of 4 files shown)


docs/support-policy.md
```diff
@@ -6,7 +6,7 @@
 # Packages support policy
 
 `tezos-packaging` provides packages for Ubuntu, Fedora and macOS for each stable and release-candidate
-version of Octez.
+version of Octez. `Ubuntu` packages can also be used on WSL, but their support is limited. Functionality limitations are described in the corresponding sections.
```
pasqu4le, line 9: :thinking: I'm not sure that we should mention this here.

Overall I think that we should treat Windows support like we do Raspberry Pi OS, we don't offer specific native packages for them, but Ubuntu ones can be used (with some limitations).

So, I suggest reverting this change and instead adding:
- a "Ubuntu packages on WSL" section in the [Ubuntu installation docs](https://github.com/serokell/tezos-packaging/blob/master/docs/distros/ubuntu.md)
- a "Windows using WSL" link to that section in the [Installing tezos](https://github.com/serokell/tezos-packaging#installing-tezos)

As for what to put there, well, I think it should point at the [Ubuntu WSL](https://ubuntu.com/wsl) docs and then list how to perform the installation with that being set up (once again, much like we do for Raspberry Pi OS).

docs/systemd.md
```diff
@@ -55,6 +55,9 @@ Apart from the `.service` file you'll need the service startup script and defaul
 configuration file, they can be found in the [`scripts`](../docker/package/scripts) and
 [`defaults`](../docker/package/defaults) folders respectively.
 
+## Systemd services on WSL
```
pasqu4le, line 58: Nitpick, but to match the previous 2 sections I'd rename this to:
```suggestion
## Systemd units on WSL
```

docs/systemd.md
```diff
@@ -55,6 +55,9 @@ Apart from the `.service` file you'll need the service startup script and defaul
...
+
+Unfortunately, `systemd` is not officially supported on WSL. However, there are some unofficial workarounds for it.
```
pasqu4le, line 60: This is in the right place and says the right thing IMO :slightly_smiling_face: 

I would only add that:
```suggestion
Unfortunately, `systemd` is not officially supported on WSL.

However, there are several unofficial workarounds for it and some have been known to work with `tezos-packaging`'s units.
If you are successfully running a distro on WSL with `systemd`, the documentation above should apply to you too.
```

docs/baking.md
```diff
@@ -192,6 +192,10 @@ sudo -u tezos tezos-client register key baker as delegate
 Check a blockchain explorer (e.g. https://tzkt.io/ or https://tzstats.com/) to see the baker status and
 baking rights of your account.
 
+### Using ledger on WSL
+
+In order to use ledger on WSL some additional preparation steps are needed. We recommed to use `usbipd-win` tool and follow [this guide](https://docs.microsoft.com/en-us/windows/wsl/connect-usb) about connecting usb devices to WSL.
+
```
pasqu4le, line 198: This seems like an odd place for this info to me because:
1. it comes after the commands that would have used the ledger itself
2. this guide is the one based on `systemd` which as we now know isn't supported officially, so many people won't look for this info here

I believe it would be better place inside the (missing) section that I mentioned two comments above instead.

DMozhevitin: Yeah, I agree that this place indeed looks odd.

> I believe it would be better place inside the (missing) section that I mentioned two comments above instead.

Putting this comment in that section would indeed be better, but probably still not the most appropriate, because it contains information about installing packages, rather than using them :thinking:

DMozhevitin:
@pasqu4le thanks for your comments, now the docs look more structured, but may still contain some inaccuracies.

Could you please take a look when you have time?

docs/baking.md
pasqu4le: well, most of the information in that Ubuntu document is in fact about installation, so I think this fits there too.

Admittedly this explains both what to install and where to find docs on how to use that, but I think that's fine (and there is no better place for that).

pasqu4le:
@DMozhevitin all LGTM now except for one thing: you squashed the new changes with the last commit, but they should have been merged with the first one instead, please fix that :slightly_smiling_face:

pasqu4le (approved):
Only one fixup suggestion for `342659c5b12aba21bf97d51cd48a68c2a7c82f43`, but LGTM.

I'm approving already, feel free to merge after reverting that change :+1:

PR #383, opened 2022-02-02
[Chore] Fix faucet link

DMozhevitin:
Problem: There are several occurencies of https://faucet.tzalpha.net/,
which is deprecated since hangzhounet and doesn't work now.

Solution: Replace them with https://teztnets.xyz/.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -42,7 +42,7 @@ snapshot_import_modes = {
 key_import_modes = {
     "ledger": "From a ledger",
     "secret-key": "Either the unencrypted or password-encrypted secret key for your address",
-    "json": "Faucet JSON file from https://faucet.tzalpha.net/",
+    "json": "Faucet JSON file from https://teztnets.xyz/",
 }
 
 systemd_enable = {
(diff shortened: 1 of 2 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -42,7 +42,7 @@
 key_import_modes = {
     "ledger": "From a ledger",
     "secret-key": "Either the unencrypted or password-encrypted secret key for your address",
-    "json": "Faucet JSON file from https://faucet.tzalpha.net/",
+    "json": "Faucet JSON file from https://teztnets.xyz/hangzhounet-faucet/",
```
rvem, line 45: Let's reference https://teztnets.xyz in general without specifying the exact testnet since it's going to change over time

DMozhevitin: Fair note

pasqu4le (approved):

rvem (approved):

docker/package/tezos_setup_wizard.py
pasqu4le: @DMozhevitin I see you changed this in the code but not in the commit message, please update that too

ixahedron (approved):

PR #392, opened 2022-02-25
[#390] Update support policy doc

### LINKED ISSUE #390: Add back full support for Fedora 35 when possible

As mentioned in the [documentation](https://github.com/serokell/tezos-packaging/blob/f812abc07188c8ad76a600f3be63848bd2ccbcb9/docs/support-policy.md#fedora-packages) support for `Fedora 35` is limited due to a compilation issue with recent versions of `gcc` and `glibc`.

Even tho the linked issue seems to still be open ([tezos/#1449](https://gitlab.com/tezos/tezos/-/issues/1449)) in the announcement and [changelog for v12.0](https://tezos.gitlab.io/CHANGES.html#version-12-0) it is mentioned that:
> Updated the build system to use a patched version of the OCaml compiler to fix a bug that could cause compilation to fail on newer versions of gcc and glibc.

We should check if it is now possible build on Fedora 35 (and what changes are needed, if any) and update the documentation accordingly.

# Acceptance criteria

1. Octez version `v12.0` is build on Fedora 35 for both supported architecture, unless it's still not possible due to new reasons.
2. The documentation in the repo (in particular `docs/support-policy.md`) is updated accordingly.

rvem:
## Description

Update list of supported OSes versions:

1) Remove Ubuntu 21.04 since it's obsolete now.
2) Mention that there are both `x86_64` and `aarch64` packages for Fedora 35 now.

## Related issue(s)

Resolves #390

  can be successfully built.

### DIFF
--- a/docker/package/package_generator.py
+++ b/docker/package/package_generator.py
@@ -80,7 +80,6 @@ common_deps = run_deps + build_deps
 ubuntu_versions = [
     "bionic",  # 18.04
     "focal",  # 20.04
-    "hirsute",  # 21.04
 ]
 
 pwd = os.getcwd()
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -239,7 +239,7 @@ Source: {self.name.lower()}
 Section: utils
 Priority: optional
 Maintainer: {self.meta.maintainer}
-Build-Depends: debhelper (>=9), {"dh-systemd (>= 1.5), " if ubuntu_version != "hirsute" else ""} autotools-dev, {str_build_deps}
+Build-Depends: debhelper (>=9), dh-systemd (>= 1.5), autotools-dev, {str_build_deps}
 Standards-Version: 3.9.6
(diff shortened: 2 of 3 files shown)


docker/package/package_generator.py
```diff
@@ -80,7 +80,6 @@
 ubuntu_versions = [
     "bionic",  # 18.04
     "focal",  # 20.04
-    "hirsute",  # 21.04
```
pasqu4le, line 83: I see that there's also some special treatment of `hirsute` in `docker/package/model.py`, wouldn't it be better to remove that too?

pasqu4le (approved):
One comment, but otherwise LGTM

docker/package/package_generator.py
pasqu4le: @rvem I see you've fixed one there, but there's two more similar lines in that file, please remove those too :slightly_smiling_face:

rvem: Updated

PR #393, opened 2022-02-25
[#389] Document the voting wizard

### LINKED ISSUE #389: Document the existence of voting wizard
In #347 a voting wizard was added to the `tezos-baking` package. However, at the moment documentation doesn't provide any
information about this wizard. We should document how and when users can use it.

Note that this document should be merged only once a new `tezos-baking` version with the voting wizard is published (which is going to be `v12.0-1`.

# Acceptance criteria
There is a document that describes how to use voting wizard.

ixahedron:
## Description

Problem: we now have released Tezos Voting Wizard, but there's no documentation
on how to use it.

Solution: added a doc on voting as well as documented how to set up a
custom baking instance and use it for voting.

## Related issue(s)

Resolves #389 

  can be successfully built.

### DIFF
--- a/docs/baking.md
+++ b/docs/baking.md
@@ -247,6 +247,42 @@ In case you want to use a different alias for the baking account:
 2. update the `BAKER_ADDRESS_ALIAS` by editing the
     `/etc/default/tezos-baking-<network>` file.
 
+### Using a custom chain
+
+In case you want to set up a baking instance on a custom chain instead of relying on mainnet
+or official testnets, you can do so:
+
+1. Create a config file for future custom baking instance:
+  ```bash
+  sudo cp /etc/default/tezos-baking-custom@ /etc/default/tezos-baking-custom@<chain-name>
+  ```
+2. In `/etc/default/tezos-baking-custom@<chain-name>`, provide the path to your custom node
+  config in `CUSTOM_NODE_CONFIG`.
+3. Start custom baking service:
+  ```bash
+  sudo systemctl start tezos-baking-custom@<chain-name>
+  ```
+4. Check that all parts are indeed running:
+  ```bash
+  systemctl status tezos-node-custom@<chain-name>
+  systemctl status tezos-baker-011-pthangz2@custom@<chain-name>.service
+  systemctl status tezos-baker-012-psithaca@custom@<chain-name>.service
+  systemctl status tezos-endorser-011-pthangz2@custom@<chain-name>.service
+```
+
+If at any point after that you want to reset the custom baking service, you can set
+`RESET_ON_STOP=true` in the `/etc/default/tezos-baking-custom@<chain-name>` config file and run:
+
+```bash
+sudo systemctl stop tezos-baking-custom@voting
+```
+
+Manually resetting is possible through:
+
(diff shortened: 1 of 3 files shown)


docs/baking.md
```diff
@@ -247,6 +247,42 @@ In case you want to use a different alias for the baking account:
...
+  systemctl status tezos-node-custom@<chain-name>
+  systemctl status tezos-baker-011-pthangz2@custom@<chain-name>.service
+  systemctl status tezos-baker-012-psithaca@custom@<chain-name>.service
+  systemctl status tezos-endorser-011-pthangz2@custom@<chain-name>.service
+```
+
+If at any point after that you want to reset the custom baking service, you can run:
+
+```bash
+sudo systemctl start tezos-baking-custom-reset@<chain-name>
```
rvem, line 276: Hmm, such service doesn't actually exist anymore. It was removed prior to `RESET_ON_STOP` option in `/etc/default/tezos-baking-custom@<chain-name>`

docs/baking.md
```diff
@@ -247,6 +247,42 @@ In case you want to use a different alias for the baking account:
...
+2. In `/etc/default/tezos-baking-custom@<chain-name>`, provide the path to your custom node
+  config in `CUSTOM_NODE_CONFIG`.
+3. Start custom baking service:
+  ```bash
+  sudo systemctl start tezos-baking-custom@<chain-name>
+  ```
+4. Check that all parts are indeed running:
+  ```bash
+  systemctl status tezos-node-custom@<chain-name>
+  systemctl status tezos-baker-011-pthangz2@custom@<chain-name>.service
```
rvem, line 268: This set of services is likely going to change over time.
However, mentioning abstract `<proto>` instead of the specific protocol is probably a bad idea because 1) custom baking service runs daemons for multiple protocols, and the set of protocols changes over time 2) since ithaca there is no endorser daemon, so users may be confused by lack of it.

Perhaps, it's worth leaving this as is, but we should keep in mind that this piece of doc should be updated each time we add/drop protocol support.

docs/baking.md
rvem: Seems like it's sufficient to delete the following:
1) Data from the node data directory.
2) `blocks`, `nonces`, `endorsements` from `tezos-client` data directory (which is `/var/lib/tezos/.tezos-client` by default).

Since this is already advanced section, I think it's fine to leave this steps to be done manually

docs/voting.md
```diff
@@ -0,0 +1,40 @@
...
+
+`tezos-packaging` provides an easy way to propose amendments and cast votes on the Tezos blockchain.
+
+This article documents the way to vote on Ubuntu, as well as any prerequisite steps necessary.
+
+## Using the wizard
+
+You can easily use Tezos Voting Wizard to vote on mainnet.
+
+A prerequisite for this is a mainnet baking instance already set up. An easy way to configure
```
rvem, line 17:
```suggestion
A prerequisite for this is a mainnet baking instance already set up and running. An easy way to configure
```
AFAIR, it uses baking instance node RPC address, to it needs to be up

rvem (approved):

PR #406, opened 2022-03-17
[Chore] Update copyright holders and licensing info

pasqu4le:
## Description

Copyright holders and licensing info need to be updated and this MR does that.

See commit messages and related issue for more details.

_For reference on how the header replacements were made see [this `morley` comment](https://gitlab.com/morley-framework/morley/-/merge_requests/1085#note_876214999)._

## Related issue(s)

[TM-592](https://issues.serokell.io/issue/TM-592)

  can be successfully built.


rvem (approved):

ixahedron (commented):
Obvs it doesn't matter for the linter, but I kinda dislike that the headers now are not in line with `reuse`'s [standard header template](https://reuse.readthedocs.io/en/stable/usage.html#templates), meaning there's no empty commented line between copyright and license lines. It's not a big deal, but it's not a necessary change, either 🤔

ixahedron (approved):
Well to be fair, that we're inconsistent in other repos should mean we need to bring those inconsistencies in line with the default template, not disrupt things further :see_no_evil: But OK, the time needed for another pass now would be more than a small aesthetic change like this warrants.

PR #411, opened 2022-03-22
[Chore] Use tweag/opam-nix for nix-based builds

rvem:
## Description
Problem: Currently we're using not really production-ready opam-nix
for nix-based builds. It's quite inconvenient to use because it requires
manual version adjustments for opam packages. Recently tweag released
their own version of machinery that can be used for building opam-based
projects via nix.

Solution: Adjust code to use tweag/opam-nix instead of our own opam-nix.

## Related issue(s)

https://issues.serokell.io/issue/TM-593

  can be successfully built.


nix/build/zcash-overlay.nix
```diff
@@ -0,0 +1,23 @@
...
+    nativeBuildInputs = oa.nativeBuildInputs ++ [ makeWrapper ];
+    postFixup = ''
+      for bin in $(find $out/bin -not -name '*.sh' -type f -executable); do
+        wrapProgram "$bin" --prefix XDG_DATA_DIRS : ${zcash}
+      done
+    '';
+  };
+
+  relevant-packages =
+    "tezos-(accuser|node|codec|admin-client|baker|endorser|signer).*";
```
ixahedron, line 18: Is there a reason not all packages that previously seemed to use `zcash` (e.g. `tezos-client`) are listed here?

ixahedron (commented):
This looks amazing to me :2022: :rocket: , I only have one question.

nix/build/zcash-overlay.nix
rvem: oh, good catch, indeed, seems like `tezos-client` is missing

ixahedron (approved):

nix/build/ocaml-overlay.nix
```diff
@@ -1,16 +1,23 @@
...
 , protocols ? builtins.fromJSON (builtins.readFile ../../protocols.json)
-, hacks ? import ./hacks.nix
-, pkgs ? import sources.nixpkgs { }, opam-nix ? import sources.opam-nix pkgs }:
-self: super: {
-  ocamlPackages = self.ocaml-ng.ocamlPackages_4_12.overrideScope'
-    (builtins.foldl' self.lib.composeExtensions (_: _: { }) [
-      (opam-nix.traverseOPAMRepo' sources.opam-repository)
-      (oself: osuper: { index-super = osuper.index.versions."1.2.0"; })
-      (opam-nix.callOPAMPackage sources.tezos)
+, hacks ? import ./hacks.nix, zcash-overlay ? import ./zcash-overlay.nix
```
pasqu4le, line 6: Maybe I'm missing something, but why is `hacks` still needed? :thinking:

rvem: There are a couple of Octez-specific workarounds needed e.g. for `tezos-admin-client` binary (since it's the part of the `tezos-client` opam package)

pasqu4le (approved):

PR #412, opened 2022-03-22
[#409] Use formulae tags for attaching bottles to releases

### LINKED ISSUE #409: Pick GitHub release from formula's version in automated bottle builds

Currently, when a new "version tag" is pushed, several `brew` bottle-building CI steps are launched that (among other things) will each:
1. check that the bottle for a formula doesn't already exist in its GitHub release (avoiding unnecessary rebuilds)
2. after building a new bottle, upload it to its GitHub release

However, in both case "its GitHub release" is assumed to be same same as the tag that started the CI pipeline, rather than the one specified in the formula's `version` field.

This works alright for new upstream versions, where we want to update the `version` for ~all formulae and build bottles for them, but not so much for our release versions (that are using an existing Octez version).

After some releases-only changes, in fact, we might want to push a new version that either only affects _some formulae_ (or none at all).

# Acceptance criteria

CI steps that build bottles will use the `version` in their respective formula to look for, and upload to, the corresponding GitHub release page.

ixahedron:
Problem: currently, both on GitHub Actions and with Buildkite, we build brew bottles on a new tag being pushed. In these runs, the bottle-building CI steps that have to do with:

1. checking that the bottle has already been built;
2. attaching built bottles to releases;
3. making commits that add new hashes to the formulae;

use the tag that triggered the pipeline. Turns out this behaviour doesn't work for any subsequent releases we make in the same upstream release, because the bottle versions usually stay the same, leading e.g. to the `v.*-2` release having `v.*-1` bottles attached.

Solution: changed the scripts to use the version of a specific formula where appropriate, or the one in `tezos-client.rb` where there is no specific bottle to be handled, as `tezos-client` is probably going to use the latest version.
In addition, changed BK steps to also check if a bottle has already been built.

## Related issue(s)

Resolves #409

  can be successfully built.

### DIFF
--- a/.buildkite/generate-tags-pipeline.sh
+++ b/.buildkite/generate-tags-pipeline.sh
@@ -32,13 +32,26 @@ n=0
 for f in "${formulae[@]}"; do
   n=$((n+1))
   ymlappend "
+
+ - label: Check if $f bottle for Big Sur arm64 is already built
+   key: check-built-$n
+   if: build.tag =~ /^v.*/
+   soft_fail:
+   - exit_status: 3 # We don't want the pipeline to fail if the bottle's already built
+   commands:
+   - nix-shell ./scripts/shell.nix
+       --run './scripts/check-bottle-built.sh \"$f\" \"arm64_big_sur\"'
+
  - label: Build $f bottle for Big Sur arm64
    key: build-bottle-$n
    agents:
      queue: \"arm64-darwin\"
    if: build.tag =~ /^v.*/
-   commands:
-   - ./scripts/build-one-bottle.sh \"$f\"
+   depends_on: \"check-built-$n\"
+   command: |
+     if [ \$\$(buildkite-agent step get \"outcome\" --step \"check-built-$n\") == "passed" ]; then
+       ./scripts/build-one-bottle.sh \"$f\"
+     fi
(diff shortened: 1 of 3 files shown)


.buildkite/generate-tags-pipeline.sh
```diff
@@ -58,21 +71,27 @@ ymlappend "   agents:
    commands:
    - brew uninstall ./Formula/tezos-sapling-params.rb
 
+   # Since using the tag that triggered the pipeline isn't very resilient, we use the version
+   # from the tezos-client formula, which hopefully will stay the most up-to-date.
  - label: Add Big Sur arm64 bottle hashes to formulae
    depends_on:
    - \"uninstall-tsp\"
    if: build.tag =~ /^v.*/
+   soft_fail: true # No artifacts to download if all the bottles are already built
```
rvem, line 80: At which command this step is expected to fail if bottles were already built?

rvem (approved):

.buildkite/generate-tags-pipeline.sh
ixahedron: At `buildkite-agent artifact download <...>`. All other commands should be expected to succeed if it comes to them

PR #413, opened 2022-03-23
[#402] Get network for voting from CL arguments

### LINKED ISSUE #402: [BUG] tezos-voting-wizard asks to choose mainnet vs custom network
# Description

Voting on custom network is only interesting to testers and developers. Regular users will only ever vote on mainnet. For regular users having to choose network first is an a confusing, unnecessary first step. Network is determined by Tezos node to which tezos-client connects. This can default to a local node and should fall back to a some public mainnet node and also be configurable as a command line option.

# To Reproduce

- run tezos-voting-wizard
- observe prompt to choose network

# Expected behavior

- tezos-voting-wizard immediately displays network's governance info and voting choice

ixahedron:
## Description

Problem: currently, Tezos Voting Wizard will ask to choose the network
to vote on first thing upon startup. This introduces an extra step
unnecessary for the vast majority of users, who will only ever want
to vote on mainnet.

Solution: removed the two steps meant to find out the network, replaced
them with a command line argument that defaults to 'mainnet'.

## Related issue(s)

Resolves #402 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_voting_wizard.py
+++ b/docker/package/tezos_voting_wizard.py
@@ -16,17 +16,26 @@ from wizard_structure import *
 
 # Global options
 
-networks = {
-    "mainnet": "Main Tezos network",
-    "custom": "Custom network, will require its name",
-}
-
 ballot_outcomes = {
     "yay": "Vote for accepting the proposal",
     "nay": "Vote for rejecting the proposal",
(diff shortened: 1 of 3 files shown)


rvem (approved):
LGTM

pasqu4le (approved):

Issue #414, opened 2022-03-23
[BUG] Support No-adx version v12


vivienbcr:
# Description

Did you plan to support no adx version for v12 build ?

# To Reproduce

Add serokell repository and `apt install tezos-baking`

Output : 
```
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following additional packages will be installed:
  acl libev4 libhidapi-libusb0 tezos-baker-011-pthangz2 tezos-baker-012-psithaca tezos-client tezos-endorser-011-pthangz2 tezos-node
  tezos-sapling-params
The following NEW packages will be installed:
  acl libev4 libhidapi-libusb0 tezos-baker-011-pthangz2 tezos-baker-012-psithaca tezos-baking tezos-client tezos-endorser-011-pthangz2
  tezos-node tezos-sapling-params
0 upgraded, 10 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/123 MB of archives.
After this operation, 421 MB of additional disk space will be used.
Do you want to continue? [Y/n] Y
Selecting previously unselected package acl.
(Reading database ... 49312 files and directories currently installed.)
Preparing to unpack .../0-acl_2.2.53-4_amd64.deb ...
Unpacking acl (2.2.53-4) ...
Selecting previously unselected package libev4:amd64.
Preparing to unpack .../1-libev4_1%3a4.25-1_amd64.deb ...
Unpacking libev4:amd64 (1:4.25-1) ...
Selecting previously unselected package libhidapi-libusb0:amd64.
Preparing to unpack .../2-libhidapi-libusb0_0.8.0~rc1+git20140818.d17db57+dfsg-2_amd64.deb ...
Unpacking libhidapi-libusb0:amd64 (0.8.0~rc1+git20140818.d17db57+dfsg-2) ...
Selecting previously unselected package tezos-sapling-params.
Preparing to unpack .../3-tezos-sapling-params_2%3a8.2-0ubuntu4~bionic_amd64.deb ...
Unpacking tezos-sapling-params (2:8.2-0ubuntu4~bionic) ...
Selecting previously unselected package tezos-client.
Preparing to unpack .../4-tezos-client_2%3a12.0-0ubuntu1~bionic_amd64.deb ...
Unpacking tezos-client (2:12.0-0ubuntu1~bionic) ...
Selecting previously unselected package tezos-baker-011-pthangz2.
[...]

PR #415, opened 2022-03-24
[#414] Compile ocaml-bls12-381 without ADX extension

rvem:
## Description
Problem: By default ocaml-bls12-381 library (which is used as a
dependency by Octez) is compiled using instructions from ADX extension.
As a result, if package was build on a CPU that has this extension, CPUs
that don't have it won't be able to use binary from this package.

Solution: Compile Octez with 'BLST_PORTABLE=y' to avoid using ADX
instruction in ocaml-bls12-381.

We still need to find a way to test built packages and binaries on old CPU.

## Related issue(s)

Relates #414 (will be resolved once updated packages are released).

  can be successfully built.

### DIFF
--- a/Formula/tezos-accuser-011-PtHangz2.rb
+++ b/Formula/tezos-accuser-011-PtHangz2.rb
@@ -11,7 +11,7 @@ class TezosAccuser011Pthangz2 < Formula
 
   url "https://gitlab.com/tezos/tezos.git", :tag => "v12.0", :shallow => false
 
-  version "v12.0-1"
+  version "v12.0-3"
 
   build_dependencies = %w[pkg-config autoconf rsync wget rustup-init]
   build_dependencies.each do |dependency|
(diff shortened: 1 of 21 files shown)


docker/package/model.py
```diff
@@ -167,6 +167,9 @@ def gen_systemd_rules_contents(package):
     rules_contents = f"""#!/usr/bin/make -f
 
 export DEB_CFLAGS_APPEND=-fPIC
+# Disable usage of instructions from the ADX extension to avoid incompatibility
+# with old CPUs, see https://gitlab.com/dannywillems/ocaml-bls12-381/-/merge_requests/135/
+export BLST_PORTABLE=yes
```
pasqu4le, line 172: not sure if it makes any difference (because it's exported here), but should this also be added to `docker/package/scripts/build-binary.sh`?

pasqu4le (approved):
One question (but I'm pretty sure it makes no difference), otherwise LGTM

docker/package/model.py
rvem: > not sure if it makes any difference (because it's exported here), but should this also be added to `docker/package/scripts/build-binary.sh`?

This script is unfortunately a part of the package's source archive, so it cannot be changed within the same upstream version.

Formula/tezos-accuser-011-PtHangz2.rb
```diff
@@ -34,6 +34,9 @@ class << self
   def make_deps
```
pasqu4le, line 34: Oh I saw now only that you also bumped the release number.

Since you changed formulae as well, I think that we should bump the `version` here too (and make new bottles)

rvem: Right, forgot about it

PR #416, opened 2022-03-25
[Chore] Hotfix 'check-bottle-built.sh'

### LINKED ISSUE #409: Pick GitHub release from formula's version in automated bottle builds

rvem:
## Description
Problem: This script returns 1 exit code in case grep fails due to
'set -e'.

Solution: Allow 'gh release view ... | grep ...' to fail by setting
'set +e' before it.

## Related issue(s)

Relates #409

  can be successfully built.

### DIFF
--- a/scripts/check-bottle-built.sh
+++ b/scripts/check-bottle-built.sh
@@ -19,6 +19,8 @@ fi
 
 tag=$(sed -n "s/^\s\+version \"\(.*\)\"/\1/p" "./Formula/$1.rb")
 
+# Command below is allowed to fail
+set +e
 gh release view "$tag" | grep "$1.*\.$2.bottle.tar.gz"
 
 if [ $? -eq 0 ]; then


pasqu4le (approved):

kirelagin:
Usually this is done by writing `... || true` though...

Issue #421, opened 2022-03-28
Update to Ithaca


yashbhavsar007:
# Clarification and motivation
Want to run two daemons together to support a smooth transition between Hangzhou and Ithaca. 

# Acceptance criteria
Can anyone produce any doc on how to run two daemons parallelly to have a smooth transition into the Ithaca without missing any baking and endorsement?

rvem:
Hi!
In case you're using Ubuntu or Fedora, the easiest way to set up all required Octez binaries to run in the background would be to use `tezos-baking` systemd services. For more information, please refer to [this doc](https://github.com/serokell/tezos-packaging/blob/master/docs/baking.md).
As an alternative, you may use systemd services separately, see [this doc](https://github.com/serokell/tezos-packaging/blob/master/docs/systemd.md). (This doc also explains how you can use systemd services on non-Ubuntu/Fedora systems).
If you're using macOS, you can use background `launchd` services, see [another doc](https://github.com/serokell/tezos-packaging/blob/master/docs/distros/macos.md#launchd-background-services-on-macos).

yashbhavsar007:
Thank you @rvem A feature for a smooth transition before 31st March will be really helpful for the bakers. Once again Thank you.

rvem:
Can the issue be closed now?

yashbhavsar007:
Yes, Thanks.

yashbhavsar007:
@rvem Is there any set of instructions that we can manually do the transition for the new protocol upgrade? As it's already going to happen tomorrow I believe it's hard to introduce the new functionality in code that handles it automatically.

rvem:
If you aren't using `systemd` services provided by the packages, you can still run all required binaries manually.
The list of binaries required for protocol transmission is the following:
1) `tezos-node`
2) `tezos-endorser-011-PtHangz2`
3) `tezos-baker-011-PtHangz2`
4) `tezos-baker-012-Psithaca`

Assuming that you already have a baking instance on mainnet, you should already have 1), 2), and 3) running. So the only thing you need to run additionally is `tezos-baker-012-Psithaca` and it should be run with the same arguments as `tezos-baker-011-PtHangz2`

rvem:
If you're using `tezos-baking`, you shouldn't need to do anything additionally apart from updating all packages to `v12.0-3`.

If you aren't using `tezos-baking`, but use `systemd` other services and orchestrate them manually, you'll need to start `tezos-baker-012-psithaca.service` additionally.

[...]

PR #422, opened 2022-03-28
[#421] Protocol switch-over with 'tezos-baking'

rvem:
## Description

Problem: Protocol switch-over requires users to run two sets of daemons.
Our baking doc should mention how `tezos-baking` services handle it.

Solution: Add a section to the baking doc which mentions that
`tezos-baking` package handles this requirement.

Also, fix `gen_systemd_service_file.py` usage instructions

## Related issue(s)

Relates #421

  can be successfully built.

### DIFF
--- a/docs/baking.md
+++ b/docs/baking.md
@@ -79,6 +79,15 @@ Packages for `tezos-node`, `tezos-baker-<proto>` and `tezos-endorser-<proto>` pr
 systemd units for running the corresponding binaries in the background, these units
 are orchestrated by the `tezos-baking-<network>` units.
 
+## Packages and protocols updates
+
+In order to have a safe transition during a new protocol activation on mainnet,
+it's required to run two sets of daemons: for the current and for the upcoming protocol.
+
+`tezos-baking` package aims to provide such a setup. This package is updated some time before
+the new protocol is activated (usually 1-2 weeks) to run daemons for two protocols. Once the new
+protocol is activated, the `tezos-baking` package is updated again to stop running daemons for the old protocol.
+
 ## Using the wizard
 
 If at this point you want to set up the baking instance, or just a node, using the wizard, run:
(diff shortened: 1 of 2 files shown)


docs/baking.md
```diff
@@ -337,3 +337,12 @@ sudo systemctl restart tezos-node-$tznet
 ```
 
 </details>
+
+## Protocol switch-over
+
+In order to have a safe transition during the protocol activation on mainnet
```
pasqu4le, line 343:
```suggestion
In order to have a safe transition during a new protocol activation on mainnet,
```

docs/baking.md
```diff
@@ -337,3 +337,12 @@ sudo systemctl restart tezos-node-$tznet
...
+it's required to run two sets of daemons: for the current and for the upcoming protocol.
+
+`tezos-baking` package aims to provide such a setup. This package is updated sometime before
```
pasqu4le, line 346:
```suggestion
`tezos-baking` package aims to provide such a setup. This package is updated some time before
```

docs/baking.md
```diff
@@ -337,3 +337,12 @@ sudo systemctl restart tezos-node-$tznet
...
+the new protocol is activated (usually 1-2 weeks) to run daemons for two protocols. Once the new
+protocol is activated, `tezos-baking` package is updated to stop running daemons for the old protocol.
```
pasqu4le, line 348:
```suggestion
protocol is activated, the `tezos-baking` package is updated  again to stop running daemons for the old protocol.
```

docs/baking.md
```diff
@@ -337,3 +337,12 @@ sudo systemctl restart tezos-node-$tznet
 ```
 
 </details>
+
+## Protocol switch-over
```
pasqu4le, line 341: This is kind of an odd (and not very visible) position to put this section IMO.

I'd move it to be a subsection of the `Installation` one instead, since it's mostly about updates to the package itself. What do you think?

rvem: Yeah, I agree that it makes sense. Though, I've renamed the section itself

pasqu4le (approved):

Issue #388, opened 2022-02-23
Properly package the two wizards


ixahedron:
# Clarification and motivation

We now have two CLI wizards with largely similar skeleton, which has been factored out into [its own file](../blob/master/docker/package/wizard_structure.py) that is used in both scripts.

Right now, when installing the `tezos-baking` package, we just copy this common file to `/usr/bin/` together with the wizards' executable scripts. This isn't very proper, and we need to figure out a better way to distribute the packages.

[Here's the guide](https://wiki.debian.org/Python/LibraryStyleGuide?action=show&redirect=Python%2FPackaging) for Debian, or maybe we can look into [`python-poetry`](https://python-poetry.org/). 

# Acceptance criteria

`wizard_structure.py` isn't cluttering `/usr/bin/` and some :proper: packaging approach is used.

PR #433, opened 2022-04-06
[#401] Don't require tezos-baking-mainnet.service to vote

### LINKED ISSUE #401: [BUG] tezos-voting-wizard requires tezos-baking-mainnet service
# Description

tezos-voting-wizard refuses to run without tezos-baking-mainnet service. The only actual requirement for voting is access to ledger device with wallet app running via tezos-client. 

# To Reproduce

- run tezos-voting-wizard
- select "mainnet"
- observe tezos-voting-wizard exits after displaying the following message:

```
Looks like the tezos-baking-mainnet service isn't running.
Please start the service or set up baking.
You can do this by running:
tezos-setup-wizard
```

# Expected behavior

tezos-voting-wizard works on any machine where tezos-client is installed

# Environment

- OS: Debian on RaspberryPi
- branch/revision: 2:12.0-0ubuntu1~bionic

ixahedron:
## Description

Problem: currently, the voting wizard only supports systems which
have baking through systemd services set up. This is not the actual
minimal requirement, which is access to a ledger via tezos-client.

Solution: only check the service for baking info if it's running,
otherwise collect the necessary information from the user.

## Related issue(s)

Resolves #401

  can be successfully built.

### DIFF
--- a/docker/package/tezos_voting_wizard.py
+++ b/docker/package/tezos_voting_wizard.py
@@ -118,23 +118,22 @@ ballot_outcome_query = Step(
     ),
 )
 
-client_data_dir_query = Step(
-    id="client_data_dir",
-    prompt="Provide the path to the data directory of tezos-client.",
-    help="The data directory will be used by tezos-client to vote. If you have baking set up\n"
-    "through systemd services, the path is usually '/var/lib/tezos/.tezos-client' by default.",
-    default=None,
-    validator=Validator([required_field_validator, dirpath_validator]),
-)
 
-node_rpc_addr_query = Step(
-    id="node_rpc_addr",
-    prompt="Provide the node's RPC address.",
-    help="The node's RPC address will be used by tezos-client to vote. If you have baking set up\n"
-    "through systemd services, the address is usually 'http://localhost:8732' by default.",
-    default=None,
-    validator=Validator([required_field_validator]),
-)
+def get_node_rpc_addr_query(default=None):
+    return Step(
+        id="node_rpc_addr",
+        prompt="Provide the node's RPC address.",
+        help="The node's RPC address will be used by tezos-client to vote. If you have baking set up\n"
+        "through systemd services, the address is usually 'http://localhost:8732' by default.",
+        default=default,
+        validator=Validator(
+            [
+                required_field_validator,
+                reachable_url_validator("chains/main/blocks/head/"),
+            ]
+        ),
+    )
+
 
 baker_alias_query = Step(
     id="baker_alias",
@@ -144,6 +143,18 @@ baker_alias_query = Step(
     default=None,
     validator=Validator([required_field_validator]),
 )
+
+# We define the step as a function to disallow choosing json
+def get_key_mode_query(modes):
+    return Step(
+        id="key_import_mode",
+        prompt="How do you want to import the voter key?",
+        help="Tezos Voting Wizard will use the 'baker' alias for the key\n"
+        "that will be used for voting. You will only need to import the key\n"
+        "once unless you'll want to change the key.",
+        options=modes,
+        validator=Validator(enum_range_validator(modes)),
+    )
 
 
 class Setup(Setup):
@@ -178,13 +189,12 @@ class Setup(Setup):
         print("Baker data detected is as follows:")
         print(f"Data directory: {self.config['client_data_dir']}")
         print(f"Node RPC address: {self.config['node_rpc_addr']}")
-        print(f"Baker alias: {self.config['baker_alias']}")
+        print(f"Voter key: {self.config['baker_key_value']}")
         return yes_or_no("Does this look correct? (Y/n) ", "yes")
 
     def search_client_config(self, field, default):
         config_filepath = os.path.join(self.config["client_data_dir"], "config")
         if not os.path.isfile(config_filepath):
-            print("No config file in this directory. Falling back on the default.")
             return default
         else:
             return search_json_with_default(config_filepath, field, default)
@@ -192,30 +202,62 @@ class Setup(Setup):
     def collect_baking_info(self):
         if self.check_baking_service():
             self.fill_baking_config()
+
+            value, _ = get_key_address(
+                self.config["tezos_client_options"], self.config["baker_alias"]
+            )
+            self.config["baker_key_value"] = value
+
             collected = self.check_data_correctness()
         else:
-            # Here, we collect some semi-sensible defaults. Since the baking service isn't
-            # running here, these defaults have a low chance of being correct.
-            print("Default values will be used. Please check them carefully.\n")
 
-            self.config["client_data_dir"] = "/var/lib/tezos/.tezos-client"
-            self.config["node_rpc_addr"] = self.search_client_config(
-                "endpoint", "http://localhost:8732"
-            )
-            self.config["baker_alias"] = "baker"
+            network_dir = "/var/lib/tezos/client-" + self.config["network"]
 
+            proc_call(f"sudo -u tezos mkdir -p {network_dir}")
+
+            print("With no tezos-baking.service running, this wizard will use")
+            print(f"the default directory for this network: {network_dir}")
+
+            self.config["client_data_dir"] = network_dir
+
+            self.config["node_rpc_addr"] = self.search_client_config("endpoint", None)
+            if self.config["node_rpc_addr"] is None:
+                self.query_and_update_config(get_node_rpc_addr_query())
+
+            key_import_modes.pop("json", None)
+            self.get_baker_key()
+
+            # Check correctness in case user wants to change this data upon reruns
             collected = self.check_data_correctness()
 
         while not collected:
-            self.query_step(client_data_dir_query)
+            self.query_and_update_config(
+                get_node_rpc_addr_query(self.config["node_rpc_addr"])
+            )
 
-            self.config["node_rpc_addr"] = search_client_config("endpoint", None)
-            if self.config["node_rpc_addr"] is None:
-                self.query_step(node_rpc_addr_query)
-
-            self.query_step(baker_alias_query)
+            replace_baker_key = self.check_baker_account()
+            if replace_baker_key:
+                key_mode_query = get_key_mode_query(key_import_modes)
+                self.import_key(key_mode_query)
 
             collected = self.check_data_correctness()
+
+    def get_baker_key(self):
+        if "baker_alias" not in self.config:
+            self.config["baker_alias"] = "baker"
+
+        self.config["tezos_client_options"] = self.get_tezos_client_options()
+
+        baker_key_value = get_key_address(
+            self.config["tezos_client_options"], self.config["baker_alias"]
+        )
+
+        if baker_key_value is not None:
+            value, _ = baker_key_value
+            self.config["baker_key_value"] = value
+        else:  # if there is no key with this alias, query import
+            key_mode_query = get_key_mode_query(key_import_modes)
+            self.import_key(key_mode_query)
 
     def get_network(self):
         if parsed_args.network == "mainnet":
--- a/docker/package/tezos_voting_wizard.py
+++ b/docker/package/tezos_voting_wizard.py
@@ -49,15 +49,9 @@ protocol_hash_regex = (
 
 welcome_text = """Tezos Voting Wizard
 
-Welcome, this wizard will help you to vote in the Tezos protocol amendment process.
-Please note that for this you need to already have set up the baking infrastructure,
-normally on mainnet, as only bakers can submit ballots or proposals.
-
-If you have installed tezos-baking, you can run Tezos Setup Wizard to set it up:
-tezos-setup-wizard
-
-Alternatively, you can use this wizard for voting on a custom chain,
-which also needs to be set up already.
+Welcome, this wizard will help you vote in the Tezos protocol amendment process.
+Please note that to vote on mainnet, the minimum requirement is to have access
+to a key that has voting rights, preverably through a connected ledger device.
 
 All commands within the service are run under the 'tezos' user.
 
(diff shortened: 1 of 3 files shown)


pasqu4le:
> This is not the actual minimal requirement, which is access to a ledger via tezos-client.

Is it? I'm pretty sure that one can vote without a ledger too (although I guess it's not recommended)

docker/package/tezos_voting_wizard.py
```diff
@@ -50,14 +50,15 @@
...
+- a connected ledger device with Tezos Wallet app running;
 
 If you have installed tezos-baking, you can run Tezos Setup Wizard to set it up:
 tezos-setup-wizard
 
 Alternatively, you can use this wizard for voting on a custom chain,
-which also needs to be set up already.
+baking on which needs to be set up already.
 
 All commands within the service are run under the 'tezos' user.
```
pasqu4le, line 63: I think that we should reconsider this part, because it gives a lot of info that in most cases people won't be interested in.

IMO here we should only mention the _actual_ minimum requirements for voting.

I think that we can avoid mentioning custom networks too, and leave that to be explained in the docs instead.

As for the baking setup(and how to run it), I suggest mentioning it later, in case it isn't found.

docker/package/tezos_voting_wizard.py
```diff
@@ -142,23 +170,38 @@ def check_baking_service(self):
...
+        if self.check_baking_service():
+            self.fill_baking_config()
+            collected = self.check_data_correctness()
+        else:
+            # Here, we collect some semi-sensible defaults. Since the baking service isn't
+            # running here, these defaults have a low chance of being correct.
+            print("Default values will be used. Please check them carefully.\n")
+
+            self.config["client_data_dir"] = "/var/lib/tezos/.tezos-client"
+            self.config["node_rpc_addr"] = "http://localhost:8732"
```
pasqu4le, line 195: :thinking: if there is a config file in the data dir (which should be the case) can't this be found in the `endpoint` there (if specified)?

docker/package/tezos_voting_wizard.py
```diff
@@ -50,14 +50,15 @@
 welcome_text = """Tezos Voting Wizard
 
 Welcome, this wizard will help you to vote in the Tezos protocol amendment process.
-Please note that for this you need to already have set up the baking infrastructure,
-normally on mainnet, as only bakers can submit ballots or proposals.
+Please note that to vote on mainnet, you need to already have one of the following setups:
+- the baking infrastructure on mainnet;
+- a connected ledger device with Tezos Wallet app running;
```
rvem, line 55: @pasqu4le already mentioned it as a comment to the commit message, but I'll repeat that we don't actually limit the ways secret keys are provided

docker/package/tezos_voting_wizard.py
rvem: That's actually a tricky part, because the user may provide _various_ `tezos-client` data directories that may potentially belong to a user that is different from the one that started `tezos-voting-wizard`. So we should either dynamically define which user is going to run `tezos-client` commands or say that they always run under the current user and if one needs the other user, he has to do `sudo -u tezos-voting-wizard`

rvem: Also, if the target is to have a wizard for voting with minimum possible requirements, I think it should be moved to a dedicated package, because currently `tezos-baking` has quite a lot of redundant dependencies for voting

ixahedron: I agree about the dedicated package, though I think in this case especially #388 should be solved before that or at the same time. It might be best to open another issue for it instead of separating it here.

I'm not sure running under the current user/through `sudo -u` will be good enough though :thinking: I can imagine there can be dependency on the two users (starting and e.g. `tezos`) being different and having different rights, but even if not, using e.g. `sudo -u tezos tezos-voting-wizard` looks unnecessarily cluttered to me. I think it's best to indeed define the second user dynamically. Looks like we only call `tezos-client` after collecting all the config data anyway, so it shouldn't be a problem to find out the user to whom `client_data_dir` belongs. Or can you think of any problems with that approach?

pasqu4le: IDK about moving the wizard somewhere else :shrug: 

IMO it can stay in the same package for now, because as I mentioned even tho we want to be able to support minimal requirements it wouldn't hurt to give a better experience for those that have set up baking already with the other wizard.

There still seems to be quite a bit of overlap between the two IMO

rvem: > I think it's best to indeed define the second user dynamically. Looks like we only call `tezos-client` after collecting all the config data anyway, so it shouldn't be a problem to find out the user to whom `client_data_dir` belongs. Or can you think of any problems with that approach?

I don't think that it should cause any problems. And I'm in favor of implementing it this way

rvem: > For both (2) and (3) I think that this wizard should guide into importing the key of the voter, rather than juggling with tezos-client data dirs. In the worst case scenario one would have to go through the import procedure on a machine where they already performed it (in a different dir)

Do you propose to use a dedicated data directory for each voting wizard invocation?

rvem: In general, it shouldn't be too inconvenient. Because one only has to import a key in order to vote (it isn't required to set up a ledger to bake and register the key as a delegate). As for different directories, I guess we can have `tezos-client` data directories per network (e.g. marked with `chain_id` obtained from the corresponding node). This way we'll avoid reimporting the key next time the user will vote because he'll highly likely use the same key.

pasqu4le: > Do you propose to use a dedicated data directory for each voting wizard invocation?

Ah no, perhaps that was confusing. What I was saying is that in the remote case that one has already that key imported 
 somewhere (e.g. they did so using `tezos-client` with their own user and the default config) then they'd have to re-import it with what I propose, which is IMO not that inconvenient (and not happening in most cases).

What I was suggesting is to use the `tezos` user and the `--network` argument to determine the data dir, like the setup wizard does (with `get_data_dir` AFAIK).

docker/package/tezos_voting_wizard.py
```diff
@@ -142,23 +169,97 @@ def check_baking_service(self):
...
+            proc_call(f"sudo -u tezos mkdir -p {network_dir}")
+
+            print("With no tezos-baking.service running, this wizard will use")
+            print(f"the default directory for this network: {network_dir}")
+
+            self.config["client_data_dir"] = network_dir
+
+            self.config["node_rpc_addr"] = self.search_client_config("endpoint", None)
+            if self.config["node_rpc_addr"] is None:
+                self.query_step(get_node_rpc_addr_query())
```
rvem, line 220: Let's check that provided RPC endpoint actually works. Currently, it will only display warnings and will fail to fetch voting period info at the very end of wizard work

docker/package/tezos_voting_wizard.py
```diff
@@ -142,23 +169,97 @@ def check_baking_service(self):
...
+                self.config["tezos_client_options"] = self.get_tezos_client_options()
+                proc_call(
+                    f"sudo -u tezos {suppress_warning_text} tezos-client "
```
rvem, line 223: For some reason, when the node endpoint is already present in the client config and another endpoint was used during voting, an endpoint in the config isn't updated. Is that expected?
```
Baker data detected is as follows:
Data directory: /var/lib/tezos/client-mainnet
Node RPC address: http://localhost:8734
Voter key: ...
Does this look correct? (Y/n) n

Provide the node's RPC address.
> http://localhost:8735
...
Thank you for voting!
$ tezos-voting-wizard 
Tezos Voting Wizard
...
the default directory for this network: /var/lib/tezos/client-mainnet
Baker data detected is as follows:
Data directory: /var/lib/tezos/client-mainnet
Node RPC address: http://localhost:8734
Voter key: ...
```

ixahedron: No, thanks for catching that. Made a fix

docker/package/tezos_voting_wizard.py
ixahedron: Do you think just pinging it (as in the last version) is enough, or should we specifically check that it's a node somehow?

rvem: I think you can check that something like `curl <endpoint>/chains/main/blocks/head/` doesn't fail

docker/package/wizard_structure.py
```diff
@@ -45,12 +50,31 @@ def _validator(input):
...
     if input and not os.path.isfile(input):
         raise ValueError("Please input a valid file path.")
     return input
 
 
+def reachable_url_validator(suffix=None):
+    def _validator(input):
+        if input:
+            try:
+                print(urllib.parse.urljoin(input, suffix))
```
rvem, line 69: looks like a leftover

docker/package/wizard_structure.py
```diff
@@ -45,12 +50,31 @@ def _validator(input):
...
+                urllib.request.urlopen(urllib.parse.urljoin(input, suffix))
+            except (urllib.error.URLError, ValueError):
+                raise ValueError("Please input a reachable URL.")
```
rvem, line 72: I think error should inform why we treat provided endoint as invalid
```suggestion
                raise ValueError(f"{urllib.parse.urljoin(input, suffix)} is unreachable. Please input valid node endpoint URL.")
```

docker/package/wizard_structure.py
```diff
@@ -302,6 +413,81 @@ def pprint_options(self):
                 else:
                     print(str_format.format(i, o, description))
                 i += 1
+        elif not self.options and self.default is not None:
+            print("Default:", self.default)
```
rvem, line 417: Now wizard prints some weird lines:
```
How do you want to import the voter key?
1. ledger (default)            From a ledger
2. secret-key                  Either the unencrypted or password-encrypted secret key for
                               your address
3. remote                      Remote key governed by a signer running on a different
                               machine
> 2
Provide either the unencrypted or password-encrypted secret key for your address.
Default: 1
>
```

rvem (approved):
LGTM. Have only a few cosmetic suggestions

docker/package/wizard_structure.py
ixahedron: It's a validator for any reachable URLs though, e.g. it's also used in the setup wizard now to validate snapshot URLs. I think it's worth it to preserve generality in the validator and just depend on the step prompt. Printing the full suffix URL sounds good though

docker/package/wizard_structure.py
ixahedron: Well, AFAIU that step shouldn't have a default anyway

docker/package/tezos_voting_wizard.py
```diff
@@ -49,15 +49,9 @@
...
-normally on mainnet, as only bakers can submit ballots or proposals.
-
-If you have installed tezos-baking, you can run Tezos Setup Wizard to set it up:
-tezos-setup-wizard
-
-Alternatively, you can use this wizard for voting on a custom chain,
-which also needs to be set up already.
+Welcome, this wizard will help you vote in the Tezos protocol amendment process.
+Please note that to vote on mainnet, the minimum requirement is to have access
+to a key that has voting rights, preverably through a connected ledger device.
```
pasqu4le, line 54:
```suggestion
to a key that has voting rights, preferably through a connected ledger device.
```

docker/package/tezos_voting_wizard.py
```diff
@@ -125,6 +119,44 @@ def get_proposal_period_hash(hashes):
...
+    return Step(
+        id="node_rpc_addr",
+        prompt="Provide the node's RPC address.",
+        help="The node's RPC address will be used by tezos-client to vote. If you have baking set up\n"
+        "through systemd services, the address is usually 'http://localhost:8732' by default.",
+        default=default,
+        validator=Validator(
+            [
+                required_field_validator,
+                reachable_url_validator("chains/main/blocks/head/"),
```
pasqu4le, line 132: Minor annoyance: it seems that this step requires the URL to be given with a `/` at the end or it won't work.

Moreover the error message will say:
```
Validation error: chains/main/blocks/head/ is unreachable. Please input a valid URL.
```
that doesn't make what the problem is very clear.

docker/package/tezos_voting_wizard.py
```diff
@@ -125,6 +119,44 @@ def get_proposal_period_hash(hashes):
 )
 
 
+def get_node_rpc_addr_query(default=None):
```
pasqu4le, line 122: There doesn't seem to be a reliable source of public nodes, but perhaps we can still give multiple choice (per network) that we know are working now and one to input it manually (like now).

This would give a default option that's good enough in most cases. What do you think? 

P.s. I'm ok with opening a separate issue for this.

docker/package/wizard_structure.py
```diff
@@ -385,3 +585,137 @@ def fill_baking_config(self):
...
+                    proc_call(
+                        f"sudo -u tezos {suppress_warning_text} tezos-client {tezos_client_options} "
+                        f"activate account {baker_alias} with {json_tmp_path} --force"
+                    )
+                    try:
+                        os.remove(json_tmp_path)
+                    except:
+                        pass
+                else:
+                    print("Please open the Tezos Baking app on your ledger.")
```
pasqu4le, line 658: So... this has no escape. If I get here by mistake (which isn't too unlikely given that this is, correctly, the default) there is no way to abort, except by closing the whole wizard.

Maybe there should be a way to stop the waiting and go back to the previous choice?

pasqu4le (commented):
@ixahedron I just tested the current version and have some minor comments.

IMO they can be solved in another MR in the scope of the same issues because they are not that big (at least they don't seem so).

docker/package/tezos_voting_wizard.py
ixahedron: If you input the full URL including the scheme, it works as intended 🤔 I've made a fix with regards to correct URL joining, but it will only help with the error reporting, as `urllib.request.urlopen` doesn't accept links without a scheme anyway, nor do I think we should make it.

PR #440, opened 2022-04-14
[#397] Support remote signers in setup wizard

### LINKED ISSUE #397: Allow using remote signers in 'tezos-setup-wizard'

Users may want to use a remote signer for their baker. However, currently `tezos-setup-wizard` rejects URIs for remote signers.

An example of remote signer URI: `http://127.0.0.1:6732/keys/tz1V8fDHpHzN8RrZqiYCHaJM9EocsYZch5Cy`.
`tezos-signer` supports multiple schemes for remote keys, see [documentation](https://tezos.gitlab.io/user/key-management.html#signer).

# Acceptance criteria

It should be possible to use URI provided by the `tezos-signer` during baker setup via `tezos-setup-wizard`

ixahedron:
## Description

Problem: one of the options for key management is using remote
signers. However, Tezos Setup Wizard currently doesn't support it.

Solution: added remote as one of the options for key import.

## Related issue(s)

Resolves #397 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -42,6 +42,7 @@ snapshot_import_modes = {
 key_import_modes = {
     "ledger": "From a ledger",
     "secret-key": "Either the unencrypted or password-encrypted secret key for your address",
+    "remote": "Remote key governed by a signer running on a different machine",
     "json": "Faucet JSON file from https://teztnets.xyz/",
 }
 
@@ -63,7 +64,6 @@ TMP_SNAPSHOT_LOCATION = "/tmp/tezos_node.snapshot"
 # Regexes
 
 ledger_regex = b"ledger:\/\/[\w\-]+\/[\w\-]+\/[\w']+\/[\w']+"
-secret_key_regex = b"(encrypted|unencrypted):(?:\w{54}|\w{88})"
 address_regex = b"tz[123]\w{33}"
 derivation_path_regex = b"(?:bip25519|ed25519|secp256k1|P-256)\/[0-9]+h\/[0-9]+h"
 
@@ -255,6 +255,15 @@ secret_key_query = Step(
     help="The format is 'unencrypted:edsk...' for the unencrypted key, or 'encrypted:edesk...'"
     "for the encrypted key.",
     validator=Validator([required_field_validator, secret_key_validator]),
+)
+
+remote_signer_uri_query = Step(
+    id="remote_signer_uri",
+    prompt="Provide your remote key with the address of the signer.",
+    help="The format is the address of your remote signer host, followed by a public key,\n"
+    "i.e. something like http://127.0.0.1:6732/tz1V8fDHpHzN8RrZqiYCHaJM9EocsYZch5Cy\n"
+    "The supported schemes are https, http, tcp, and unix.",
+    validator=Validator([required_field_validator, signer_uri_validator]),
 )
 
 json_filepath_query = Step(
@@ -511,7 +520,15 @@ class Setup(Setup):
             f"show address {baker_alias} --show-secret"
         )
         if address.returncode == 0:
-            value_regex = b"(?:" + ledger_regex + b")|(?:" + secret_key_regex + b")"
+            value_regex = (
+                b"(?:"
+                + ledger_regex
+                + b")|(?:"
+                + secret_key_regex
+                + b")|(?:remote\:"
+                + address_regex
+                + b")"
+            )
             value = re.search(value_regex, address.stdout).group(0)
             address = re.search(address_regex, address.stdout).group(0)
             print()
@@ -524,6 +541,14 @@ class Setup(Setup):
             )
         else:
             return True
+
+    def fill_remote_signer_infos(self):
+        self.query_step(remote_signer_uri_query)
+
+        rsu = re.search(signer_uri_regex.decode(), self.config["remote_signer_uri"])
+
+        self.config["remote_host"] = rsu.group(1)
+        self.config["remote_key"] = rsu.group(2)
 
     # Importing the baker key
     def import_baker_key(self):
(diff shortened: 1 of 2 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -66,6 +67,7 @@
 secret_key_regex = b"(encrypted|unencrypted):(?:\w{54}|\w{88})"
 address_regex = b"tz[123]\w{33}"
 derivation_path_regex = b"(?:bip25519|ed25519|secp256k1|P-256)\/[0-9]+h\/[0-9]+h"
+signer_uri_regex = b"((?:tcp|unix|https|http):\/\/.+)\/keys\/(tz[123]\w{33})\/?"
```
rvem, line 70: Can we import it from `wizard_structure` to avoid duplication?

docker/package/wizard_structure.py
```diff
@@ -18,6 +18,7 @@
 protocol_hash_regex = (
     b"P[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{50}"
 )
+signer_uri_regex = b"((?:tcp|unix|https|http):\/\/.+)\/keys\/(tz[123]\w{33})\/?"
```
rvem, line 21: Seems like this regex is too strict:
```
Provide your remote key with the address of the signer.
> tcp://localhost:7732/tz1QMRfsLZUyyHVthxPwoKgCyubSG1RWoLnV
Validation error: The input doesn't match the format for a remote signer URI: (tcp|unix|https|http)://<host address>/<public key address>
Please check the input and try again.
```

rvem, line 21: Does `keys` part of the URL actually needed?
```suggestion
signer_uri_regex = b"((?:tcp|unix|https|http):\/\/.+)\/(tz[123]\w{33})\/?"
```

rvem: Looks like `keys/` part is actually redundant for key importing:
```
sudo -u tezos tezos-client import secret key remote tcp://localhost:7732/keys/tz1atFdTEzZPDvq4XRpghYQ9g3SrXUHvduSC --force
Error:
  Cannot parse the key uri: tcp://localhost:7732/keys/tz1atFdTEzZPDvq4XRpghYQ9g3SrXUHvduSC
  Failed to read a b58check_encoding data (Signature.Public_key_hash): "keys/tz1atFdTEzZPDvq4XRpghYQ9g3SrXUHvduSC"
```

rvem: Oh, I think I understand now. This regex doesn't follow the remote URL format specified by `tezos-client` and is used to determine the key and the host separately. @ixahedron, do I understand it correctly?

ixahedron: 🤔 I only tested it with `http`, but when running e.g. `tezos-client -R http://127.0.0.1:6732 import secret key baker remote:tz1atFdTEzZPDvq4XRpghYQ9g3SrXUHvduSC`, it actually tries to connect to `http://127.0.0.1:6732/keys/tz1atFdTEzZPDvq4XRpghYQ9g3SrXUHvduSC`, with the `keys/` part. It's also the format you used as an example in the original issue. I assumed this is the format one will use when sharing key addresses, too. So, no, my intention wasn't to invent a new format 🙈

rvem: hmm, seems like different signing schemes have different key URLs then :confused:

docker/package/tezos_setup_wizard.py
ixahedron: Good catch, missed it 👍

docker/package/wizard_structure.py
rvem: Surprisingly `tcp://localhost:7732/keys/tz1atFdTEzZPDvq4XRpghYQ9g3SrXUHvduSC` works with the wizard, but doesn't work with `tezos-client`

ixahedron: > tcp://localhost:7732/keys/tz1atFdTEzZPDvq4XRpghYQ9g3SrXUHvduSC works with the wizard
Yeah, 'cause it indeed separates the URI into the host and the key to use with the supposedly more flexible `-R` flag, throwing out the `keys/` part.

ixahedron: I couldn't find documentation on the signer URI format btw 😕

rvem: Neither do I

rvem:
```
$ sudo -u tezos tezos-client import secret key remote http://localhost:6732/keys/tz1atFdTEzZPDvq4XRpghYQ9g3SrXUHvduSC --force
Error:
  Rpc request failed:
     - meth: GET
     - uri: http://localhost:6732/keys/keys/tz1atFdTEzZPDvq4XRpghYQ9g3SrXUHvduSC
     - error: Oops! It looks like we forged an invalid HTTP request.
                Failed to parsed an argument in path. After "keys/keys", the value "failed to decode Base58Check-encoded data (Signature.Public_key_hash): "keys"" is not acceptable for type "pkh"
```

Hmm, apparently, I provided the wrong URI example initially :see_no_evil:

rvem: And `keys/` part is actually added by the `tezos-client` implicitly

rvem: Let's stick with the format used by the `tezos-client import secret key` command to be more convenient for `tezos-signer` users

docker/package/tezos_setup_wizard.py
```diff
@@ -257,6 +257,15 @@ def get_key_mode_query(modes):
     validator=Validator([required_field_validator, secret_key_validator]),
 )
 
+remote_signer_uri_query = Step(
+    id="remote_signer_uri",
+    prompt="Provide your remote key with the address of the signer.",
+    help="The format is the address of your remote signer host, followed by a public key,\n"
+    "i.e. something like http://127.0.0.1:6732/keys/tz1V8fDHpHzN8RrZqiYCHaJM9EocsYZch5Cy\n"
```
rvem, line 264:
```suggestion
    "i.e. something like http://127.0.0.1:6732/tz1V8fDHpHzN8RrZqiYCHaJM9EocsYZch5Cy\n"
```

docker/package/wizard_structure.py
```diff
@@ -97,12 +98,26 @@ def secret_key_validator(input):
...
-            "\nPlease check the input and try again."
+            "The input doesn't match the format for a derivation path: "
+            + derivation_path_regex_str
+            + "\nPlease check the input and try again."
+        )
+    return input
+
+
+# To be validated, the input should adhere to the signer URI format:
+# (tcp|unix|https|http)://<host address>/keys/tz[123]\w{33}
```
rvem, line 113:
```suggestion
# (tcp|unix|https|http)://<host address>/tz[123]\w{33}
```

rvem (approved):
A couple of leftovers. The rest LGTM

PR #443, opened 2022-04-25
[#442] Support testnets in the voting wizards

### LINKED ISSUE #442: Support testnets in the voting wizard

Currently the voting wizard supports `mainnet` and custom networks, but not testnets.

Arguably this isn't a very useful feature, but it would be nice to have.

Note: this will need a list of currently supported testnets, which the setup wizard already has, so it would be great to have this list common between the two (to make it easier to update it with time).

# Acceptance criteria

It's possible to select a testnet in the voting wizard (through the `--network` command line option).

krendelhoff:
## Description

Problem: Currently the voting wizard supports mainnet and custom networks, but not testnets.

Solution: Add support for the testnets.

## Related issue(s)

Resolves #442 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -19,12 +19,6 @@ from wizard_structure import *
 
 # Global options
 
-networks = {
-    "mainnet": "Main Tezos network",
-    "hangzhounet": "Test network using version 011 of Tezos protocol (Hangzhou)",
-    "ithacanet": "Test network using version 012 of Tezos protocol (Ithaca2)",
-}
-
 modes = {
     "baking": "Set up and start all services for baking: "
     "tezos-node, tezos-baker, and tezos-endorser.",
--- a/docker/package/tezos_voting_wizard.py
+++ b/docker/package/tezos_voting_wizard.py
@@ -22,7 +22,6 @@ ballot_outcomes = {
     "pass": "Submit a vote not influencing the result but contributing to quorum",
 }
 
-
 # Command line argument parsing
 
 parser.add_argument(
@@ -30,7 +29,7 @@ parser.add_argument(
     required=False,
     default="mainnet",
     help="Name of the network to vote on. Is 'mainnet' by default, "
-    "but can take the (part after @) name of any custom instance. "
+    "but can be a testnet or the (part after @) name of any custom instance. "
     "For example, to use the tezos-baking-custom@voting service, input 'voting'. "
     "You need to already have set up the custom network using systemd services.",
 )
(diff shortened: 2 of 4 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -14,17 +14,12 @@
 import urllib.request
 import json
 from typing import List
+from available_networks import networks
```
ixahedron, line 17: Yeah, this would be nicer than sticking everything into one shared file. Unfortunately, this is not going to work in its current form. You can see #388 for a bit more context, but the gist of it is that we can't assume we can import stuff from files local to this repository, as the .deb package we build with docker behaves/installs things mostly very differently on our end users' machines.

If you are set on retaining this additional file, you need to also add to [`model.py`](https://github.com/serokell/tezos-packaging/blob/master/docker/package/model.py) (look how the rest of this wizard is handled there). But honestly, until #388 is solved, I'd prefer not to clutter the `/usr/bin/` with any more files than necessary, so I think it's better to just stick the networks (and anything else you factor out) into the one `wizard_structure.py`.

ixahedron (commented):
Looks good except for that one comment :tada:

docker/package/wizard_structure.py
```diff
@@ -14,6 +14,14 @@
...
 import json
 
+# Currently available networks
+
+networks = {
+    "mainnet": "Main Tezos network",
+    "hangzhounet": "Test network using version 011 of Tezos protocol (Hangzhou)",
+    "ithacanet": "Test network using version 012 of Tezos protocol (Ithaca2)",
+}
+
```
ixahedron, line 24: One last thing: there's a section `# Global options` further down, please move this there for ease of reference.

ixahedron (approved):
One last nitpick, otherwise LGTM

rvem (approved):

docker/package/tezos_voting_wizard.py
```diff
@@ -22,15 +22,14 @@
...
-
 # Command line argument parsing
 
 parser.add_argument(
     "--network",
     required=False,
     default="mainnet",
     help="Name of the network to vote on. Is 'mainnet' by default, "
-    "but can take the (part after @) name of any custom instance. "
+    "but can take the (part after @) name of any custom instance or testnet. "
```
pasqu4le, line 32: Minor nitpick, I'd suggest changing this to:
```suggestion
    "but can be a testnet or the (part after @) name of any custom instance. "
```
because it seems a little bit less confusing: in the current wording one could think of looking for the _part after @ of testnets_ which is not there.

pasqu4le (approved):
LGTM too, I only left a minor suggestion.

Feel free to squash your fixups already :+1:

PR #446, opened 2022-04-26
[#401] Fix URLs and waiting for ledger

ixahedron:
## Description

Problem: there are several minor inconveniences affecting both Tezos
wizards: the URL validator sometimes fails to report an issue correctly,
and it's not possible to go back after choosing ledger as your key import
method to choose another one without closing the whole wizard.

Solution: changed the URL joining method, added an out to waiting for the
ledger app.

## Related issue(s)

Follows #433

  can be successfully built.

### DIFF
--- a/docker/package/tezos_voting_wizard.py
+++ b/docker/package/tezos_voting_wizard.py
@@ -50,7 +50,7 @@ welcome_text = """Tezos Voting Wizard
 
 Welcome, this wizard will help you vote in the Tezos protocol amendment process.
 Please note that to vote on mainnet, the minimum requirement is to have access
-to a key that has voting rights, preverably through a connected ledger device.
+to a key that has voting rights, preferably through a connected ledger device.
 
 All commands within the service are run under the 'tezos' user.
 
@@ -128,7 +128,7 @@ def get_node_rpc_addr_query(default=None):
         validator=Validator(
             [
                 required_field_validator,
-                reachable_url_validator("chains/main/blocks/head/"),
+                reachable_url_validator("chains/main/blocks/head/header"),
             ]
         ),
     )
@@ -160,14 +160,7 @@ class Setup(Setup):
 
     # Check whether the baker_alias account is set up to use ledger
     def check_ledger_use(self):
-        baker_alias = self.config["baker_alias"]
-        address = get_proc_output(
-            f"sudo -u tezos {suppress_warning_text} tezos-client "
-            f"{self.config['tezos_client_options']} show address {baker_alias} --show-secret"
-        )
-        if address.returncode == 0:
-            value_regex = b"(?:" + ledger_regex + b")"
-            return bool(re.match(value_regex, address.stdout))
+        return bool(re.match(ledger_regex.decode(), self.config["baker_key_value"]))
 
     def check_baking_service(self):
         net = self.config["network"]
@@ -175,11 +168,8 @@ class Setup(Setup):
             proc_call(f"systemctl is-active --quiet tezos-baking-{net}.service")
             return True
         except:
-            print(f"Looks like the tezos-baking-{net} service isn't running.")
-            print("If this is a mistake, and you should have a baking instance")
-            print("running on mainnet on this machine, please check if it is set up.")
-            print("If it isn't, you can use Tezos Setup Wizard to set it up:")
-            print("tezos-setup-wizard")
+            print(f"No local baking services for {net} running on this machine.")
+            print("If there should be, you can run 'tezos-setup-wizard' to set it up.")
             print()
 
             return False
@@ -201,6 +191,7 @@ class Setup(Setup):
     def collect_baking_info(self):
         if self.check_baking_service():
             self.fill_baking_config()
+            self.config["tezos_client_options"] = self.get_tezos_client_options()
 
             value, _ = get_key_address(
                 self.config["tezos_client_options"], self.config["baker_alias"]
@@ -237,7 +228,7 @@ class Setup(Setup):
             replace_baker_key = self.check_baker_account()
             if replace_baker_key:
                 key_mode_query = get_key_mode_query(key_import_modes)
-                self.import_key(key_mode_query)
+                self.import_key(key_mode_query, "Wallet")
 
             collected = self.check_data_correctness()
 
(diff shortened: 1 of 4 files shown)


docker/package/tezos_voting_wizard.py
```diff
@@ -129,7 +129,7 @@ def get_node_rpc_addr_query(default=None):
         validator=Validator(
             [
                 required_field_validator,
-                reachable_url_validator("chains/main/blocks/head/"),
+                reachable_url_validator("chains/main/blocks/head/header"),
```
rvem, line 132: Are there nodes that restrict `chains/main/blocks/head`, but don't restrict `chains/main/blocks/head/header`?

rvem (approved):
LGTM. Ledger app requirements work as expected now :relieved:

pasqu4le:
I just realized that the voting test instructions in `docs/tests` still mention the need to launch a local baking instance.

Could this too be avoided now? AFAIK one only needs a node to be running there and pushing the local chain forward.

docker/package/tezos_voting_wizard.py
```diff
@@ -257,7 +251,7 @@ def get_baker_key(self):
             self.config["baker_key_value"] = value
         else:  # if there is no key with this alias, query import
             key_mode_query = get_key_mode_query(key_import_modes)
-            self.import_key(key_mode_query)
+            self.import_key(key_mode_query, "Wallet")
 
     def get_network(self):
         if parsed_args.network == "mainnet":
```
pasqu4le, line 257: _leaving this comment here because GH doesn't allow me to leave it in a better place_

I noticed now the `Choose again?` option given when voting in the proposing period (aka submitting a proposal).

I think we should remove it, because I don't think it's that useful and if needed one can quickly re-launch the wizard and submit a new hash in a matter of seconds.

docker/package/tezos_voting_wizard.py
```diff
@@ -161,14 +161,7 @@ class Setup(Setup):
...
-        address = get_proc_output(
-            f"sudo -u tezos {suppress_warning_text} tezos-client "
-            f"{self.config['tezos_client_options']} show address {baker_alias} --show-secret"
-        )
-        if address.returncode == 0:
-            value_regex = b"(?:" + ledger_regex + b")"
-            return bool(re.match(value_regex, address.stdout))
+        return bool(re.match(ledger_regex.decode(), self.config["baker_key_value"]))
 
     def check_baking_service(self):
```
pasqu4le, line 166: I still think that the current message for when no baking is detected is a bit too long:
```
Looks like the tezos-baking-{net} service isn't running.
If this is a mistake, and you should have a baking instance
running on mainnet on this machine, please check if it is set up.
If it isn't, you can use Tezos Setup Wizard to set it up:
tezos-setup-wizard
```
(_also, it says "mainnet" on the third line no matter what_)

I think it should be simplified to something like:
```
No local baking setup for {net} was detected on this machine.
If you want, you can run the `tezos-setup-wizard` to set one up.
```

docker/package/tezos_voting_wizard.py
```diff
@@ -202,6 +195,7 @@ def search_client_config(self, field, default):
     def collect_baking_info(self):
```
pasqu4le, line 195: Likewise IMO when no baking is detected the message here:
```
the default directory for this network: {network_dir}
```
is unnecessary, because it's immediately followed by the recap from `check_data_correctness` which contains a line for the data directory,

This info is hardly ever necessary to know, IMO there's no need to repeat it twice.

pasqu4le (commented):
I have a couple of suggestions, but otherwise it LGTM now.

docker/package/tezos_voting_wizard.py
ixahedron: I disagree. With no message like this, there is no indication that it's not possible to change the directory like it's possible to change the other two parameters shown: node RPC address and the voter key. It's not a long message, so cutting it would not really conserve screen space, but IMO would make it (a bit) less clear how the wizard operates.

docker/package/tezos_voting_wizard.py
ixahedron: Again, I disagree slightly :see_no_evil: In the proposal period, it's allowed to submit/vote up to 20 times. As the voting wizard must process these one-by-one and not in batches (note this is a conscious and IMO good decision, as it's [officially](https://obsidian.systems/blog/voting-on-tezos-with-your-ledger-nano-s) not recommended to batch-submit proposals when using ledgers), it's very likely that what you're proposing will lead to 20 invocations with a lot of "does this info seem right to you?" spam, i.e. it will actively bloat the amount of "clicks" (key presses, I guess) one needs to make in order to submit the maximum 20 proposals. The `Choose again?` question is a better alternative IMO.

pasqu4le: Oh, interesting, I didn't know about the limit of 20 proposals.

I agree that it's a better experience if one has to submit many of them, but ... how often will that happen.

AFAIU this only makes sense for _different_ proposal and from what I recall the most were proposed after (the first) ithaca was rejected at [period #64](https://agora.tezos.com/period/64) and it was only 4 for the whole community :shrug:

docker/package/tezos_voting_wizard.py
pasqu4le: :thinking: okay, I'm not sure if that would be a problem, but it's a fair point.

However isn't this the case also when there is a baking service running? IIRC there is no such message in that case tho.

ixahedron: That's true, but from the docs I think it should be more-or-less clear that the two wizards are able to somewhat work in tandem. I'm not against adding a line like "With the service running, it will use the directory specified in the service config" (though I don't think it's strictly necessary), but I'm against removing this one.

docker/package/tezos_voting_wizard.py
ixahedron: Yeah, fair point. For some reason I thought duplicates counted as votes (e.g. 20 submissions for one proposal would boost its rate more than 1 submission), but I see that's not the case, so it definitely makes sense to exit right away.

docker/package/tezos_voting_wizard.py
pasqu4le: Ok, I suppose that's fair. 

I don't mind these lines so much, so I'm ok with leaving them in :ok_hand:

pasqu4le (approved):

PR #449, opened 2022-04-28
[#403] Services fail to start

### LINKED ISSUE #403: [BUG] Services set up via tezos-baking-mainnet fail to start
# Description
tezos-baking-mainnet fails for me: it takes longer than ~13s for node to start responding to endorser, but endorser service gives up  trying to reconnect to node after several attempts:  retrying in 1 second, 1.50 seconds, 2.25, 3.38, 5.06 and then exits, which causes systemd to restart other service dependencies of tezos-baking-mainnet including tezos-node (and the same with baker service) and it never gets out of this cycle.

# To Reproduce

- set up baking with tezos-setup-wizard on RaspberryPi (or other low spec/slow machine)
- observe tezos-node is constantly restarted

# Expected behavior

- everything works

# Environment

- OS: Debian 64-bit on RaspberyPi4
- branch/revision: 2:12.0-0ubuntu1~bionic

# Additional context

krendelhoff:
## Description

If node needs more time to start responding, than baker/endorser services can wait for it, systemd restarts all the tezos services indefinetily.

## Related issue(s)

Resolves #403

  can be successfully built.

### DIFF
--- a/docker/package/scripts/tezos-baker-start
+++ b/docker/package/scripts/tezos-baker-start
@@ -56,6 +56,12 @@ launch_baker() {
     fi
 }
 
+# TODO we should use --keep-alive instead of this loop once
+# https://gitlab.com/tezos/tezos/-/issues/2873 is fixed
+# Waiting till node start to respond
+
+while ! "$tezos_client" --endpoint "$NODE_RPC_ENDPOINT" rpc list &> /dev/null; do sleep 1; done
+
 if [[ -z "$BAKER_ADDRESS_ALIAS" ]]; then
     launch_baker "$@"
 else


docker/package/scripts/tezos-node-start
```diff
@@ -41,3 +42,12 @@ if [[ -z "$CERT_PATH" || -z "$KEY_PATH" ]]; then
 else
     exec "$node" run "${node_run_args[@]}" --rpc-tls="$CERT_PATH","$KEY_PATH"
 fi
+
+# Waiting till node start to respond
+
+started=false
+while ! $started
```
rvem, line 49: I think this should work as well
```suggestion
while ! $started
```

docker/package/scripts/tezos-node-start
```diff
@@ -41,3 +41,7 @@ if [[ -z "$CERT_PATH" || -z "$KEY_PATH" ]]; then
 else
     exec "$node" run "${node_run_args[@]}" --rpc-tls="$CERT_PATH","$KEY_PATH"
 fi
+
+# Waiting till node start to respond
+
+while ! curl "$NODE_RPC_ADDR" &> /dev/null; do sleep 1; done
```
rvem, line 47: Seems like commands that are supposed to be executed after `exec` aren't actually performed since `exec` replace the current process with the provided:
```
$ cat test.sh 
#!/usr/bin/env bash

exec echo "kek"

echo "mda"
$ ./test.sh
kek
```

krendelhoff: Why do we use exec here?

rvem: I don't remember, to be honest. `baker` and `accuser` services use `<...> &` and `Type=forking` instead. Maybe we can use it in the node service as well

krendelhoff: Actually responding is needed for baker and not (crucial) for node, so it makes sense to wait in baker

docker/package/scripts/tezos-endorser-start
```diff
@@ -46,3 +46,7 @@ if [[ -z "$BAKER_ADDRESS_ALIAS" ]]; then
 else
     launch_endorser "$BAKER_ADDRESS_ALIAS" "$@"
 fi
+
+# Waiting till node start to respond
+
```
rvem, line 51: Let's leave a `TODO` that we should use `--keep-alive` instead of this loop once https://gitlab.com/tezos/tezos/-/issues/2873 is fixed

docker/package/scripts/tezos-baker-start
```diff
@@ -45,3 +45,7 @@ if [[ -z "$BAKER_ADDRESS_ALIAS" ]]; then
 else
     launch_baker "$BAKER_ADDRESS_ALIAS" "$@"
 fi
+
+# Waiting till node start to respond
+
+while ! "$tezos_client" --endpoint "$NODE_RPC_ENDPOINT" rpc list &> /dev/null; do sleep 1; done
```
rvem, line 51: Hmmm, does it make sense to run this loop after the `launch_baker` is invoked? This way baker process still may die in the background due to connection error

krendelhoff: Makes no sense actually, fixed

rvem (approved):

PR #455, opened 2022-05-03
[Chore] More tezos modules options

rvem:
## Description

This PR adds a few options to the existing NixOS modules with background Octez daemons.

## Related issue(s)

https://issues.serokell.io/issue/TM-645

  can be successfully built.

### DIFF
--- a/nix/modules/tezos-baker.nix
+++ b/nix/modules/tezos-baker.nix
@@ -12,12 +12,22 @@ let
     "013-PtJakart" =
       "${pkgs.ocamlPackages.tezos-baker-013-PtJakart}/bin/tezos-baker-013-PtJakart";
   };
+  tezos-client = "${pkgs.ocamlPackages.tezos-client}/bin/tezos-client";
   cfg = config.services.tezos-baker;
   common = import ./common.nix { inherit lib; inherit pkgs; };
   instanceOptions = types.submodule ( {...} : {
     options = common.daemonOptions // {
 
       enable = mkEnableOption "Tezos baker service";
+
+      bakerSecretKey = mkOption {
+        type = types.nullOr types.str;
+        default = null;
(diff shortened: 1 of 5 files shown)


nix/modules/tezos-baker.nix
```diff
@@ -27,6 +37,12 @@ let
         '';
       };
 
+      liquidityBakingToggleVote = mkOption {
+        type = types.enum [ "on" "of" "pass" ];
+        default = "pass";
+        description = "Vote to continue (option "on") or end (option "off") the liquidity baking subsidy. Or choose to pass (option "pass")";
```
ixahedron, line 43: Does nix handle the quotes-inside-quotes correctly? I'd use multi-line string for description as in the previous option, for consistency and to keep everything ~80 chars wide, but at the minimum I suspect you'd have to escape them.

nix/modules/tezos-baker.nix
```diff
@@ -12,13 +12,23 @@ let
...
   instanceOptions = types.submodule ( {...} : {
     options = common.daemonOptions // {
 
       enable = mkEnableOption "Tezos baker service";
 
+      bakerSecretKey = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = ''
+          Secret key stat will be imported as 'bakerAccountAlias' before
```
ixahedron, line 27: > stat

:thinking: 

```suggestion
          If provided, secret key will be imported for 'bakerAccountAlias' before
```

ixahedron (approved):
LGTM aside from a couple nitpicks

nix/modules/tezos-baker.nix
rvem: Hmm, not sure what I meant here 🤔

PR #456, opened 2022-05-04
[#418] Reduce bottle building in BK to one step

### LINKED ISSUE #417: CI: always upload new brew bottles singularly

Some of our CI infrastructure that builds brew bottles running on `buildkite` waits until the end of the pipeline to upload all the newly built bottles.

We currently also have checks to avoid rebuilding bottles if they are already attached to the corresponding release on GitHub.

The combination of these two facts means that if the pipeline is interrupted in the middle then it's possible to trigger an unnecessary rebuild of the bottles built up until that point.

Given that bottle building takes much time and resources, we should fix this up.

# Acceptance criteria

On `buildkite` pipelines newly build bottles are uploaded to their GH release as soon as they are available.

### LINKED ISSUE #418: CI: reduce brew bottle builds to one script run (per OS)

Currently the CI builds bottle automatically for different macOS version using an infrastructure that's split between GitHub Actions and buildkite pieplines.

Moreover in both cases, to avoid the repetition of unnecessary builds in case of failures, each bottle is built in separate steps, which results in pipeline that are partially generated, with many interactions between them.

This is all quite a bit messy and will likely get harder to maintain over time, as more differences between the two build systems might need to be taken into account.

We should attempt to reduce this by having a single, bigger and smarter, step for bottle building (per OS version).

# Acceptance criteria

Bottle building happens (in both GA and BK) in a single large step that:
1. filters down the full batch to only formulae that don't have a bottle in a GH release already
2. installs the common `tezos-sapling-params` once
3. builds and uploads bottles one by one (see #417)
4. correctly reports failure if any of the bottles couldn't be built (or uploaded)

ixahedron:
## Description
Problem: currently, after each release, we partially generate the Buildkite pipeline to handle building brew bottles for formulae relevant to the release. This script has been extended since its conception and is becoming difficult to maintain.

Solution: reworked the BK tags pipeline to include a single step for each OS, calling an external script that filters the formulae to the list of those that have been changed, and builds and uploads them one-by-one.

## Related issue(s)

Resolves #418
Resolves #417

  can be successfully built.


.buildkite/pipeline-for-tags.yml
```diff
@@ -32,3 +32,37 @@ steps:
...
+ - label: Build Big Sur x86_64 bottles
+   key: build-bottles-big-sur-x86_64
+   if: build.tag =~ /^v.*/
+   agents:
+     queue: "x86_64-rosetta-darwin"
+   commands:
+   - nix-shell ./scripts/shell.nix --run './scripts/build-all-bottles.sh "big_sur"'
+   artifact_paths:
+     - '*.bottle.*'
+
```
rvem, line 45:
```suggestion

 - wait

```

Brew may misbehave when there are two `brew` processes running together

scripts/build-all-bottles.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ -z "$1" ] ; then
+    echo "Please call this script with the name of the OS for which to build the bottles."
+    exit 1
+fi
+
+# we don't bottle meta-formulas that contain only services
+formulae=("tezos-accuser-011-PtHangz2" "tezos-accuser-012-Psithaca" "tezos-admin-client" "tezos-baker-011-PtHangz2" "tezos-baker-012-Psithaca" "tezos-client" "tezos-codec" "tezos-endorser-011-PtHangz2" "tezos-node" "tezos-sandbox" "tezos-signer")
+
+# we filter out only the formulas that were changed and so need to be rebuilt
+changed_formulae=()
```
rvem, line 18: Nitpick: I think that "changed" is slightly confusing. This script needs to be restarted when e.g. there are issues with the network connection to the builder. So I'd rather say that we filter out formulae that don't have bottles in the corresponding release

scripts/build-all-bottles.sh
```diff
@@ -0,0 +1,47 @@
...
+for f in "${formulae[@]}"; do
+  if ./scripts/check-bottle-built.sh "$f" "$1"; then
```
rvem, line 20: Does it make sense to move this check to the `for` loop below (and traverse over `formulae` in it) to reduce the amount of code in this script a bit?

scripts/build-all-bottles.sh
```diff
@@ -0,0 +1,42 @@
...
+brew install --formula ./Formula/tezos-sapling-params.rb
+
+for f in "${formulae[@]}"; do
+  # check if the formula doesn't already have a bottle in its respective release
+  if ./scripts/check-bottle-built.sh "$f" "$1"; then
+    # build a bottle
+    if ./scripts/build-one-bottle.sh "$f"; then
+      # upload the bottle to its respective release
+      FORMULA_TAG=$(sed -n "s/^\s\+version \"\(.*\)\"/\1/p" "./Formula/$f.rb")
+      export FORMULA_TAG
```
rvem, line 29: btw, do we actually need to export this variable now?

rvem (approved):
LGTM, only few minor nitpicks left

scripts/build-all-bottles.sh
ixahedron: You're right, we don't

PR #460, opened 2022-05-06
[#447] Provide a list of public nodes

### LINKED ISSUE #447: Provide a list of public nodes in the voting wizard

The voting wizard sometimes has to ask the user to input a node RPC address. It's fully possible this will be a public node.

We would like to provide a list of _working_ public nodes to choose from at this step already, to provide some options for the average user to choose from without searching for an address themself.

It seems [there is](https://tezostaquito.io/docs/rpc_nodes/) a small list of public nodes, though not all of them seem to work with the `tezos-client` and the format is not easily scrappable. Consequently, we will probably have to not only maintain our own list, but also check if they work on each wizard invocation.

# Acceptance criteria

- [ ] The 'node RPC address' query step provides a list of working public nodes to choose from.
- [ ] One of them, the most trusted one if possible, is the default option.
- [ ] It's possible to manually input an address.

krendelhoff:
## Description

Problem: It's possible that user will use public node for voting.
We should provide a list of existing public nodes to choose.

Solution: Provide list of reachable public nodes for corresponding network.

## Related issue(s)

Resolves #447 

  can be successfully built.


docker/package/tezos_voting_wizard.py
```diff
@@ -13,6 +13,15 @@
 import re
 
 from wizard_structure import *
+from dataclasses import dataclass
```
ixahedron, line 16: IIRC we can't use `dataclass` because we still need to support Ubuntu 18.04, and that LTS has Python 3.6, whereas `dataclass` is introduced in 3.7 :cry:

docker/package/tezos_voting_wizard.py
```diff
@@ -22,6 +22,21 @@
...
 
+public_nodes = {
+    "mainnet": [
+        ("https://rpc.tzbeta.net", "Tezos Foundation"),
+        ("https://mainnet.api.tez.ie", "ECAD Labs"),
+        ("https://mainnet.smartpy.io", "SmartPy"),
+        ("https://teznode.letzbake.com", "LetzBake!"),
+        ("https://mainnet-tezos.giganode.io", "GigaNode"),
+    ],
+    "ithacanet": [
```
rvem, line 33: Voting doesn't usually happen on testnet, so I don't think that we should keep a list of public nodes on `ithacanet` and `jakartanet`

rvem:
I had some previous voting setup and now voting wizard fails with an error when I attempt to reconfigure it:
```
$ tezos-voting-wizard 
Tezos Voting Wizard

...
the default directory for this network: /var/lib/tezos/client-mainnet
Baker data detected is as follows:
Data directory: /var/lib/tezos/client-mainnet
Node RPC address: http://localhost:8735
Voter key: ledger://live-fossa-eager-walrus/bip25519/0h/0h
Does this look correct? (Y/n) n

Error in Tezos Voting Wizard, exiting.
The error has been logged to /home/roman/tezos_voting_wizard.log
$ cat /home/roman/tezos_voting_wizard.log
'http://localhost:8735'
```

docker/package/tezos_voting_wizard.py
krendelhoff: Don't we want to have an opportunity to vote on testnet? There was issue related to that #442

docker/package/tezos_voting_wizard.py
```diff
@@ -118,17 +133,42 @@ def get_proposal_period_hash(hashes):
...
+
+    def get_node_rpc_addr_query_validator(nodes):
+        def helper(answer):
+            if answer.isdigit():
+                answer = int(answer)
+                if answer not in range(1, len(nodes) + 1):
+                    raise ValueError("Please choose one of the provided values")
+                return nodes[answer - 1][0]
+            else:
+                return node_validator(answer)
```
ixahedron, line 147: This is very similar to `enum_range_validator` which already exists. Perhaps you can use something like `required_field_validator if relevant_nodes is None/<empty> else enum_range_validator(relevant_nodes)`? I haven't thought about this in depth, but seems like it should be possible

docker/package/tezos_voting_wizard.py
ixahedron: Yeah, we want to be able to vote on testnets, but it's mostly for our own testing purposes. AFAIU no actual user will ever vote on testnet, and if they do, they should be able to find or run a node themself. So I agree that IMO it's better to just keep a list of mainnet nodes, as they are likely to be used.

ixahedron:
I'm going to re-review this a bit later, but just for the future reference, please-please-please don't squash fixup commits until everyone involved approves to merge the PR. It makes re-reviewing a hundred times more difficult

krendelhoff:
I found last solution to be significantly different from the first version, so decided to squash everything

docker/package/wizard_structure.py
```diff
@@ -416,7 +433,7 @@ def pprint_options(self):
                     i += 1
         elif self.options and isinstance(self.options, dict):
             index_len = len(str(len(self.options)))
-            str_format = f"{{:{index_len}}}. {{:<26}}  {{}}"
+            str_format = f"{{:{index_len}}}. {{:<43}}  {{}}"
```
rvem, line 436: Now some `tezos-setup-wizard` steps look a bit weird:
```
Do you want to set up baking or to run the standalone node?
1. baking (default)                             Set up and start all services for baking: tezos-node and
                               tezos-baker.
2. node                                         Only bootstrap and run the Tezos node.
```
```
Which history mode do you want your node to run in?
1. rolling (default)                            Store a minimal rolling window of chain data, lightest
                               option
2. full                                         Store enough chain data to reconstruct the complete chain
                               state
3. archive                                      Store all the chain data, very storage-demanding


```
I'd like to avoid hardcoding `43` here. Perhaps you can use the length of the largest option plus some constant instead

docker/package/tezos_voting_wizard.py
```diff
@@ -118,17 +126,31 @@ def get_proposal_period_hash(hashes):
...
 
-def get_node_rpc_addr_query(default=None):
+def get_node_rpc_addr_query(network="mainnet", default=None):
+    node_not_alive = reachable_url("chains/main/blocks/head/header")
+    custom_url_validator = reachable_url_validator("chains/main/blocks/head/header")
+
+    relevant_nodes = {
+        url: provider
+        for url, provider in public_nodes.items()
+        if not (network != "mainnet" or node_not_alive(url))
```
ixahedron, line 136: :che:

ixahedron (changes requested):

docker/package/wizard_structure.py
```diff
@@ -416,10 +424,14 @@ def pprint_options(self):
...
             index_len = len(str(len(self.options)))
-            str_format = f"{{:{index_len}}}. {{:<26}}  {{}}"
+            max_option_len = max(map(len, self.options.keys()))
+            padding = max(26, max_option_len + 2)
+            indent_size = index_len + 2 + padding
+            str_format = f"{{:{index_len}}}. {{:<{padding}}}  {{}}"
             for o in self.options:
                 description = textwrap.indent(
-                    textwrap.fill(self.options[o], 60), " " * 31
+                    textwrap.fill(self.options[o], int(padding * (60 / 26))),
```
ixahedron, line 433: Oh nice, that's clever :+1: Unfortunately, I don't think it'll work well. This makes the right half of the options listing (all the descriptions) approximately twice (`60/26` times) as long as the left part. Works out to 60 characters in most cases, but precisely when the options are very long, like with URLs, this will stretch the description out a lot and make it ill-indented or not fit onto most screens. See what I mean:

```
>>> max_option_len = 80
>>> padding = max(26, max_option_len + 2)
>>> indent_size = 1 + 2 + padding
>>> print(textwrap.indent(textwrap.fill("a" * 400, 60), " " * indent_size))
                                                                                     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                                                     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                                                     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                                                     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                                                     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                                                     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                                                     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

>>> print(textwrap.indent(textwrap.fill("a" * 400, int(padding * (60/26))), " " * indent_size))
                                                                                     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                                                     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                                                     aaaaaaaaaaaaaaaaaaaaaa
```

Leaving the width at 60 is fine IMO, it's the indentation of it that matters for the padding anyway, which you have already fixed.

ixahedron (approved):
Looks good to me! 🎉 I left another two small-ish suggestions, but I'll approve already, though it would be cool to still wait for a second re-review from @rvem. If you haven't already, please go through the voting process testing out your changes.

docker/package/wizard_structure.py
krendelhoff: Actually the idea was to make it shorter, but I somehow wrote the oblivosly wrong formula 🙈

rvem: Hmm, still doesn't look well in some cases
```
Do you want to set up baking or to run the standalone node?
1. baking (default)            Set up and start all services for baking: tezos-node and
                             tezos-baker.
2. node                        Only bootstrap and run the Tezos node.
```

rvem: Looks like `index_len` indent is missing

krendelhoff: Indent for dot + space was missing actually

rvem (approved):

Issue #441, opened 2022-04-21
Use xtz-shots snapshot metadata


pasqu4le:
# Clarification and motivation

Our setup wizard has an option to import a snapshot, when setting up a new node, by downloading it from `xtz-shots.io`.

Currently this option is based on the fact that the latest snapshots there follow a predictable pattern and that we can "guess" the URL where we can find them.

Recently, however, they started serving snapshot metadata on `https://mainnet.xtz-shots.io/base.json`, which would be a much more stable mechanism, and asked for feedback on it.

We should attempt to use this in our wizard, to pick a snapshot to download, and help them fine-tune that endpoint it if possible.

**NOTE**: I think that this issue shouldn't be closed as soon as a PR for it is ready, specifically because we may go back and forth with them on the specifics of the endpoint mentioned above and its content.

# Acceptance criteria

The setup wizard picks what snapshot to download from `xtz-shots` using the metadata endpoint, rather than expecting it to match a URL format.

ixahedron:
But we can still rely on the network part of the URL, do I understand correctly? As in, it'll always be `https://mainnet.xtz-shots.io/<...>` but maybe not `https://mainnet.xtz-shots.io/rolling`

pasqu4le:
Upstream issues about where and how this metadata is served:
- [tezos-k8s/#446](https://github.com/oxheadalpha/tezos-k8s/issues/446)
- [tezos-k8s/#447](https://github.com/oxheadalpha/tezos-k8s/issues/447)

ixahedron:
Upstream issue about JSON formatting: 
- [tezos-k8s/#451](https://github.com/oxheadalpha/tezos-k8s/issues/451)

PR #465, opened 2022-05-13
[#441] Use xtz-shots.io snapshot metadata

ixahedron:
## Description

Problem: in Tezos Setup Wizard, we allow to automatically download a suitable snapshot from xtz-shots.io. The links have a set pattern that we can predict and reuse, but it's not necessarily a stable solution. xtz-shots.io recently added a metadata JSON to their site, and using it to get snapshot URLs would be more robust.

Solution: switched the snapshot download step to fetching snapshot metadata from base.json and choosing the most recent available entry to offer as options.

## Related issue(s)

Follows #441

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -232,6 +232,35 @@ class Setup(Setup):
             return False
         return True
 
+    # Check https://xtz-shots.io/tezos-snapshots.json and collect the most recent snapshot
+    # that is suited for the chosen history mode and network
+    def get_snapshot_link(self):
+        self.config["snapshot_url"] = None
+        json_url = "https://xtz-shots.io/tezos-snapshots.json"
+        try:
+            with urllib.request.urlopen(json_url) as url:
+                snapshot_array = json.load(url)
+                snapshot_array.sort(reverse=True, key=lambda x: int(x["block_height"]))
+        except (urllib.error.URLError, ValueError):
+            print(f"Couldn't collect snapshot metadata from {json_url}")
+            return
+
+        self.config["snapshot_url"] = next(
+            filter(
+                lambda artifact: artifact["artifact_type"] == "tezos-snapshot"
+                and artifact["chain_name"] == self.config["network"]
+                and (
+                    artifact["history_mode"] == self.config["history_mode"]
+                    or (
+                        self.config["history_mode"] == "archive"
+                        and artifact["history_mode"] == "full"
+                    )
+                ),
+                iter(snapshot_array),
+            ),
+            {"url": None},
+        )["url"]
+
     # Importing the snapshot for Node bootstrapping
     def import_snapshot(self):
         do_import = self.check_blockchain_data()
@@ -245,10 +274,16 @@ class Setup(Setup):
                 f"--history-mode {self.config['history_mode']}"
             )
 
-            if self.config["history_mode"] == "rolling":
+            self.get_snapshot_link()
+
+            if self.config["snapshot_url"] is None:
+                snapshot_import_modes.pop("download rolling", None)
+                snapshot_import_modes.pop("download full", None)
+            elif self.config["history_mode"] == "rolling":
                 snapshot_import_modes.pop("download full", None)
             else:
                 snapshot_import_modes.pop("download rolling", None)
+
         else:
             return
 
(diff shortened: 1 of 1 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -222,6 +222,37 @@ def check_blockchain_data(self):
...
+        network_url = f"https://{self.config['network']}.xtz-shots.io"
+        json_url = f"{network_url}/base.json"
+        try:
+            with urllib.request.urlopen(json_url) as url:
+                snapshot_array = json.load(url)
+        except (urllib.error.URLError, ValueError):
+            print(f"Couldn't collect snapshot metadata from {json_url}")
+            return
+
+        for artifact in snapshot_array:
```
rvem, line 238: The structure of metadata JSON is highly likely going to change in the future, it'll probably be an unexpected change for us and users. What do you think about using a fallback option (e.g. use the previous `https://<network>.xtz-shots.io/rolling`) in case the JSON that we got has some unexpected format? This fallback workaround will be removed once the format of JSON will be stabilized

pasqu4le: I'd also like to remind that we might not want to merge this right away and wait a bit to see if out feedback is taken care of.

In fact, I'd recommend setting this as a `Draft` temporarily

docker/package/tezos_setup_wizard.py
```diff
@@ -232,6 +232,32 @@ def check_blockchain_data(self):
...
+                continue
+
+            # We just need the first instance of the right snapshot. To speed filtering up,
+            # it is assumed metadata is sorted chronologically in descending order.
+            if artifact["history_mode"] == self.config["history_mode"] or (
+                self.config["history_mode"] == "archive"
+                and artifact["history_mode"] == "full"
+            ):
+                self.config["snapshot_url"] = artifact["url"]
+                return
```
pasqu4le, line 259: Not sure if it makes a difference in terms of performance (I suspect so tho) but how about using python's `filter` to remove the non-needed elements from the iterator and `next` to grab the first element from it?

I believe iterators are fairly efficient and should keep the memory down and avoid unnecessary computations.

rvem: Also, I'm looking at the https://jakartanet.xtz-shots.io/base.json and it seems like the newest snapshot is actually the last one in the JSON, so seems like this snapshot array should be traversed from the tail

krendelhoff2: is it ok? maybe it's time for another issue about `base.json` format?

pasqu4le: I think that there is no need for a separate issue, especially because this issue/MR is still waiting for https://github.com/oxheadalpha/tezos-k8s/issues/446

docker/package/tezos_setup_wizard.py
```diff
@@ -232,6 +232,35 @@ def check_blockchain_data(self):
             return False
         return True
 
+    # Check <network>.xtz-shots.io/base.json and collect the most recent snapshot
```
pasqu4le, line 235: This comment seems outdated

docker/package/tezos_setup_wizard.py
```diff
@@ -245,10 +274,17 @@ def import_snapshot(self):
...
+            self.get_snapshot_link()
+
             if self.config["history_mode"] == "rolling":
                 snapshot_import_modes.pop("download full", None)
+                if self.config["snapshot_url"] is None:
+                    snapshot_import_modes.pop("download rolling", None)
             else:
                 snapshot_import_modes.pop("download rolling", None)
+                if self.config["snapshot_url"] is None:
+                    snapshot_import_modes.pop("download full", None)
```
pasqu4le, line 286: Small suggestion:
```suggestion
            if self.config["snapshot_url"] is None:
                snapshot_import_modes.pop("download rolling", None)
                snapshot_import_modes.pop("download full", None)
            elif self.config["history_mode"] == "rolling":
                snapshot_import_modes.pop("download full", None)
            else:
                snapshot_import_modes.pop("download rolling", None)
```

docker/package/tezos_setup_wizard.py
```diff
@@ -232,6 +232,35 @@ def check_blockchain_data(self):
...
 
+    # Check <network>.xtz-shots.io/base.json and collect the most recent snapshot
+    # that is suited for the chosen history mode
+    def get_snapshot_link(self):
+        self.config["snapshot_url"] = None
+        json_url = "https://xtz-shots.io/base.json"
+        try:
+            with urllib.request.urlopen(json_url) as url:
+                snapshot_array = json.load(url)
+                snapshot_array.reverse()
```
pasqu4le, line 243: IIUC this is done this so that the most recent snapshot will be on top, right?

If so, wouldn't it be better to sort them by `block_height` instead, which would be a more stable approach?

docker/package/tezos_setup_wizard.py
```diff
@@ -277,34 +313,15 @@ def import_snapshot(self):
...
-                        "Please check your internet connection or choose another option."
-                    )
-                    print()
-                    continue
-            elif self.config["snapshot"] == "download full":
-                url = "https://" + self.config["network"] + ".xtz-shots.io/full"
+            else:
+                url = self.config["snapshot_url"]
                 try:
                     snapshot_file = fetch_snapshot(url)
```
pasqu4le, line 318:
```suggestion
                try:
                    snapshot_file = fetch_snapshot(self.config["snapshot_url"])
```

docker/package/tezos_setup_wizard.py
```diff
@@ -232,6 +232,35 @@ def check_blockchain_data(self):
             return False
         return True
 
+    # Check https://xtz-shots.io/base.json and collect the most recent snapshot
+    # that is suited for the chosen history mode and network
+    def get_snapshot_link(self):
+        self.config["snapshot_url"] = None
+        json_url = "https://xtz-shots.io/base.json"
```
pasqu4le, line 239: This was moved to a different URL
```suggestion
        json_url = "https://xtz-shots.io/tezos-snapshots.json"
```
I guess that the previous one is still accessible, but as we noticed it doesn't serve the same results.

pasqu4le (approved):

PR #468, opened 2022-05-17
[#388] Properly package two wizards

### LINKED ISSUE #388: Properly package the two wizards

krendelhoff:
## Description

Problem: Right now, when installing the tezos-baking package, we just
copy this common file to /usr/bin/ together with the wizards'
executable scripts.

Solution: Change the buildsystem to pybuild.

## Related issue(s)

Resolves #388

  can be successfully built.

### DIFF
--- a/docker/package/Dockerfile-fedora
+++ b/docker/package/Dockerfile-fedora
@@ -4,7 +4,7 @@
 FROM fedora:32
 RUN dnf update -y
 RUN dnf install -y libev-devel gmp-devel hidapi-devel libffi-devel zlib-devel libpq-devel m4 perl git pkg-config \
-  rpmdevtools python3 wget opam rsync which cargo autoconf
+  rpmdevtools python3-devel python3-setuptools wget opam rsync which cargo autoconf
 ENV USER dockerbuilder
 RUN useradd dockerbuilder && mkdir /tezos-packaging
 ENV HOME /tezos-packaging
@@ -17,3 +17,4 @@ COPY docker/package/defaults /tezos-packaging/docker/package/defaults
 COPY docker/package/scripts /tezos-packaging/docker/package/scripts
 COPY LICENSE /tezos-packaging/LICENSE
 ENTRYPOINT ["python3", "-m", "package.package_generator"]
+
--- a/docker/package/fedora.py
+++ b/docker/package/fedora.py
@@ -20,7 +20,7 @@ def build_fedora_package(
     home = os.environ["HOME"]
 
     pkg.fetch_sources(dir)
-    pkg.gen_makefile(f"{dir}/Makefile")
+    pkg.gen_buildfile("/".join([dir, pkg.buildfile]))
     pkg.gen_license(f"{dir}/LICENSE")
     for systemd_unit in pkg.systemd_units:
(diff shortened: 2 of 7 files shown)


docker/package/model.py
```diff
@@ -633,28 +644,22 @@ def gen_spec_file(self, build_deps, run_deps, out):
...
-
-install: tezos-baking tezos-setup-wizard tezos-voting-wizard
-	mkdir -p $(DESTDIR)$(BINDIR)
-	cp $(CURDIR)/wizard_structure.py $(DESTDIR)$(BINDIR)/wizard_structure.py
-	cp $(CURDIR)/tezos-setup-wizard $(DESTDIR)$(BINDIR)/tezos-setup-wizard
-	cp $(CURDIR)/tezos-voting-wizard $(DESTDIR)$(BINDIR)/tezos-voting-wizard
+from setuptools import setup
+
+setup(
+    name='tezos-baking',
```
rvem, line 652: Currently `tezos-baking` python package has `0.0.0` version. However, it's possible to define version as an argument of `setup`, let's use the version from `self.meta` here to provide the actual version of the package

rvem:
Hmm, I've tried to build `tezos-client` binary `.deb` package and it seems like binary wasn't actually built

docker/package/Dockerfile-fedora
```diff
@@ -16,4 +16,4 @@ COPY docker/package/*.py /tezos-packaging/docker/package/
 COPY docker/package/defaults /tezos-packaging/docker/package/defaults
 COPY docker/package/scripts /tezos-packaging/docker/package/scripts
 COPY LICENSE /tezos-packaging/LICENSE
-ENTRYPOINT ["python3", "-m", "package.package_generator"]
+ENTRYPOINT ["python3", "-m", "package.package_generator"]
```
rvem, line 19: The newline at the end of the file is missing now

rvem:
Overall, the approach seems quite good, but some additional fixes are still needed. Also, we should check that this works on the oldest Ubuntu version that we support

docker/package/fedora.py
```diff
@@ -20,7 +20,7 @@ def build_fedora_package(
     home = os.environ["HOME"]
 
     pkg.fetch_sources(dir)
-    pkg.gen_makefile(f"{dir}/Makefile")
+    pkg.gen_buildfile("/".join([dir, pkg.buildfile]))
```
rvem, line 23: Do I understand correctly that `rpmbuild` is capable of using `setup.py` instead of `Makefile` out of the box?

PR #471, opened 2022-05-19
[#438] Document PPA installation steps on Debian

### LINKED ISSUE #438: Document PPA installation steps on Debian

tezos-packaging currently has a [launchpad PPA](https://launchpad.net/~serokell) to distribute Tezos software on Ubuntu.

There are no instructions for an installation on Debian and the ones for Ubuntu won't work because of the different way in which the two OSs are versioned.

However, it should be possible to use out PPA on _some_ Debian releases with similar instructions as [the ones provided for Raspberry Pis](https://github.com/serokell/tezos-packaging/blob/master/docs/distros/ubuntu.md#raspberry) by making the correct translation of OS version names (see [related askubuntu thread](https://askubuntu.com/a/445496)).

# Acceptance criteria

The documentation in this repo explains:
1. what commands can be run on Debian to use our launchpad PPA 
2. what versions of Debian are supported (depending on the Ubuntu ones available)

krendelhoff:
## Description

Problem: We should be able to use serokell launchpad ppa
on some debian releases, but straightforward way won't work
because of different versioning policies.

Solution: Provide commands to install on debian and update docs.

## Related issue(s)

Resolves #438 

  can be successfully built.

### DIFF
--- a/docs/distros/ubuntu.md
+++ b/docs/distros/ubuntu.md
@@ -30,6 +30,26 @@ As an addition, `tezos-baking` package provides `tezos-baking-<network>` service
 systemd units for `tezos-node` and `tezos-baker-<proto>`.
 Configuration files for these services are located in `/etc/default/tezos-baking-<network>`.
 
+<a name="debian"></a>
+## Ubuntu packages on Debian
+
+You can add the PPA using:
+```
+# Install software properties commons
+sudo apt-get install software-properties-common gnupg
+# Add PPA with Tezos binaries
+sudo add-apt-repository 'deb http://ppa.launchpad.net/serokell/tezos/ubuntu bionic main'
+sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 37B8819B7D0D183812DCA9A8CE5A4D8933AE7CBB
+sudo apt-get update
+```
+If packages for `bionic` are not suited for your Debian version, see the
+[related askubuntu thread](https://askubuntu.com/a/445496) to choose a valid one.
+
+Then install with `apt-get`, e.g. for `tezos-client`:
+```
+sudo apt-get install tezos-client
+```
+
 <a name="raspberry"></a>
 ## Ubuntu packages on Raspberry Pi OS
 
@@ -38,7 +58,7 @@ you can use the Launchpad PPA to install `tezos-*` executables on it as well.
 
 You can add the PPA using:
 ```
-# Intall software properties commons
+# Install software properties commons
 sudo apt-get install software-properties-common
 # Add PPA with Tezos binaries
 sudo add-apt-repository 'deb http://ppa.launchpad.net/serokell/tezos/ubuntu focal main'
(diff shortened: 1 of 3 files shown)


docs/support-policy.md
```diff
@@ -22,6 +22,16 @@ Currently, these are versions:
 
 There are packages for `arm64` and `amd64` architectures.
 
+## Debian packages
+
+We aim to provide packages for all [stable releases](https://www.debian.org/releases/)
```
pasqu4le, line 27: Ah, I don't think we should really say that, because as mentioned in the issue:
> what versions of Debian are supported (depending on the Ubuntu ones available)

so, we really only aim to support some Ubuntu versions, which incidentally means that some Debian versions end up being supported at all.

So here I think we should say that we _don't_ aim to support all stable debian releases, but only those that are matching the supported Ubuntu versions.

docs/distros/ubuntu.md
```diff
@@ -30,6 +30,24 @@ As an addition, `tezos-baking` package provides `tezos-baking-<network>` service
...
 
+<a name="debian"></a>
+## Ubuntu packages on Debian
+
+You can add the PPA using:
+```
+# Install software properties commons
+sudo apt-get install software-properties-common gnupg
+# Add PPA with Tezos binaries
+sudo add-apt-repository 'deb http://ppa.launchpad.net/serokell/tezos/ubuntu bionic main'
```
pasqu4le, line 41: The problem with Debian is that using `bionic` here only works on some Debian versions, but not others.

As mentioned in the issue description:
> (see [related askubuntu thread](https://askubuntu.com/a/445496)).

which needs to be used to figure out what to pick in this command.

krendelhoff: I checked on bullseye and buster, it works

pasqu4le: Ah, I see, nice to hear! 

AFAICT It looks like:
1. `stretch` is still under LTS, but we no longer support the equivalent Ubuntu
2. `bookworm` maybe we could support, but OTOH it's still in "testing" phase, so we don't need to right now

so, it looks like currently these two are all we can support.

Most likely we'll need to update this in the future, but right now these instructions LGTM :ok_hand:

docs/distros/ubuntu.md
```diff
@@ -30,6 +30,25 @@ As an addition, `tezos-baking` package provides `tezos-baking-<network>` service
...
+sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 37B8819B7D0D183812DCA9A8CE5A4D8933AE7CBB
+sudo apt-get update
+```
+If packages for `bionic` are not suited for your debian version, see [related askubuntu thread](https://askubuntu.com/a/445496) to choose valid version.
```
pasqu4le, line 45: A small suggestion about phrasing (and line length):
```suggestion
If packages for `bionic` are not suited for your Debian version, see the
[related askubuntu thread](https://askubuntu.com/a/445496) to choose a valid one.
```

pasqu4le (approved):
LGTM, I only have one small suggestion left

PR #478, opened 2022-06-03
[#464] Support liquidity baking toggle vote

### LINKED ISSUE #464: Properly support the liquidity baking toggle vote option

Octez 13 introduces some [breaking changes to the baker](http://tezos.gitlab.io/CHANGES.html#baker), in particular, it:
> The `--liquidity-baking-toggle-vote` command-line option is now mandatory

Keeping in mind that the intent of making this required is to push bakers into making a decision, this requires some careful consideration and changes (some of which may have already been done for `jakartanet`):
1. the setup wizard should ask for this choice and use `pass` as the default (as it was the case already)
2. existing running `systemd` services should not break on a package update 
3. it would be nice to notify users about this option during a native package update (e.g. on `apt-get update`)

# Acceptance criteria

The now mandatory liquidity baking option is supported properly, considering the points above.

ixahedron:
## Description

Problem: in anticipation of `jakarta` on `mainnet`, we need to properly support the liquidity baking toggle vote option that is now mandatory. The Setup Wizard needs to ask for the user's preference.

Solution: added a step that queries the user's preference. There seems to be no good way to provide this option for our systemd service, so we modify the service's config file on each wizard invocation. Also added a post-install notification to `tezos-baking` .deb.

NB: another part of the issue is to make sure running systemd services can handle package upgrade. I was unable to break them, but perhaps it requires more consideration 🤔 

## Related issue(s)

Resolves #464

  can be successfully built.

### DIFF
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -577,7 +577,11 @@ class TezosBakingServicesPackage(AbstractPackage):
         custom_unit.poststop_script = "tezos-baking-custom-poststop"
         custom_unit.instances = []
         self.systemd_units.append(custom_unit)
-        self.postinst_steps = ""
+        # TODO: we will likely need to remove this once toggle vote isn't new anymore
+        self.postinst_steps = """echo "Please note that the liquidity baking toggle vote option"
+echo "is now mandatory when baking. You can read more about it here:"
+echo "https://tezos.gitlab.io/alpha/liquidity_baking.html#toggle-vote"
+"""
         self.postrm_steps = ""
 
     def fetch_sources(self, out_dir):
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -51,6 +51,12 @@ history_modes = {
     "archive": "Store all the chain data, very storage-demanding",
 }
 
+toggle_vote_modes = {
+    "pass": "Abstain from voting",
+    "off": "Request to end the subsidy",
+    "on": "Request to continue or restart the subsidy",
+}
+
 
 TMP_SNAPSHOT_LOCATION = "/tmp/tezos_node.snapshot"
 
@@ -124,6 +130,17 @@ systemd_mode_query = Step(
     "if you want to experiment. Enabling it will make it start on every boot.",
     options=systemd_enable,
     validator=Validator(enum_range_validator(systemd_enable)),
+)
+
+liquidity_toggle_vote_query = Step(
+    id="liquidity_toggle_vote",
+    prompt="Would you like to request to end the Liquidity Baking subsidy?",
+    help="Tezos chain offers a Liquidity Baking subsidy mechanism to incentivise exchange\n"
+    "between tez and tzBTC. You can ask to end this subsidy, ask to continue it, or abstain.\n"
+    "\nYou can read more about this in the here:\n"
+    "https://tezos.gitlab.io/alpha/liquidity_baking.html",
+    options=toggle_vote_modes,
+    validator=Validator(enum_range_validator(toggle_vote_modes)),
 )
 
 # We define this step as a function to better tailor snapshot options to the chosen history mode
@@ -408,6 +425,18 @@ class Setup(Setup):
             "to see the baker status and baking rights of your account."
         )
 
+    # There is no changing the toggle vote option at a glance,
+    # so we need to change the config every time
+    def set_toggle_vote(self):
+        self.query_step(liquidity_toggle_vote_query)
+
+        config_filepath = self.get_baking_config_filepath()
+        replace_in_service_config(
+            config_filepath,
+            "LIQUIDITY_BAKING_TOGGLE_VOTE",
+            f"\"{self.config['liquidity_toggle_vote']}\"",
+        )
+
     def start_baking(self):
         self.systemctl_start_action("baking")
 
@@ -446,6 +475,8 @@ class Setup(Setup):
                 print("Please check your input and try again.")
                 continue
             executed = True
+
+        self.set_toggle_vote()
 
         print()
         print("Starting the baking instance")
--- a/docker/package/wizard_structure.py
+++ b/docker/package/wizard_structure.py
@@ -363,6 +363,18 @@ def search_json_with_default(json_filepath, field, default):
         return json_dict.pop(field, default)
 
 
+def replace_in_service_config(config_filepath, field, value):
+    with open(config_filepath, "r") as f:
+        config_contents = f.read()
+
+    old = re.search(f"{field}=.*", config_contents)
+    if old is None:
+        return None
+    else:
+        new = f"{field}={value}"
+        proc_call(f"sudo sed -i 's/{old.group(0)}/{new}' {config_filepath}")
+
+
 class Step:
     def __init__(
         self,
@@ -584,7 +596,7 @@ class Setup:
             f"{self.config['tezos_client_options']} config update"
         )
 
-    def fill_baking_config(self):
+    def get_baking_config_filepath(self):
         net = self.config["network"]
         output = get_proc_output(f"systemctl show tezos-baking-{net}.service").stdout
         config_filepath = re.search(b"EnvironmentFiles=(.*) ", output)
@@ -596,7 +608,10 @@ class Setup:
             config_filepath = f"/etc/default/tezos-baking-{net}"
         else:
             config_filepath = config_filepath.group(1).decode().strip()
+        return config_filepath
 
+    def fill_baking_config(self):
+        config_filepath = self.get_baking_config_filepath()
         with open(config_filepath, "r") as f:
             config_contents = f.read()
             self.config["client_data_dir"] = search_baking_service_config(


docker/package/wizard_structure.py
```diff
@@ -363,6 +363,18 @@ def search_json_with_default(json_filepath, field, default):
...
+def replace_in_service_config(config_filepath, field, value):
+    with open(config_filepath, "r") as f:
+        config_contents = f.read()
+
+    old = re.search(f"{field}=.*", config_contents)
+    if old is None:
+        return None
+    else:
+        new = f"{field}={value}"
+        proc_call(f"sudo sed -i 's/{old.group(0)}/{new}' {config_filepath}")
```
rvem, line 375: I got the following `sed` error, when wizard attempted to update config file, seems like `/` is missing a the end.
```
You can check a blockchain explorer (e.g. https://tzkt.io/ or https://tzstats.com/)
to see the baker status and baking rights of your account.
Would you like to request to end the Liquidity Baking subsidy?
1. pass (default)              Abstain from voting
2. off                         Request to end the subsidy
3. on                          Request to continue or restart the subsidy
> 2
sed: -e expression #1, char 72: unterminated `s' command
The error has been logged to /home/roman/tezos_setup_wizard.log
```
```suggestion
        proc_call(f"sudo sed -i 's/{old.group(0)}/{new}/' {config_filepath}")
```

docker/package/tezos_setup_wizard.py
```diff
@@ -447,6 +476,8 @@ def run_setup(self):
                 continue
             executed = True
 
+        self.set_toggle_vote()
```
rvem, line 479: In some cases, this wizard can be run while the baking instance is already running. E.g. user wants to update its liquidity toggle vote. In that case, the baker service needs to be restarted manually to propagate changes, because config changes take effect only when service restarts

rvem: I guess it should be safe to restart the whole baking instance when the wizard finishes

docker/package/tezos_setup_wizard.py
```diff
@@ -408,6 +425,18 @@ def register_baker(self):
             "to see the baker status and baking rights of your account."
         )
 
+    # There is no changing the toggle vote option at a glance,
+    # so we need to change the config every time
```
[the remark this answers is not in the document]
rvem: > I suppose that the possible danger is that someone has already changed the config in ithaca to set a liquidity toggle vote, which was/is already possible even tho it's optional.

Currently, there is no convenient way to set liquidity vote within our services. And we cannot actually expect in which way users modified the baker service to use this flag :unamused:

docker/package/tezos_setup_wizard.py
```diff
@@ -126,6 +132,17 @@ def is_full_snapshot(import_mode):
     validator=Validator(enum_range_validator(systemd_enable)),
 )
 
+liquidity_toggle_vote_query = Step(
+    id="liquidity_toggle_vote",
+    prompt="Would you like to request to end the Liquidity Baking subsidy?",
+    help="Tezos chain offers a Liquidity Baking subsidy mechanism to incentivise exchange\n"
+    "between tez and tzBTC. You can ask to end this subsidy, ask to continue it, or abstain.\n"
+    "\nYou can read more about this in the here:\n"
+    "https://tezos.gitlab.io/alpha/liquidity_baking.html",
```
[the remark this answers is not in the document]
rvem: Alternatively, we can point to https://tezos.gitlab.io/jakarta/liquidity_baking.html and update it later if needed

docker/package/tezos_setup_wizard.py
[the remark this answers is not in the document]
pasqu4le: I think that you are right, there are too many different ways that one could have set this up manually, but none to do so conveniently with the wizard.

I think that, for users that went through that effort, all that we can do is try to make them notice that they need to update this. The post-install notification and possibly an announcement should be good enough.

However, we still need to make sure that (most) users that didn't change this option in ithaca will have a working setup without changing anything (but just updating), aka IMO the jakarta baker should have "pass" by default set in the config.

Likewise updating the baking docs I think it is still something that needs to be done.

docker/package/model.py
```diff
@@ -577,7 +577,11 @@ def __init__(
         custom_unit.poststop_script = "tezos-baking-custom-poststop"
         custom_unit.instances = []
         self.systemd_units.append(custom_unit)
-        self.postinst_steps = ""
+        # TODO: we will likely need to remove this once toggle vote isn't new anymore
```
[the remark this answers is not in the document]
ixahedron: I've opened #482.

docker/package/tezos_setup_wizard.py
```diff
@@ -447,6 +476,8 @@ def run_setup(self):
                 continue
             executed = True
 
+        self.set_liquidity_toggle_vote()
+
         print()
         print("Starting the baking instance")
         self.start_baking()
```
rvem, line 483: seems like `restart` argument is missing now

docker/package/tezos_setup_wizard.py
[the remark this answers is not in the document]
rvem: > aka IMO the jakarta baker should have "pass" by default set in the config

Yes, `pass` is already a default option, see [baking default config](https://github.com/serokell/tezos-packaging/blob/master/docker/package/defaults/tezos-baking.conf#L10)

docs/baking.md
```diff
@@ -155,6 +155,13 @@ sudo -u tezos tezos-client bootstrapped
 By default `tezos-baking-<network>.service` will use the `baker` alias for the
 key that will be used for baking and endorsing.
 
+### Setting the Liquidity Baking toggle vote option
+
+Since `013-PtJakart`, the `--liquidity-baking-toggle-vote` command line option for
+`tezos-baker` is now mandatory. In our systemd services, it is set it to `pass` by
+default. You can change it as desired in the service config file at
+`/etc/defaults/tezos-baking.conf`.
```
rvem, line 163:
```suggestion
`/etc/defaults/tezos-baking-mainnet`.
```
Defaults files are network specific

docs/baking.md
```diff
@@ -155,6 +155,13 @@ sudo -u tezos tezos-client bootstrapped
 By default `tezos-baking-<network>.service` will use the `baker` alias for the
 key that will be used for baking and endorsing.
 
+### Setting the Liquidity Baking toggle vote option
```
rvem, line 158: There are two additional things that I think need to be mentioned in this section:
1) Once the default file is updated baking service needs to be restarted for changes to take effect
2) It's also possible to re-run `tezos-setup-wizard` and change vote there, and I think that this option is a bit more convenient than editing a file and restarting services manually

rvem:
I have one additional UX concern that I want to discuss. Currently, the setup wizard asks to set the liquidity baking toggle vote on the Ithaca protocol as well. However, voting on Ithaca has a different format.
I think it's fine to transform Jakarta-specific format to Ithaca-specific, but I think that this needs to be documented, i.e. that `off` vote is transformed to `--liquidity-baking-escape-vote`, while `pass` and `on` will take effect only when Jakarta is activated.
What do you think?

docker/package/model.py
```diff
@@ -577,7 +577,11 @@ def __init__(
...
+        self.postinst_steps = """echo "Please note that the liquidity baking toggle vote option"
+echo "is now mandatory when baking. You can read more about it here:"
+echo "https://tezos.gitlab.io/alpha/liquidity_baking.html#toggle-vote"
```
pasqu4le, line 583:
```suggestion
echo "https://tezos.gitlab.io/jakarta/liquidity_baking.html#toggle-vote"
```

docs/baking.md
```diff
@@ -155,6 +155,13 @@ sudo -u tezos tezos-client bootstrapped
...
+
+Since `013-PtJakart`, the `--liquidity-baking-toggle-vote` command line option for
+`tezos-baker` is now mandatory. In our systemd services, it is set it to `pass` by
```
pasqu4le, line 161:
```suggestion
`tezos-baker` is now mandatory. In our systemd services, it is set to `pass` by
```

docker/package/model.py
```diff
@@ -494,7 +494,7 @@ class TezosBakingServicesPackage(AbstractPackage):
     # native releases, so we append an extra letter to the version of
     # the package.
     # This should be reset to "" whenever the native version is bumped.
-    letter_version = ""
+    letter_version = "a"
```
pasqu4le, line 497: It would be nice to have the usual `Problem`/`Solution` in this last commit message, but other than that this LGTM now

docs/baking.md
```diff
@@ -155,6 +155,26 @@ sudo -u tezos tezos-client bootstrapped
...
+Since `013-PtJakart`, the `--liquidity-baking-toggle-vote` command line option for
+`tezos-baker` is now mandatory. In our systemd services, it is set to `pass` by
+default. You can change it as desired in the service config file at
+`/etc/defaults/tezos-baking-<network>`.
+
+If the baking service is running already, changing this options requires a restart.
+You can do so manually by running:
+```bash
+sudo systemctl restart tezos-baking-<network>.service
+```
```
pasqu4le, line 169: I just realized that one piece of documentation is missing: in the `docs/tests/voting.md` file there is an example of the service config file to edit containing:
```
BAKER_ADDRESS_ALIAS="baker"
CUSTOM_NODE_CONFIG="<paste your path to the node config here>"
RESET_ON_STOP=""
```
which is missing the `LIQUIDITY_BAKING_TOGGLE_VOTE` line

PR #480, opened 2022-06-08
Add option to create packages from pre-built static binaries

z8v:
## Description 

Currently there is no option to use pre-built static binaries on the Debian/Fedora packages. Creating the packages will always involve building from scratch. This might not be desirable when there is another CI process that builds the binaries.

## Related issue(s)

None

  can be successfully built.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -43,6 +43,24 @@ Once this is done you should run the following command to build `aarch64` static
 ./docker-static-build.sh aarch64
 ```
 
+### Packages from statically linked binaries
+
+It's possible to create packages (ubuntu/debian & fedora) using already
+built binaries. This means that the build step of the package will be reduced
+to copying the binaries instead of building again from scratch.
+
+Using Ubuntu/Debian and assuming all Tezos binaries are located in the `binaries` directory:
+```sh
+cd .. && ./docker/docker-tezos-packages.sh --os ubuntu --type binary --binaries-dir binaries
+```
+
+Using Fedora and assuming all Tezos binaries are located in the `binaries` directory:
+```sh
+cd .. && ./docker/docker-tezos-packages.sh --os fedora --type binary --binaries-dir binaries
+```
+
+The resulting packages will be located in `../out` directory.
+
 ## Ubuntu packages
 
 We provide a way to build both binary and source native Ubuntu packages.
@@ -142,7 +160,7 @@ dput tezos-rc-serokell ../out/<package>.changes
 
 #### Updating release in scope of the same upstream version
 
-In case you're uploading the same version of the package but with the a different
+In case you're uploading the same version of the package but with a different
 release number, you'll highly likely have to use the same source archive (`.orig.tar.gz` archive)
 that was used for the first release in the scope of the same version, it can be downloaded from
 the launchpad package details (e.g. https://launchpad.net/~serokell/+archive/ubuntu/tezos/+sourcefiles/tezos-client/2:7.4-0ubuntu2/tezos-client_7.4.orig.tar.gz).
--- a/docker/package/package_generator.py
+++ b/docker/package/package_generator.py
@@ -4,8 +4,66 @@
 import os, shutil, argparse
 
 from .fedora import build_fedora_package
-from .packages import packages
+from .packages import packages, sapling_package
 from .ubuntu import build_ubuntu_package
+
+
+def get_ubuntu_run_deps(args):
+    """
+    List all the ubuntu run dependencies. Return an empty list when using prebuilt static binaries.
+    """
+    if args.binaries_dir:
+        return []
+
+    return [
+        "libev-dev",
+        "libgmp-dev",
+        "libhidapi-dev",
+        "libffi-dev",
+        "zlib1g-dev",
+        "libpq-dev",
+    ]
+
+
+def get_fedora_run_deps(args):
+    """
+    List all the fedora run dependencies. Return an empty list when using prebuilt static binaries.
+    """
+    if args.binaries_dir:
+        return []
+
+    return [
+        "libev-devel",
+        "gmp-devel",
+        "hidapi-devel",
+        "libffi-devel",
+        "zlib-devel",
+        "libpq-devel",
+    ]
+
+
+def get_build_deps(args):
+    """
+    List all the common build dependencies. Return an empty list when using prebuilt static binaries.
+    """
+    if args.binaries_dir:
+        return ["make", "wget"]
+
+    return [
+        "make",
+        "m4",
+        "perl",
+        "pkg-config",
+        "wget",
+        "unzip",
+        "rsync",
+        "gcc",
+        "cargo",
+        "opam",
+        "git",
+        "autoconf",
+    ]
+
 
 is_ubuntu = None
 is_source = None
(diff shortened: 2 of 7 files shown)


rvem:
Out of curiosity, what is the use-case for packages with pre-built static binaries?

docker/package/packages.py
```diff
@@ -411,6 +411,12 @@ def mk_node_unit(
         )
     )
 
+packages.append(
```
rvem, line 414: `TezosSaplingParamsPackage` was removed from `packages` on purpose to avoid re-publishing it by CI, since it isn't supposed to be updated. If you want to have a way to build it, I propose to add a conditional for it

docker/README.md
```diff
@@ -43,6 +43,32 @@ Once this is done you should run the following command to build `aarch64` static
...
+built binaries. This means that the build step of the package will be reduced
+to copying the binaries instead of building again from scratch.
+
+Using Ubuntu/Debian and assuming all Tezos binaries are located in the `out` directory:
+```sh
+# DEB_BUILD_OPTIONS=nostrip is needed if the provided binaries have already
+# been stripped of any debug symbols.
+
+docker build -t tezos-ubuntu -f docker/package/Dockerfile-ubuntu .
+docker run -it -e TEZOS_VERSION=v13.0 \
```
rvem, line 58: Tried following these instructions, got the following:
```
docker run -it -e TEZOS_VERSION=v13.0 \
               -e DEB_BUILD_OPTIONS=nostrip \
               -v `pwd`:/tezos-packaging tezos-ubuntu \
               --os ubuntu --type binary --binaries-dir ./out            
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/tezos-packaging/docker/package/package_generator.py", line 122, in <module>
    build_ubuntu_package(
  File "/tezos-packaging/docker/package/ubuntu.py", line 28, in build_ubuntu_package
    pkg.fetch_sources(dir, binaries_dir)
  File "/tezos-packaging/docker/package/model.py", line 228, in fetch_sources
    os.makedirs(out_dir)
  File "/usr/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: 'tezos-client-13.0'
```
Am I missing something?

docker/package/package_generator.py
```diff
@@ -7,6 +7,61 @@
...
+
+def get_build_deps(args):
+    """
+    List all the common build dependencies. Return an empty list when using prebuilt static binaries.
+    """
+    if args.binaries_dir:
+        return []
+
+    return [
+        "make",
```
rvem, line 51: I believe that `make` is still needed for the case when static binaries are used

docker/README.md
rvem, line 58: Does this command suppose to be run from the repo's root from the `docker` directory? For some reason I can't get it to work in both cases :disappointed:

docker/README.md
```diff
@@ -43,6 +43,32 @@ Once this is done you should run the following command to build `aarch64` static
...
+               -e DEB_BUILD_OPTIONS=nostrip \
+               -v `pwd`:/tezos-packaging tezos-ubuntu \
```
rvem, line 60: I slightly dislike the idea of mounting the whole directory as a volume because build artifacts and intermediate build directories pollute it. Does it make sense to mount only the directory with the binaries and the directory with the build artifacts?

rvem: This will also probably help to solve the issues I've mentioned above

docker/package/packages.py
z8v: Makes sense, I add a boolean flag for this. It's disabled by default.

docker/package/package_generator.py
z8v: Yes, I've also included `wget` for the sapling package.

docker/README.md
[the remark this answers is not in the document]
z8v: I've added an extra argument to make the output dir configurable and updated the README accordingly.

docker/README.md
z8v: Yep, I've updated the example.

z8v:
> Out of curiosity, what is the use-case for packages with pre-built static binaries?

We're trying to integrate the packaging scripts in the tezos repo where there is already a process in place which builds the static binaries.

rvem:
Do you intend to include resulting `.deb` packages into `tezos/tezos` releases?

docker/README.md
```diff
@@ -43,6 +43,34 @@ Once this is done you should run the following command to build `aarch64` static
 ./docker-static-build.sh aarch64
 ```
 
+### Packages from statically linked binaries
+
+It's possible to create packages (ubuntu/debian & fedora) using already
+built binaries. This means that the build step of the package will be reduced
+to copying the binaries instead of building again from scratch.
+
+Using Ubuntu/Debian and assuming all Tezos binaries are located in the `out` directory:
```
rvem, line 52: This comment looks outdated now, same for Fedora packages
```

**Comment by rvem** [comment]
CI currently cannot be started from forks, so I moved your changes and pushed them. See [CI run details](https://buildkite.com/serokell/tezos-packaging/builds/2693)

**Comment by rvem** [comment]
I also attempted to build source packages made of static binaries using both `pbuilder-dist` locally and my private PPA. Unfortunately, both failed :disappointed:.

See [launchpad build log](https://launchpadlibrarian.net/606845060/buildlog_ubuntu-focal-amd64.tezos-client_2%3A13.0-0ubuntu1~focal_BUILDING.txt.gz) for more details about the build error

File: docker/package/packages.py (hunk 1 above)
Comment (rvem) [reply docker/package/packages.py]: Note that there is also `tezos-baking` package, which doesn't actually package any Octez binaries, but packages systemd services and some python scripts. Not sure if you intend to put it into `tezos/tezos` releases as well since this package tends to update multiple times between upstream releases

File: docker/README.md (hunk 5 above)
Comment (z8v) [reply docker/README.md]: Thanks, I updated the docs.

File: docker/package/packages.py (hunk 1 above)
Comment (z8v) [reply docker/package/packages.py]: I see that `tezos-baking` is already built.

**Comment by z8v** [comment]
I see that it errors at the following:
```
cp /<<PKGBUILDDIR>>/tezos-client /usr/bin
cp: cannot create regular file '/usr/bin/tezos-client': Permission denied
```
which probably means `sudo` is required, although I don't see any `sudo` reference on the default Makefile or in the `build-binary.sh` script.

**Comment by rvem** [comment]
I have an impression that it's caused by the fact that directories mounted to the docker container belong to `root` by default and `chown`ing `--output-dir` doesn't help

File: docker/package/packages.py (hunk 1 above)
Comment (z8v) [reply docker/package/packages.py]: Yep, I don't see a reason against it.

**Comment by z8v** [comment]
> I have an impression that it's caused by the fact that directories mounted to the docker container belong to `root` by default and `chown`ing `--output-dir` doesn't help

Is there a script with `pbuilder-dist` that I can use to reproduce this locally?

**Comment by rvem** [comment]
I did the following:
```
$ docker run -it -e TEZOS_VERSION=v13.0 \
               -e DEB_BUILD_OPTIONS=nostrip \
               -v `pwd`/binaries:/binaries \
               -v `pwd`/out:/out \
               tezos-ubuntu --os ubuntu --type source --package tezos-client --binaries-dir /binaries --output-dir /out
$ sudo chown -R <user>:users out
$ pbuilder-dist focal build out/tezos-client_13.0-0ubuntu1~focal.dsc
```

z8v:
@rvem I've pushed your commit with the docker fix into the PR.

rvem:
Nice, could you please squash the fixup and I'll run the CI?

rvem (approved):

PR #481, opened 2022-06-10
[#476] Flakify tezos-packaging

### LINKED ISSUE #476: Flakify tezos-packaging
Currently, tezos-packaging project is using `niv` to manage dependencies.

However, `nix` provides a native way for convenient dependencies handling via flakes.

# Acceptance criteria
* Repo uses nix flakes to handle dependencies.
* CI configuration is updated accordingly (including automatic Octez updates and release PR creation).

krendelhoff:
## Description

Problem: tezos-packaging uses niv

Solution: switch to using flakes

## Related issue(s)

Resolves #476

  can be successfully built.


flake.nix
```diff
@@ -0,0 +1,77 @@
...
+    serokell-nix = {
+      url = "github:serokell/serokell.nix?rev=46d762e5107d10ad409295a7f668939c21cc048d";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+    flake-compat = {
+      url = "github:edolstra/flake-compat";
+      inputs.nixpkgs.follows = "nixpkgs";
+      flake = false;
+    };
+    crate2nix = {
```
rvem, line 22: I think it's no longer used, actually

release.nix
```diff
@@ -1,36 +1,17 @@
-# SPDX-FileCopyrightText: 2021 Oxhead Alpha
-# SPDX-License-Identifier: LicenseRef-MIT-OA
+# SPDX-FileCopyrightText: 2019 TQ Tezos <https://tqtezos.com/>
```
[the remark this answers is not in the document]
krendelhoff: release.nix now tezos-release.nix, maybe temporarily

balsoft:
Just FYI, there's a flake in my branch: https://github.com/serokell/tezos-packaging/blob/new-opam-nix/flake.nix

flake.nix
```diff
@@ -0,0 +1,83 @@
...
+    };
+    opam-repository = {
+      url = "git+https://gitlab.com/tezos/opam-repository?rev=845375388d477153b36344f69a23a1b34b2bf82e";
+      flake = false;
+    };
+    tezos = {
+      url = "git+https://gitlab.com/tezos/tezos?rev=cb9f439e58c761e76ade589d1cdbd2abb737dc68";
+      flake = false;
+    };
+  };
```
balsoft, line 34: 1. No need to specify commit hashes here
2. There's a `gitlab:` schema for gitlab, e.g. `gitlab:tezos/tezos`
3. You can use [serokell's flake-registry](https://github.com/serokell/flake-registry) to reduce the number of inputs here

nix/build/ocaml-overlay.nix
```diff
@@ -17,7 +16,7 @@ with opam-nix.lib.${pkgs.system}; let
     resolveArgs = { };
   } tezosSourcesResolved { };
 in {
-  ocamlPackages = tezosScope.overrideScope' (pkgs.lib.composeManyExtensions [
+  ocamlPackages = (tezosScope.overrideScope' (pkgs.lib.composeManyExtensions [
```
balsoft, line 19: Why not just `tezosPackages = ...` here and then not have the awkward `// super.ocamlPackages` ?

balsoft (changes requested):

.buildkite/pipeline-for-tags.yml
```diff
@@ -5,7 +5,7 @@ steps:
  - label: Check for new Tezos release
    if: build.source == "schedule" && build.branch == "master"
    commands:
-   - nix-shell ./scripts/shell.nix --run ./scripts/update-tezos.sh
+   - nix-unstable develop .#autorelease -c ./scripts/update-tezos.sh
```
rvem, line 8: I think it's better to use the usual `nix` and `USE_NEWER_NIX=1` env variable so that we'll have to only remove this env variable once `nix` is updated build servers.

rvem, line 8: I think it's better to use the usual `nix` and `USE_NEWER_NIX=1` env variable so that we'll have to only remove this env variable once `nix` is updated build servers.

flake.nix
```diff
@@ -0,0 +1,87 @@
...
+
+      tezos-release = callPackage ./release.nix {};
+
+    in rec {
+
+      legacyPackages = pkgs;
+
+      inherit tezos-release;
+
+      apps.tezos-client = {
```
rvem, line 66: Why only `tezos-client`? :thinking: 

Are these `apps` some default flake outputs?

rvem, line 66: Why only `tezos-client`? :thinking: 

Are these `apps` some default flake outputs?

nix/modules/tezos-node.nix
```diff
@@ -72,7 +71,19 @@ let
 
       nodeConfig = mkOption {
         default = null;
-        type = types.nullOr (import sources.serokell-nix).lib.types.jsonConfig;
```
rvem, line 75: Were these types removed from `serokell-nix`?

rvem, line 75: Were these types removed from `serokell-nix`?

flake.nix
krendelhoff: apps is thing to run with nix run, do we need it at all?

nix/modules/tezos-node.nix
krendelhoff: I failed to understand how to pass argument to nixos module, actually I just copied definition so it works fine till you help with proper solution

flake.nix
rvem: I don't think that we really need it

nix/modules/tezos-node.nix
```diff
@@ -1,14 +1,13 @@
 # SPDX-FileCopyrightText: 2021 Oxhead Alpha
 # SPDX-License-Identifier: LicenseRef-MIT-OA
 
-{config, lib, pkgs, ...}:
+{ config, lib, pkgs, ... }:
```
rvem, line 4: I think you can add `inputs` here and then use `inputs.serokell-nix` below for the `jsonConfig` type

scripts/update-input.py
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: 2022 Oxhead Alpha
```
rvem, line 1: Does it make sense for you to add a shebang `#!/usr/bin/env python3` and use this script as an executable?

scripts/update-tezos.sh
```diff
@@ -39,12 +37,10 @@ if [[ "$latest_upstream_tag" != "$our_tezos_tag" ]]; then
     git switch -c "$branch_name"
     echo "Updating Tezos to $latest_upstream_tag"
 
-    cd nix
-    niv update tezos -a ref="refs/tags/$latest_upstream_tag" -a rev="$latest_upstream_tag_hash"
-    niv update opam-repository -a rev="$opam_repository_tag" -a ref="$opam_repository_branch" -b "$opam_repository_branch"
+    python3 ./scripts/update-input.py tezos "$latest_upstream_tag_hash"
+    python3 ./scripts/update-input.py opam-repository "$opam_repository_tag"
```
rvem, line 41: This one didn't work for me for some reason when I tried to simulate an update from `v12.3` to `v13.0` locally. Though `tezos` in flake.nix was updated correctly. Also, seems like `flake.lock` still requires an update. I guess the best way to do this is to run `nix flake lock --update-input tezos` and `nix flake lock --update-input opam-repository` afterwards

scripts/update-tezos.sh
```diff
@@ -24,13 +24,11 @@ source scripts/version.sh
...
-git clone --bare https://gitlab.com/tezos/opam-repository.git upstream-opam-repository
-opam_repository_branch="$(git --git-dir=upstream-opam-repository for-each-ref --contains "$opam_repository_tag" --count=1 --sort='-refname' --format="%(refname:short)" refs/)"
-rm -rf upstream-opam-repository
-
 branch_name="auto/$latest_upstream_tag-release"
 
-our_tezos_tag="$(jq -r '.tezos.ref' nix/nix/sources.json | cut -d'/' -f3)"
+our_tezos_tag="$(jq -r '.tezos_ref' meta.json | cut -d'/' -f3)"
+
+jq ".tezos_ref=$latest_upstream_tag" meta.json | tee meta.json > /dev/null
```
rvem, line 31: Failed for me locally, seems like quotes are missing
```suggestion
jq ".tezos_ref=\"$latest_upstream_tag\"" meta.json | tee meta.json > /dev/null
```

nix/modules/tezos-node.nix
krendelhoff: I don't think it's possible. But it could be easily added to pkgs, I just did it the wrong way

balsoft:
BTW, after [this bugfix](https://github.com/NixOS/nix/pull/6662) makes it to a release, it will be possible to use `nix flake lock --override-input tezos gitlab:tezos/tezos/v<release>` to update the `flake.lock` file. May I suggest postponing merging this until then, to prevent hacky sed-ding scripts from being merged?

rvem:
> May I suggest postponing merging this until then, to prevent hacky sed-ding scripts from being merged?

I like this idea

> BTW, after this bugfix makes it to a release

And another half an eternity for us to start using this nix version in CI :see_no_evil:

.buildkite/pipeline-for-tags.yml
```diff
@@ -5,7 +5,7 @@ steps:
  - label: Check for new Tezos release
    if: build.source == "schedule" && build.branch == "master"
    commands:
-   - nix-shell ./scripts/shell.nix --run ./scripts/update-tezos.sh
+   - nix develop .#autorelease -c ./scripts/update-tezos.sh
```
rvem, line 8: I believe that we need to use `USE_NEWER_NIX` in this file as well

rvem:
Looks good, but I think we should wait a bit and use native `nix` mechanism to update non-flake inputs

.buildkite/pipeline-raw.yml
```diff
@@ -2,31 +2,32 @@
...
  - wait
 
  - label: reuse lint
-   command: nix run -f . pkgs.reuse -c reuse lint
+   command: nix shell .#reuse -c reuse lint
  - label: check trailing whitespace
    command: .buildkite/check-trailing-whitespace.sh
  - label: crossref-verify
-   command: "nix run -f https://github.com/serokell/crossref-verifier/archive/68a1f9d25b6e7835fea8299b18a3e6c61dbb2a5c.tar.gz -c crossref-verify"
+   command: "USE_NEWER_NIX=0; nix run -f https://github.com/serokell/crossref-verifier/archive/68a1f9d25b6e7835fea8299b18a3e6c61dbb2a5c.tar.gz -c crossref-verify"
```
rvem, line 23: While we're at it, could you please update this outdated piece of CI pipeline as well? Apparently, [xrefcheck](https://github.com/serokell/xrefcheck) should be added as another flake input and used here instead of `nix run`ing specific revision of old repo

scripts/update-tezos.sh
```diff
@@ -39,12 +38,10 @@ if [[ "$latest_upstream_tag" != "$our_tezos_tag" ]]; then
     git switch -c "$branch_name"
     echo "Updating Tezos to $latest_upstream_tag"
 
-    cd nix
-    niv update tezos -a ref="refs/tags/$latest_upstream_tag" -a rev="$latest_upstream_tag_hash"
-    niv update opam-repository -a rev="$opam_repository_tag" -a ref="$opam_repository_branch" -b "$opam_repository_branch"
+    nix flake lock --override-input tezos "gitlab:tezos/tezos/$latest_upstream_tag_hash"
```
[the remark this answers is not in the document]
krendelhoff: `nix flake lock` update is sticky now, so i think we could just update to relevant revisions and left inputs generic

scripts/shell.nix
```diff
@@ -1,11 +1,11 @@
...
 # SPDX-License-Identifier: LicenseRef-MIT-OA
 
-{ pkgs ? import (import ../nix/nix/sources.nix {}).nixpkgs { } }:
+{ pkgs, nix, ...}:
 with pkgs;
 mkShell {
   buildInputs = [
     coreutils gnused gh git rename gnupg dput rpm debian-devscripts which util-linux perl
-    jq niv
+    jq python3 nix
```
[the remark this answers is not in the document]
krendelhoff2: I decided to use it for better UX

scripts/update-input.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
```
rvem, line 1: Is this script used now?

rvem (approved):
I have a single nitpick, the rest looks good now

balsoft (approved):

scripts/update-input.py
krendelhoff2: I forgot to use it in autorelease, will fix it now

PR #492, opened 2022-07-06
[#491] 'setup ledger to bake for' only for ledger key

### LINKED ISSUE #491: [BUG] Setting up baking with remote signer does not work
# Description

tezos-setup-wizard exits with an error when importing remote signer url as the baker key

# To Reproduce

- run tezos-setup-wizard
- follow the prompts to set up baking, select "remote" for baker key import
- type/paste remote signer url
- observe error

```
How do you want to import the baker key?
1. ledger (default)            From a ledger
2. secret-key                  Either the unencrypted or password-encrypted secret key for
                               your address
3. remote                      Remote key governed by a signer running on a different
                               machine
4. json                        Faucet JSON file from https://teztnets.xyz/
> 3
Provide your remote key with the address of the signer.
> http://192.168.64.1:6732/tz1SFbdg2jjkixHNU1Jm9f8JA4pYnMXsgATC
Tezos address added: tz1SFbdg2jjkixHNU1Jm9f8JA4pYnMXsgATC
Error:
  Erroneous command line argument 6 (baker).
  Cannot parse host of URI: remote:tz1SFbdg2jjkixHNU1Jm9f8JA4pYnMXsgATC

Usage:
  tezos-client [global options] command [command options]
  tezos-client --help (for global options)
  tezos-client [global options] command --help (for command options)
  tezos-client --version (for version information)

To browse the documentation:
  tezos-client [global options] man (for a list of commands)
  tezos-client [global options] man -v 3 (for the full manual)

Global options (must come before the command):
  -d --base-dir <path>: client data directory (absent: TEZOS_CLIENT_DIR env)
  -c --config-file <path>: configuration file
  -t --timings: show RPC request times
  --chain <hash|tag>: chain on which to apply contextual commands (commands dependent on the context associated with the specified chain). Possible tags are 'main' and 'test'.
  -b --block <hash|level|tag>: block on which to apply contextual commands (commands dependent on the context associated with the specified block). Possible tags include 'head' and 'genesis' +/- an optional offset (e.g. "tezos-client -b head-1 get timestamp"). Note that block queried must exist in node's storage.
  -w --wait <none|<int>>: how many confirmation blocks are needed before an operation is considered included
  -p --protocol <hash>: use commands of a specific protocol
  -l --log-requests: log all requests to the node
  --better-errors: Error reporting is more detailed. Can be used if a call to an RPC fails or if you don't know the input accepted by the RPC. It may happen that the RPC calls take more time however.
  -A --addr <IP addr|host>: [DEPRECATED: use --endpoint instead] IP address of the node
  -P --port <number>: [DEPRECATED: use --endpoint instead] RPC port of the node
  -S --tls: [DEPRECATED: use --endpoint instead] use TLS to connect to node.
  -m --media-type <json, binary, any or default>: Sets the "media-type" value for the "accept" header for RPC requests to the node. The media accept header indicates to the node which format of data serialisation is supported. Use the value "json" for serialisation to the JSON format.
  -E --endpoint <uri>: HTTP(S) endpoint of the node RPC interface; e.g. 'http://localhost:8732'
  -s --sources <path>: path to JSON file containing sources for --mode light. Example file content: {"min_agreement": 1.0, "uris": ["http://localhost:8732", "https://localhost:8733"]}
  -R --remote-signer <uri>: URI of the remote signer
  -f --password-filename <filename>: path to the password filename
  -M --mode <client|light|mockup|proxy>: how to interact with the node
The error has been logged to /home/ubuntu/tezos_setup_wizard.log
```

```
$ cat /home/ubuntu/tezos_setup_wizard.log
Command '['sudo', '-u', 'tezos', 'TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=YES', 'tezos-client', '--base-dir', '/var/lib/tezos/.tezos-client', '--endpoint', 'http://localhost:8732', 'setup', 'ledger', 'to', 'bake', 'for', 'baker', '--main-hwm', '765952']' returned non-zero exit status 1.
```

# Expected behavior

Remote signer URL is imported as key without error

# Environment

- OS: Ubuntu 20.04
- branch/revision: 2:13.0b-0ubuntu3~focal

# Additional context

rvem:
## Description
Problem: Currently setup wizard attempts to call 'setup ledger to bake
for' for every key type. As a result, an attempt to setup remote or
secret-key baker fails with an error.

Solution: Call this function only when the selected key type is ledger.

## Related issue(s)

Resolves #491 

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -33,13 +33,6 @@ snapshot_import_modes = {
     "skip": "Skip snapshot import and synchronize with the network from scratch",
 }
 
-key_import_modes = {
-    "ledger": "From a ledger",
-    "secret-key": "Either the unencrypted or password-encrypted secret key for your address",
-    "remote": "Remote key governed by a signer running on a different machine",
-    "json": "Faucet JSON file from https://teztnets.xyz/",
-}
-
 systemd_enable = {
     "yes": "Enable the services, running them both now and on every boot",
     "no": "Start the services this time only",
@@ -398,18 +391,21 @@ class Setup(Setup):
 
             key_mode_query = get_key_mode_query(key_import_modes)
 
-            ledger_set_up = False
-            while not ledger_set_up:
+            baker_set_up = False
+            while not baker_set_up:
                 self.import_key(key_mode_query, "Baking")
 
-                try:
-                    proc_call(
-                        f"sudo -u tezos {suppress_warning_text} tezos-client {tezos_client_options} "
-                        f"setup ledger to bake for {baker_alias} --main-hwm {self.get_current_head_level()}"
-                    )
-                    ledger_set_up = True
-                except PermissionError:
-                    print("Going back to the import mode selection.")
+                if self.config["key_import_mode"] == "ledger":
+                    try:
+                        proc_call(
+                            f"sudo -u tezos {suppress_warning_text} tezos-client {tezos_client_options} "
+                            f"setup ledger to bake for {baker_alias} --main-hwm {self.get_current_head_level()}"
+                        )
+                        baker_set_up = True
+                    except PermissionError:
+                        print("Going back to the import mode selection.")
+                else:
+                    baker_set_up = True
 
     def register_baker(self):
         print()


pasqu4le (approved):
LGTM and tested the changes, it seems to solve the issue :ok_hand:

PR #498, opened 2022-07-08
[#486] Ensure formula have the same version as current tag to update bottle

### LINKED ISSUE #486: Avoid CI from opening new formulae update PRs when no new bottle was built

Since #409 CI correctly uses the version specified in the formulae to check if `brew` bottles have been built already and then determine which ones need to be build (if any).

This allows us to make a packaging release for some distribution methods (e.g. Ubuntu) without re-building and re-pushing macOS bottles that already exist (only with a new release number).

However, there seems to be a flaw because, even when building no bottles, CI still opens a PR to update the formula hashes (see our [release workflow](https://github.com/serokell/tezos-packaging/blob/master/docs/release-workflow.md) for reference).

Example: #483 where the two commits correspond to the push of the [v13.0-2](https://github.com/serokell/tezos-packaging/tree/v13.0-2) and [v13.0-3](https://github.com/serokell/tezos-packaging/tree/v13.0-3) tag, respectively.

# Acceptance criteria

CI does not automatically open PRs to update bottle hashes in `brew` formulae if no new bottle was built.

rvem:
## Description
Problem: In some cases bottles' hashes are updated even when current
release hasn't updated any formulae.

Solution: Assert that the formula version as the same as current tag
before updating bottle hash.

I've tested the changes with `v13.0-test` tag, CI refused to push any changes:
* https://github.com/serokell/tezos-packaging/runs/7250322868
* https://buildkite.com/serokell/tezos-packaging-tags/builds/1887

## Related issue(s)

Resolves #486

  can be successfully built.

### DIFF
--- a/.github/workflows/build-bottles.yml
+++ b/.github/workflows/build-bottles.yml
@@ -89,9 +89,9 @@ jobs:
       - name: Download Catalina bottles from the release
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-        run: gh release download "${{ env.tag }}" -p "*.catalina.bottle.tar.gz" -D "./Catalina"
+        run: gh release download "${{ github.ref_name }}" -p "*.catalina.bottle.tar.gz" -D "./Catalina"
 
       - name: Add bottle hashes to formulae
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-        run: ./scripts/sync-bottle-hashes.sh "${{ env.tag }}" "Catalina"
+        run: ./scripts/sync-bottle-hashes.sh "${{ github.ref_name }}" "Catalina"
--- a/scripts/bottle-hashes.sh
+++ b/scripts/bottle-hashes.sh
@@ -2,7 +2,7 @@
 # SPDX-FileCopyrightText: 2021 Oxhead Alpha
 # SPDX-License-Identifier: LicenseRef-MIT-OA
 
-# This script takes a directory where the bottles are stored as its argument.
+# This script takes a directory where the bottles are stored and release tag as its arguments.
 # Run it from the base directory (tezos-packaging).
 
 set -e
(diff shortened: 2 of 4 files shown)


scripts/bottle-hashes.sh
```diff
@@ -19,10 +19,13 @@ then
                 os="${BASH_REMATCH[2]}"
                 formula_file="./Formula/$formula_name.rb"
                 if [[ -f $formula_file ]]; then
+                    formula_tag="$(sed -n "s/^\s\+version \"\(.*\)\"/\1/p" "$formula_file")"
                     line="\    sha256 cellar: :any, $os: \"$bottle_hash\""
                     # Update only when this formula doesn't have a hash for the bottle
-                    if ! grep "$line" "$formula_file" &> /dev/null; then
+                    if [[ "$formula_tag" == "$2" ]]; then
```
pasqu4le, line 25: Related to the point above: instead of checking this here, aka when trying to update the hashes, couldn't we check this earlier and avoid the bottle where this doesn't match from being built in the first place?

Not sure if it would be any better, in fact AFAIK the system that already checks for existing bottles should work quite well in preventing rebuilds, but I wanted to mention/ask about it

rvem: The problem is that GA spawns a job per bottle and there is already a check that prevents building the bottle in case it was already built. And IMO such an approach works a bit better when there are some flaky build errors

.github/workflows/build-bottles.yml
```diff
@@ -94,4 +94,4 @@ jobs:
       - name: Add bottle hashes to formulae
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-        run: ./scripts/sync-bottle-hashes.sh "${{ env.tag }}" "Catalina"
+        run: ./scripts/sync-bottle-hashes.sh "${{ github.ref_name }}" "Catalina"
```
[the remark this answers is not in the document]
rvem: Indeed, 'Download Catalina bottles from the release' should probably also use `github.ref_name`. Also, this way it'll be consistent with BK pipeline

PR #501, opened 2022-07-17
[Chore] autorelease.sh hotfix

krendelhoff2:
## Description

Problem: release derivation wasn't built due to incorrect
flake output declaration.

Solution: Get rid of misleading flake-utils functions.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/.buildkite/pipeline-raw.yml
+++ b/.buildkite/pipeline-raw.yml
@@ -87,7 +87,7 @@ steps:
    commands:
    - buildkite-agent artifact download "docker/*" . --step "build-via-docker"
    - chmod +x ./docker/*
-   - nix-build . -A binaries-test.x86_64-linux --no-out-link --arg path-to-binaries ./docker
+   - nix-build . -A binaries-test --no-out-link --arg path-to-binaries ./docker
    branches: "!master"
    depends_on:
     - "build-via-docker"
--- a/.buildkite/pipeline-raw.yml
+++ b/.buildkite/pipeline-raw.yml
@@ -87,7 +87,7 @@ steps:
    commands:
    - buildkite-agent artifact download "docker/*" . --step "build-via-docker"
    - chmod +x ./docker/*
-   - nix-build . -A binaries-test --no-out-link --arg path-to-binaries ./docker
+   - nix-build . -A binaries-test.x86_64-linux --no-out-link --arg path-to-binaries ./docker
    branches: "!master"
    depends_on:
     - "build-via-docker"
--- a/.buildkite/pipeline-raw.yml
+++ b/.buildkite/pipeline-raw.yml
@@ -170,7 +170,7 @@ steps:
    - buildkite-agent artifact download "docker/*" arm-binaries --step "build-arm-via-docker"
    - ls binaries
    - ./scripts/autorelease.sh "$BUILDKITE_MESSAGE"
-   branches: master
+   #branches: master
    depends_on:
     - "build-via-docker"
     - "build-arm-via-docker"
--- a/scripts/autorelease.sh
+++ b/scripts/autorelease.sh
@@ -21,7 +21,7 @@ trap finish EXIT
 assets_dir=$TEMPDIR/assets
 
 # Build tezos-release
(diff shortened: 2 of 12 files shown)


scripts/autorelease.sh
```diff
@@ -21,7 +21,7 @@ trap finish EXIT
 assets_dir=$TEMPDIR/assets
 
 # Build tezos-release
-nix-build . -A tezos-release -o "$TEMPDIR"/"$project" \
+nix-build . -A release -o "$TEMPDIR"/"$project" \
           --arg docker-binaries ./binaries/docker --arg docker-arm-binaries ./arm-binaries/docker
```
rvem, line 25: Looks like it should be `arm-docker-binaries` instead of `docker-arm-binaries` now

balsoft (changes requested):
Please just fix it with flake-utils instead. Hardcoding x86_64-linux is bad.

balsoft:
(also, the `release.nix` thing is just broken with flakes since it reads gitignored files)

krendelhoff2:
Wasn't it hardcoded before the same way? Also using nix-shell shebangs in scripts force us to use outputs with hardcoded x86_64-linux

balsoft:
> Wasn't it hardcoded before the same way?

AFAIR it wasn't.

```
$ gco 83f526f9e25a5f99e9fe5caa082ae9199d11a325^; rg x86_64-linux | wc -l
0
```

> Also using nix-shell shebangs in scripts force us to use outputs with hardcoded x86_64-linux

No. It can be fixed with a slightly differrent `shell.nix`, using `builtins.currentSystem` (since that's available when we're evaluating without flakes)

rvem:
> also, the `release.nix` thing is just broken with flakes since it reads gitignored files

What is the common way to deal with such problems? Create a dummy git repo on the fly and add all required files to it?

balsoft:
> What is the common way to deal with such problems? Create a dummy git repo on the fly and add all required files to it?

I'd say for now it's ok to use `nix build path:.#tezos-release` (or some other way to disable the git integration).

shell.nix
```diff
@@ -12,6 +12,8 @@ with pkgs; mkShell {
     python3Packages.black
     shellcheck
     vagrant
+    jq
+    buildkite-agent
   ];
-  TEZOS_VERSION="v13.1";
+  TEZOS_VERSION="v13.0";
```
rvem, line 18: can't we read `meta.json` here, btw?

rvem:
Btw, seems like _pre-flake_ `nix-build` works with gitignored files as well.
```nix-build . -A release -o "$TEMPDIR"/"$project" --arg docker-binaries ./binaries/docker --arg docker-arm-binaries ./arm-binaries/docker``` worked for me locally

**Comment by krendelhoff2** [comment]
@balsoft could you elaborate please about using of builtins.currentSystem

**Comment by balsoft** [comment]
With a `shell.nix` something like this:

```nix
let
  inherit (import
    (let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
    in fetchTarball {
      url =
        "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
      sha256 = lock.nodes.flake-compat.locked.narHash;
    }) { src = ./.; })
    shellNix;
in shellNix // shellNix.devShells.${__currentSystem}
```

you should be able to just do `nix-shell -A autorelease`

.buildkite/pipeline-raw.yml
```diff
@@ -170,7 +170,7 @@ steps:
    - buildkite-agent artifact download "docker/*" arm-binaries --step "build-arm-via-docker"
    - ls binaries
    - ./scripts/autorelease.sh "$BUILDKITE_MESSAGE"
-   branches: master
+   #branches: master
```
rvem, line 173: Don't forget to drop this change:)

balsoft (approved):

.buildkite/pipeline-raw.yml
balsoft: _spoiler: he did indeed forget to drop this change_

PR #506, opened 2022-07-26
[Chore] Expose 'overlay' from flake

rvem:
## Description
Problem: It's required to use specific overlay when using NixOS modules
from 'tezos-packaging' flake.

Solution: Expose 'overlay' that includes all required overlays.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -27,6 +27,12 @@
   outputs = inputs@{ self, nixpkgs, nixpkgs-unstable, flake-utils, serokell-nix, nix, ... }:
   let
     pkgs-darwin = nixpkgs-unstable.legacyPackages."aarch64-darwin";
+
+    protocols = nixpkgs.lib.importJSON ./protocols.json;
+    meta = nixpkgs.lib.importJSON ./meta.json;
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -39,13 +45,15 @@
       devShells."aarch64-darwin".autorelease-macos =
         import ./scripts/macos-shell.nix { pkgs = pkgs-darwin; };
 
+      overlay = final: prev: nixpkgs.lib.composeManyExtensions [
+        ocaml-overlay
+        (final: prev: { inherit (inputs) serokell-nix; })
```
MagicRB, line 50: that adds all of serokell-nix flake into the overlay, shouldnt it be `[ ocaml-overlay ]++ (mapAttrsToList (_: v: v) serokell-nix.overalays)`

MagicRB (changes requested):

flake.nix
rvem: The goal of this overlay is to provide types from `serokell.nix`, that aren't exposed in the overlay, AFAICS

balsoft (approved):

PR #507, opened 2022-07-27
[Chore] Fix 'build-binary.sh' script

rvem:
## Description

Problem: Native packages cannot be built on Launchpad due to the fact
build consumes too much disk space because of excessively copied Octez
sources. Also, in the latest release, all packages '.opam' files were
moved to a dedicated directory.

Solution:
1) Avoid installing packages during opam switch creation.
2) Add 'opam' directory to the list in which opam files are searched.

Also, provide a way to propagate this fix to the existing packages via patch file.

## Related issue(s)

Resolves #

  can be successfully built.


rvem:
Checked that packages can actually be built on launchpad https://launchpad.net/~rvem/+archive/ubuntu/tezos-experiments-1/+build/24204638

pasqu4le (approved):

PR #510, opened 2022-07-28
[Chore] Support multiple protocols simultaneously in daemon modules 

rvem:
## Description
Problem: In some cases, it's required to run bakers for two protocols
simultaneously.

Solution: Update daemons service config to take a list of protocols as
an option and run a daemon for each of listed protocols.

Currently based on #506 for convenience since both of these PRs provide changes needed in single place.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -27,6 +27,12 @@
   outputs = inputs@{ self, nixpkgs, nixpkgs-unstable, flake-utils, serokell-nix, nix, ... }:
   let
     pkgs-darwin = nixpkgs-unstable.legacyPackages."aarch64-darwin";
+
+    protocols = nixpkgs.lib.importJSON ./protocols.json;
+    meta = nixpkgs.lib.importJSON ./meta.json;
+    sources = { inherit (inputs) tezos opam-repository; };
+
+    ocaml-overlay = import ./nix/build/ocaml-overlay.nix (inputs // { inherit sources protocols meta; });
   in pkgs-darwin.lib.recursiveUpdate
   {
       nixosModules = {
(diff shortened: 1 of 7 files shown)


nix/modules/common.nix
```diff
@@ -20,12 +20,12 @@ rec {
 
   daemonOptions = sharedOptions // {
 
-    baseProtocol = mkOption {
-      type = types.enum [ "013-PtJakart" "014-PtKathma" ];
+    baseProtocols = mkOption {
+      type = types.listOf (types.enum [ "013-PtJakart" "014-PtKathma" ]);
```
MagicRB, line 24: I think there was a type for a list of unique elements, but im not sure

MagicRB: there wasnt

scripts/shell.nix
```diff
@@ -15,6 +15,7 @@ mkShell {
     rename
     gnupg
     dput
+    copr-cli
```
MagicRB, line 18: why are you adding this? its not mentioned anywhere else

flake.nix
```diff
@@ -39,13 +45,15 @@
       devShells."aarch64-darwin".autorelease-macos =
         import ./scripts/macos-shell.nix { pkgs = pkgs-darwin; };
 
+      overlay = final: prev: nixpkgs.lib.composeManyExtensions [
+        ocaml-overlay
+        (final: prev: { inherit (inputs) serokell-nix; })
```
MagicRB, line 50: same as the previous PR

MagicRB (changes requested):

scripts/shell.nix
rvem: Oh, look like I've added it accidentally

rvem: Removed

nix/modules/common.nix
rvem: I haven't found such type as well:(

balsoft (approved):

PR #543, opened 2022-10-19
[Chore] Symlink new and legacy names on fedora

krendelhoff2:
Problem: Fedora packages lack `tezos`-based executables.

Solution: Symlink `octez`-based binaries with `tezos` ones.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -323,7 +323,7 @@ Maintainer: {self.meta.maintainer}
 make {binary_name}
 mkdir -p %{{buildroot}}/%{{_bindir}}
 install -m 0755 {binary_name} %{{buildroot}}/%{{_bindir}}
-ln -s %{{_bindir}}/{binary_name} %{{buildroot}}/%{{_bindir}}/{self.name}
+ln -sf %{{_bindir}}/{binary_name} %{{buildroot}}/%{{_bindir}}/{self.name}
 {systemd_install}
 %files
 %license LICENSE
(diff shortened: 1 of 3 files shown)


pasqu4le (commented):
@krendelhoff2 since we'll need to make a new release for Fedora after this, can you please bump the version number too here?

docker/package/model.py
```diff
@@ -323,10 +323,12 @@ def gen_spec_file(self, build_deps, run_deps, out):
 make {binary_name}
 mkdir -p %{{buildroot}}/%{{_bindir}}
 install -m 0755 {binary_name} %{{buildroot}}/%{{_bindir}}
+ln -s %{{_bindir}}/{binary_name} %{{buildroot}}/%{{_bindir}}/{self.name}
```
pasqu4le, line 326: These work alright, but symlinks for `tezos-node-<protocol>` don't, as we discussed probably because they are not listed under `%files`.

For the time being let's add the `--force` to their `ln` invocation in the node post-installation

krendelhoff2: the problem is not inside that line

rvem: I guess it's a good idea to do `-sf` here as well, since this symlinking can be done for some subsequent version as well, and thus `ln -s` during the installation may fail since `/usr/bin/{self.name}` already exists

rvem (approved):
LGTM now

docker/package/model.py
krendelhoff2: force pushed since commit needed reword in presence of this suggestion

PR #547, opened 2022-10-24
Provide tezos packages for nix

### LINKED ISSUE #544: Provide `tezos`-based aliases for nix packages

Since `v15.0-rc1`, binaries and `opam` packages were renamed. Thus, we should do some work to provide backwards compatibility.

# Acceptance criteria
1.`tezos`-based symlinks to `octez`-based binaries are provided
2. There are `tezos`-based aliases to `octez`-based nix packages

krendelhoff2:
## Description

After rename, tezos-based packages are no longer available. However, they still heavily used in nix infrastructure, as well as 
legacy binary names.

This PR provide it.

## Related issue(s)

Resolves #544 

  can be successfully built.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -6,23 +6,22 @@
 let
   source = sources.tezos;
   release-binaries = import ./build/release-binaries.nix protocols;
-  binaries = builtins.listToAttrs (map (meta: {
+in {
+  octez-binaries = builtins.listToAttrs (map (meta: {
     inherit (meta) name;
     value = pkgs.octezPackages.${meta.name} // { inherit meta; };
   }) release-binaries);
 
-  # Bundle the contents of a package set together, leaving the original attrs intact
-  bundle = name: pkgSet:
-    pkgSet // (pkgs.buildEnv {
-      inherit name;
-      paths = builtins.attrValues pkgSet;
-    });
-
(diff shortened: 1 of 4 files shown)


nix/default.nix
```diff
@@ -6,23 +6,22 @@
...
-
-  artifacts = { inherit binaries; };
-  bundled = (builtins.mapAttrs bundle artifacts);
-
-in (d: with builtins;
-     listToAttrs (map (drv:
-       { name = drv; value = d.${drv}; }
-     ) (filter (name: substring 0 5 name == "octez") (attrNames d)))
-   ) bundled.binaries // { inherit (bundled) binaries; }
+  tezos-binaries = builtins.listToAttrs (map (meta:
```
rvem, line 15: What do you think about `trace`ing a deprecation warning here that binaries were renamed and one should switch to `octez-*`?

rvem (approved):

nix/default.nix
krendelhoff2: Since it used primarily in our CI, i don't see why we should do this

krendelhoff2: We shouldn't, since warning already raises for aliases, but packages still should be called as tezos-* (to be consistent with ubuntu/fedora packaging)

PR #554, opened 2022-10-28
[#545] Provide way to build fedora packages for different distributions

### LINKED ISSUE #545: Split `tezos-baking` package for different fedora versions

During binary package build,`python` version get pinned according to fedora version inside build environment. 
By now, every updates system-wide python version, and there is no easy way to provide legacy python toolchain, which makes impossible to install binary `rpm` package on fresh versions of fedora.

# Acceptance criteria
1. There is instance of `tezos-baking` binary package for every supported version of Fedora

krendelhoff2:
## Description

`tezos-baking` package seems to be non-portable across different fedora distributions, thus we should provide way to build packages for specific fedora distribution.

## Related issue(s)

Resolves #545 

  can be successfully built.


docker/package/packages.py
```diff
@@ -486,11 +507,13 @@ def mk_rollup_package(name, type, proto):
...
+        {f"tezos-{type}-rollup-{name}-{proto}": mk_rollup_package(name, type, proto)}
         for name in packages
         for type in types
         for proto in active_protocols
     ]
 
 
 packages.extend(mk_rollup_packages())
+
+packages = dict(ChainMap(*packages))
```
pasqu4le, line 524: This looks a bit odd to me, what's the purpose of using a `ChainMap` if it's immediately given to `dict` again?

gen_systemd_service_file.py
```diff
@@ -12,7 +12,7 @@
 
 if len(sys.argv) > 1:
     binary_name = sys.argv[1]
-    for package in packages:
+    for _, package in packages.items():
```
pasqu4le, line 15: Why not:
```suggestion
    for package in packages.values():
```

docker/package/packages.py
krendelhoff2: analogue of `foldr (//) empty`

docker/package/fedora.py
```diff
@@ -59,13 +61,39 @@ def build_fedora_package(
         build_deps + run_deps, run_deps, f"{home}/rpmbuild/SPECS/{pkg.name}.spec"
     )
```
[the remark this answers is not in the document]
pasqu4le: ok, please update the order then

it seems like it would be better and it would make the review easier too :slightly_smiling_face:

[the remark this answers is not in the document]
krendelhoff2: done

docker/package/fedora.py
```diff
@@ -59,13 +61,39 @@ def build_fedora_package(
...
+            )
+            if not is_source:
+                os.makedirs(f"{home}/rpmbuild/RPMS/x86_64", exist_ok=True)
+                subprocess.run(
+                    [
+                        "mock",
+                        "--resultdir",
+                        f"{home}/rpmbuild/RPMS/x86_64",
+                        "-r",
+                        f"/etc/mock/fedora-{dist}-x86_64.cfg",
```
rvem, line 87: Do these configs always exist?

docker/package/packages.py
```diff
@@ -441,8 +441,13 @@ def mk_node_unit(
         }
     )
 
-sapling_package = TezosSaplingParamsPackage(
-    meta=packages_meta, params_revision="95911b0639ff01807b8fd7b9e36d508e657d80a8"
+packages.append(
```
rvem, line 444: The problem is that this packages is now built by default, which means that it'll be uploaded to the Copr/Launchpad on the next `v*-1` release, while we don't really want to update it

krendelhoff2: What about adding `--release` flag and filter it out in `docker-tezos-packages.py`?

docker/package/fedora.py
krendelhoff2: Yes, it's stable paths if you fix host system version

docker/package/packages.py
rvem: There was a flag that explicitly enables this package building, I liked that approach

docker/package/packages.py
pasqu4le: :thinking: why not use a single dict and add single elements to it, or join other dicts with it, instead? :thinking:

docker/docker-tezos-packages.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+if os.getenv("USE_PODMAN", None):
+    virtualisation_engine = "podman"
+else:
+    virtualisation_engine = "docker"
+
+parser.add_argument(
+    "--output-dir",
+    help="provide a directory to place the built packages",
```
pasqu4le, line 56:
```suggestion
    help="provide a directory to place the built packages in",
```

docker/docker-tezos-packages.py
```diff
@@ -0,0 +1,168 @@
...
+else:
+    if args.distributions:
+        for dist in args.distributions:
+            if dist not in fedora_versions:
+                raise Exception(
+                    f"Distribution {dist} is not supported for {target_os}."
+                )
+        distributions = args.distributions
+    else:
+        distributions = ["native"]
```
pasqu4le, line 99: Two suggestions here:
1. make a function for the part that's not in the outermost `if`/`else` and call it twice
2. replace the outermost `else` with an `elif` to check that the target os is `fedora` and fail otherwise

docker/package/package_generator.py
```diff
@@ -66,113 +98,80 @@ def get_build_deps(args):
...
             )
+
+        if not is_source:
+            exts = [".deb"]
         else:
+            exts = [".orig.tar.gz", ".dsc", ".changes", ".debian.tar.xz", ".buildinfo"]
+
+        artifacts_dir = "."
+
+    else:
```
pasqu4le, line 151: Here too it's worth check if "fedora" was passed as an argument instead

docker/package/package_generator.py
```diff
@@ -66,113 +98,80 @@ def get_build_deps(args):
...
-    subdir = "SRPMS" if is_source else "RPMS/x86_64"
-    artifacts_dir = f"{home}/rpmbuild/{subdir}"
-for f in os.listdir(artifacts_dir):
-    for ext in exts:
-        if f.endswith(ext):
-            shutil.copy(f"{artifacts_dir}/{f}", os.path.join(args.output_dir, f))
+        if not is_source:
+            exts = [".rpm"]
+        else:
+            exts = [".src.rpm"]
```
pasqu4le, line 165:
```suggestion
        exts = [".src.rpm"] if is_source else [".rpm"]
```

docker/package/package_generator.py
```diff
@@ -66,113 +98,80 @@ def get_build_deps(args):
...
+        artifacts_dir = f"{home}/rpmbuild/{subdir}"
+
+    for f in os.listdir(artifacts_dir):
+        for ext in exts:
+            if f.endswith(ext):
+                shutil.copy(f"{artifacts_dir}/{f}", os.path.join(output_dir, f))
+
+
+if __name__ == "__main__":
+    main()
```
pasqu4le, line 177: nitpick, but I believe that it's now unnecessary to have `main()` defined above, that function body could simply be moved here instead

pasqu4le (commented):
Mostly minor comments, but overall LGTm and works as intended

docker/package/package_generator.py
krendelhoff2: Okay, but it's intended that no validation made in `package_generator.py`

krendelhoff2: that's why no `else` clause

docker/package/package_generator.py
krendelhoff2: We import argument parser defined here in `docker-tezos-packages.py`

docker/package/packages.py
krendelhoff2: I just wanted to change as much existing code as possible

docker/docker-tezos-packages.py
```diff
@@ -79,29 +85,23 @@ def get_packages_to_build(arg):
...
-            if dist not in fedora_versions:
-                raise Exception(
-                    f"Distribution {dist} is not supported for {target_os}."
-                )
         distributions = args.distributions
+        validate_dists(distributions, fedora_versions, target_os)
     else:
         distributions = ["native"]
+else:
+    raise Exception("Unsupported target os.")
```
pasqu4le, line 99:
```suggestion
    raise Exception("Unsupported target os: " + target_os)
```

pasqu4le (approved):

docker/package/package_generator.py
pasqu4le: You can still do that, my point is that we don't import `main()` anywhere, so it doesn't need to be defined separately.

But I don't think this is a major point, so I'll let you decide.

rvem (approved):

docker/package/packages.py
pasqu4le: @krendelhoff2 I don't see how that would be a bigger difference tho :thinking: 

I mean, you have changed the code to make `packages` a list, then define this list:
```python
packages = [
    {
        "tezos-somethingA": TezosBinaryPackage(<definiton_here>)
    },
    {
        "tezos-somethingB": TezosBinaryPackage(<definiton_here>)
    },
    {
        "tezos-somethingC": TezosBinaryPackage(<definiton_here>)
    },
    {
        "tezos-somethingD": TezosBinaryPackage(<definiton_here>)
    },
]
```
or add an element to it:
```python
packages.append(
    {
        "tezos-somethingE": TezosBinaryPackage(<definiton_here>)
    }
)
```

What I'm suggesting is to have `packages` be a dictionary to begin with, and the two examples above become:
```python
packages = {
  "tezos-somethingA": TezosBinaryPackage(<definiton_here>),
  "tezos-somethingB": TezosBinaryPackage(<definiton_here>),
  "tezos-somethingC": TezosBinaryPackage(<definiton_here>),
  "tezos-somethingD": TezosBinaryPackage(<definiton_here>),
}
```
and
```python
packages["tezos-somethingE"] = TezosBinaryPackage(<definiton_here>)
```
respectively, which seems even more compact to me (and doesn't need `ChainMap` at all).

Don't you think so? If you do, since this has been merge already, please open another PR

PR #556, opened 2022-10-29
[#549] Move convenience script installation to main install phase

### LINKED ISSUE #549: Make sure that convenience scripts/aliases are known to packages

In our packages we define some convenience scripts (e.g. `octez-node-kathmandunet`) and aliases (e.g. `tezos-code-kathmandunet`)  that are added to the system during post-installation.

While this works we don't track these into their package definition, so the package managers are unaware of their existence and don't handle them properly.

This was noticed in Fedora packages (and maybe it only needs to be fixed there) where these aren't listed in the `%files` clause, which made a recent update more difficult than necessary.

# Acceptance criteria

Script and aliases added during post-installation are listed as artifacts in their package definition.

krendelhoff2:
## Description

Problem: Convenience scripts for `tezos-node` were installed as part of postinst phase, thus they were not tracked on fedora.

Solution: Move installation to spec file for fedora.

## Related issue(s)

Resolves #549 

  can be successfully built.


krendelhoff2:
Since additional files install was moved to main installation phase, fedora would refuse to evaluate if you don't list all of them.
On ubuntu, you can check that every file is known running `dpkg-deb -c <binary>.deb`.

docker/package/packages.py
```diff
@@ -272,11 +279,13 @@ def mk_node_unit(
             systemd_units=node_units,
             postinst_steps=node_postinst_steps,
             postrm_steps=node_postrm_steps,
+            supply_additional_files=node_supply_additional_files,
```
pasqu4le, line 282: I'm not sure that I understand the addition of this `supply_additional_files` field.

AFAICT its only definition is `node_supply_additional_files` and it is only set for nodes, so it doesn't seem great to add it (empty/`None`) to all packages only for these ones.

TBC: since this avoids duplication IMO it should remain a function that can be used for both ubuntu and fedora, just not one to be included in the `AbstractPackage` class

krendelhoff2: For now it's relevant only tor `tezos-node`, yes, but it could be useful in the future for another package.
There is actually problem, that `supply_additional_files` function make sense when and only when `additional_files` field is not empty. 
Instead of that two fields I propose to have one field with list of pairs: additional file name and function for delivering it.
Thus, if that list is empty, it just means that package does not have any additional files to install except for main binary and services-related scripts.

pasqu4le: > it could be useful in the future for another package.

:thinking: fair, but the question is: would `supply_additional_files` be any different from `node_supply_additional_files` for them?

I may be wrong, but I think that a single function might do the job. IOW what I'm suggesting is to keep only `node_supply_additional_files`  (renamed) without adding the `supply_additional_files` field.

In that case `additional_files` would be the only field needed (and of course nothing should happen if this is empty).

krendelhoff2: At first I did it that way, but the only place it could happen is `ubuntu.py` or `fedora.py`, where generic package builds, it doesn't know it's name beforehand
That's why i put it in the package definition, it would be quite bad instructions for `tezos-node` specifically in the generic package build script

pasqu4le: > it doesn't know it's name beforehand

sorry, I didn't get that, what do you mean by that?

> it would be quite bad to put instructions specifically for tezos-node in the generic package build script

that I agree, if the installation process is specific to `tezos-node` it shouldn't be added to every other package.

However I don't understand why it is :thinking: I would prefer this to be as simple as possible.

Doing this in the post-install phase allowed us to keep a common main phase and put this node-specific stuff there. The problem is that these files are unknown to packages, but that could be changed AFAIK.

If we modify this approach (and we need to keep it node-specific) we need to pick one that's simpler and more robust and this one doesn't look like that to me TBH :shrug:

pasqu4le: So, I have a few questions:

First: if we keep the post-install approach and add a list for these files, would that be enough? 

Second: why do you think that the approach currently in this PR is better than keeping things in post-install? 

Lastly, if we change the main phase, isn't all that needs to be added:
1. files copied from a local copy
2. symlinks added from said files

then isn't it enough to add one (or two at most) list of strings for which we do that (for every package)?

I'm asking this last question mostly to understand, but I also don't think that this is necessarily better than keeping the post-install approach.

docker/package/model.py
```diff
@@ -300,6 +312,37 @@ def gen_spec_file(self, build_deps, run_deps, out):
...
+                    self.additional_files,
+                )
+            )
+            if self.additional_files
+            else ""
+        )
+
+        additional_files = (
+            "\n".join(
+                map(
```
rvem, line 331: `map(lambda ..., <list>` can be replaced with list comprehension (`["%{bindir}/" + x for x in self.additional_file + ...]`}. Thus it'll be a bit more readable, I believe

docker/package/scripts/octez-node
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
rvem, line 1: A tiny nitpick: the name of this script is the same as binary, which can be confusing (at least it confused me), AFAICS, it's renamed during the actual installation, but I still propose to rename it to something like `octez-node-network`

docker/package/packages.py
krendelhoff2: > that I agree, if the installation process is specific to tezos-node it shouldn't be added to every other package.

krendelhoff2: > First: if we keep the post-install approach and add a list for these files, would that be enough?

I believe keeping post-install approach is semantically incorrect and that's why it already caused some trouble

krendelhoff2: > Lastly, if we change the main phase, isn't all that needs to be added:

krendelhoff2: The pros of current approach is simplicity, the cons is less flexibility (we can only install additional scripts now and nothing more, but I'm not sure if we will ever need something more complicated)

pasqu4le: > What do you think about it?

I'd prefer to indeed remove the `supply_additional_files` and instead only adding a list of scripts to install as part of the main phase.

I'd also like not to keep the current names, as they are more convenient for users (and also to avoid a breaking change).

Now, as you correctly pointed out, a list with only the script names isn't sufficient, because of the renaming, so I think that the best approach would be a list of objects. Each object should contain:
1. the local name of the script
2. (optionally, only if different) the installed name of the script
3. (optionally, only if desired) the name for a symlink to the installed script

This is simpler than the current approach in this PR IMO (at least to use in the future) and should have enough flexibility as well.

Of course, the main installation phase needs to be modified so that (instead of running the `supply_additional_files` lambda) for each of the `additional_files` objects:
1. the script is installed with the given name (or the same one, if no renaming is necessary)
2. a symlink is created for it (if one is required)

Does that sound doable to you?

krendelhoff2: There is updated implementation now, don't you like it? I'll implement as you offered if so

krendelhoff2: But it's still not that simple. Not only script's name changes, but the contents too: `octez-node-kathmandunet` contains reference to `tezos-node-kathmandunet.service` and etc.

docker/package/ubuntu.py
```diff
@@ -5,6 +5,7 @@
 from typing import List
 
 from .model import AbstractPackage
+from .packages import networks
```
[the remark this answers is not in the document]
pasqu4le: @krendelhoff2 not sure if you've seen this comment

docker/package/packages.py
pasqu4le: > There is updated implementation now

Yes, I've seen that already, but it's still based on the `additional_files_install`, so the point still applies.

> Not only script's name changes, but the contents too

And that should remain the case, no? I mean, we don't want to rename `tezos-node-kathmandunet.service` for example.

I don't get why this is a problem, can you please clarify?

pasqu4le (approved):
Aside from the comment above and the commit message that needs an update, this now LGTM :ok_hand:

PR #558, opened 2022-11-02
[#557] Add `tezos-node` as native dep for `tezos-baking`

### LINKED ISSUE #557: [BUG] tezos-setup-wizard fails when setting up baker on limanet
tezos-setup-wizard fails with the following error:

```
Which Tezos network would you like to use?
Currently supported:
1. mainnet (default)           Main Tezos network
2. ghostnet                    Long running test network, currently using the kathmandu
                               Tezos protocol
3. kathmandunet                Test network using the kathmandu Tezos protocol
4. limanet                     Test network using the lima Tezos protocol
> 4
Do you want to set up baking or to run the standalone node?
1. baking (default)            Set up and start all services for baking: tezos-node and
                               tezos-baker.
2. node                        Only bootstrap and run the Tezos node.
>

Would you like your setup to automatically start on boot?
1. yes (default)               Enable the services, running them both now and on every boot
2. no                          Start the services this time only
>
Trying to bootstrap octez-node
tezos-node-limanet.service configuration not found, defaulting to /var/lib/tezos/node-limanet
The error has been logged to /home/ubuntu/tezos_setup_wizard.log
```
```
$ cat /home/ubuntu/tezos_setup_wizard.log
[Errno 2] No such file or directory: '/var/lib/tezos/node-limanet'
```
- install `tezos-baking` from `ppa:serokell/tezos-rc`
- start tezos-setup-wizard
- select limanet
- hit enter to accept defaults  in the following steps
- observe error

# Environment

- OS: ubuntu
- branch/revision: tezos-baking/jammy,now 2:15.0-rc1-0ubuntu1~jammy arm64

krendelhoff2:
## Description
Problem: `tezos-baking` does not depends on `tezos-node`, thus does not install it, resulting in error during setup.

Solution: Add `tezos-node` as native dep for `tezos-baking`.

## Related issue(s)

Resolves #557

  can be successfully built.

### DIFF
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -514,7 +514,7 @@ class TezosBakingServicesPackage(AbstractPackage):
     # native releases, so we append an extra letter to the version of
     # the package.
     # This should be reset to "" whenever the native version is bumped.
-    letter_version = ""
+    letter_version = "a"
 
     buildfile = "setup.py"
 
--- a/docker/package/model.py
+++ b/docker/package/model.py
@@ -557,10 +557,12 @@ class TezosBakingServicesPackage(AbstractPackage):
         target_networks: List[str],
         network_protos: Dict[str, List[str]],
         meta: PackagesMeta,
+        additional_native_deps: List[str],
     ):
         self.name = "tezos-baking"
         self.desc = "Package that provides systemd services that orchestrate other services from Tezos packages"
         self.meta = deepcopy(meta)
+        self.additional_native_deps = additional_native_deps
         self.meta.version = self.meta.version + self.letter_version
         self.target_protos = set()
         self.patches = []
@@ -610,9 +612,7 @@ class TezosBakingServicesPackage(AbstractPackage):
         shutil.copy(f"{os.path.dirname(__file__)}/tezos_voting_wizard.py", package_dir)
 
     def gen_control_file(self, deps, ubuntu_version, out):
-        run_deps_list = ["acl"]
-        for proto in self.target_protos:
-            run_deps_list.append(f"tezos-baker-{proto.lower()}")
+        run_deps_list = map(lambda x: x.lower(), self.additional_native_deps)
         run_deps = ", ".join(run_deps_list)
         file_contents = f"""
 Source: {self.name}
@@ -633,9 +633,7 @@ Description: {self.desc}
             f.write(file_contents)
 
     def gen_spec_file(self, build_deps, run_deps, out):
-        run_deps = ", ".join(
-            ["acl"] + [f"tezos-baker-{proto}" for proto in self.target_protos],
-        )
+        run_deps = ", ".join(self.additional_native_deps)
         (
             systemd_deps,
             systemd_install,
--- a/docker/package/packages.py
+++ b/docker/package/packages.py
@@ -431,6 +431,8 @@ packages.append(
         target_networks=networks.keys(),
         network_protos=networks_protos,
         meta=packages_meta,
+        additional_native_deps=[f"tezos-baker-{proto}" for proto in active_protocols]
+        + ["tezos-node", "acl"],
     )
 )
 
--- a/meta.json
+++ b/meta.json
@@ -1,5 +1,5 @@
 {
-  "release": "2",
+  "release": "3",
   "maintainer": "Serokell <hi@serokell.io>",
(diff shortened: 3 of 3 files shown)


krendelhoff2:
`tezos-baking` depends on bakers on both os, transitively getting `tezos-node` as well.

docker/package/packages.py
```diff
@@ -390,6 +390,7 @@ def mk_node_unit(
             additional_native_deps=[
                 "tezos-sapling-params",
                 "tezos-client",
+                "tezos-node",
```
pasqu4le, line 393: This seems to me the wrong place to add this, as this is the `tezos-baker-*` package definition.

One should be able to install the baker without also installing a node (it's not a strict dependency for it).

OTOH this dependency should be part of the `tezos-baking` package instead

krendelhoff2: why then it makes sense to add `tezos-client` as dependency?

pasqu4le: because `tezos-client` is necessary to register a baker, before one can run the baker process.

docker/package/packages.py
```diff
@@ -431,6 +431,8 @@ def mk_node_unit(
         target_networks=networks.keys(),
         network_protos=networks_protos,
         meta=packages_meta,
+        additional_native_deps=[f"tezos-baker-{proto}" for proto in active_protocols]
+        + ["tezos-node"],
```
pasqu4le, line 435: Why not add `acl` here too?

It seems that we do so for `tezos-baker-*` processes, so it would be nice to be consistent.

docker/package/packages.py
```diff
@@ -431,6 +431,8 @@ def mk_node_unit(
         target_networks=networks.keys(),
         network_protos=networks_protos,
         meta=packages_meta,
+        additional_native_deps=[f"tezos-baker-{proto}" for proto in active_protocols]
```
pasqu4le, line 434: I think it would be better to lowercase `proto` here instead of `model.py`.

It would be done once instead of twice and we don't need to lowercase `tezos-node` nor `acl` (also maybe in the future we'll need to add more)

docker/package/model.py
```diff
@@ -557,10 +557,12 @@ def __init__(
         target_networks: List[str],
         network_protos: Dict[str, List[str]],
         meta: PackagesMeta,
+        additional_native_deps: List[str],
```
pasqu4le, line 560: Since this is now no longer exclusive to `TezosBinaryPackage`, wouldn't be better to add this to `AbstractPackage` instead with an empty list as a default value?

docker/package/packages.py
krendelhoff2: but we already depend on `tezos-baker-*`, thus there is no need in duplication

docker/package/packages.py
krendelhoff2: we can't lowercase it here, `fedora` packages not lowercased

krendelhoff2: since we lowercase only `ubuntu` packages, lowercasing is always safe even if there is no need to

docker/package/packages.py
pasqu4le: Oh, but what I mean is that we already do, except it's added to the list in (two places in) `model.py`.

Which BTW I think it's a good idea to have, to make it clear that it's a dependency for this package too.

docker/package/model.py
krendelhoff2: How it would help? We use that field in class constructor and still have to specify it in `__init__` consturtor

pasqu4le: that's a good point, I was thinking it could have saves some unnecessary repetition.

Never mind then :ok_hand:

pasqu4le (commented):
One more comment, since we'll make a new release after this, can you also update `letter_version`?

docker/package/packages.py
```diff
@@ -431,6 +431,8 @@ def mk_node_unit(
...
+        + ["tezos-node" "acl"],
```
pasqu4le, line 435: This needs a fix:
```suggestion
        + ["tezos-node", "acl"],
```

pasqu4le (approved):

pasqu4le:
@krendelhoff2 please next time remember to add the issue number in the branch name

PR #565, opened 2022-11-15
[#564] Remove dry-run flag

### LINKED ISSUE #564: Work around `dry-run` bug when registering delegates on lima

In the tezos setup wizard we currently check that an address has sufficient funds to register as a delegate by calling a command like:
```
octez-client register key <mgr> as delegate --dry-run
```
which used to work, but it doesn't on `lima` in the latest release if the `dry-run` option is used, see [tezos issue 4200](https://gitlab.com/tezos/tezos/-/issues/4200) for more info on the matter.

We need to work around this problem for the time being so that the wizard keeps working on the `kathmandu` as well as on the `lima` protocol.

# Acceptance criteria

1. The failing command mentioned above is no longer used on `lima`.
2. a better solution to obtain the same result on `kathmandu` and `lima` is implemented

krendelhoff2:
## Description

Problem: It's impossible to dry-run delegate registration
on lima due to bug.

Solution: Remove `--dry-run` for now, until
https://gitlab.com/tezos/tezos/-/issues/4200 resolved.

## Related issue(s)

Resolves #564

  can be successfully built.

### DIFF
--- a/docker/package/wizard_structure.py
+++ b/docker/package/wizard_structure.py
@@ -715,13 +715,13 @@ class Setup:
                         f"Waiting for funds to arrive... (Ctrl + C to choose another option)."
                     )
                     try:
-                        # dry-run delegate registration until it succeeds
                         while True:
                             result = get_proc_output(
                                 f"sudo -u tezos {suppress_warning_text} octez-client {tezos_client_options} "
-                                f"register key {baker_alias} as delegate --dry-run"
+                                f"register key {baker_alias} as delegate"
                             )
                             if result.returncode == 0:
+                                print(result.stdout.decode("utf8"))
                                 break
                             else:
                                 proc_call("sleep 1")


docker/package/wizard_structure.py
```diff
@@ -715,13 +715,16 @@ def import_key(self, key_mode_query, ledger_app=None):
...
                             result = get_proc_output(
                                 f"sudo -u tezos {suppress_warning_text} octez-client {tezos_client_options} "
-                                f"register key {baker_alias} as delegate --dry-run"
+                                f"get balance for {baker_alias}"
                             )
-                            if result.returncode == 0:
+                            # workaround until https://gitlab.com/tezos/tezos/-/issues/4200 resolved
+                            if result.returncode == 0 and (
+                                int(re.search(b"[0-9]+", result.stdout).group(0))
+                                >= 6000
```
pasqu4le, line 726: There's a few problems with this, because AFAIK 6000 is needed for baking (and baking only), but there's more to keep in mind.

If the delegate has 6000 tez but hasn't been registered yet then the next operation in the setup wizard (the delegate registration itself) will put the balance below the minimum (due to fees) needed for baking, which is not great.

OTOH if the delegate has been registered already, 6000 is perfectly fine for baking, so this logic is "correct".

However it really isn't the best way to do this either way, because one can _have_ less than 6000 tez, but still be able to bake. For example one might have a very low balance (IIRC 600 might be the minimum) but others have delegated to this address. This still works and isn't an impediment to baking... :confused: 
So IMO the logic based on balance is not the best for this workaround.

Another point is that AFAIK (but please check, I didn't make sure) voting doesn't require 6000 tez and this function here is also called in the voting wizard. 

As I mentioned elsewhere maybe we should drop the `--dry-run` and stop as soon as the registration command returns successfully (either because it registered the baker now or because the baker was registered already).

pasqu4le (changes requested):

docker/package/tezos_setup_wizard.py
```diff
@@ -424,12 +424,14 @@ def import_baker_key(self):
...
-            f"register key {baker_alias} as delegate"
-        )
+        # TODO uncomment after https://gitlab.com/tezos/tezos/-/issues/4200 resolved
+        # and validation fixed
+        # tezos_client_options = self.get_tezos_client_options()
+        # baker_alias = self.config["baker_alias"]
+        # proc_call(
+        #    f"sudo -u tezos {suppress_warning_text} octez-client {tezos_client_options} "
+        #    f"register key {baker_alias} as delegate"
+        # )
```
pasqu4le, line 434: Actually, now that we are using this logic (of running it without `--dry-run`) perhaps it makes sense to make this solution permanent.

AFAIK one needs an address that is a delegate both for baking and voting (note that both wizards call `import_key` from the wizard structure).

So IMO the right solution is to call the "register delegate" command (without `--dry-run`) and until it has an exit code of `0` (either because a delegate got registered or because it was already) in `import_key`.

@krendelhoff2 how does that sound to you?
If you implement this solution please make sure to test both wizards, because I haven't checked that it works alright.

pasqu4le (changes requested):

docker/package/tezos_setup_wizard.py
krendelhoff2: Actually yes, it sounds perfectly fine. Reregistering delegate is safe and free, so it does not cause any problem anywhere.

krendelhoff2: I'll reword commits before merge

docker/package/wizard_structure.py
```diff
@@ -715,13 +715,13 @@ def import_key(self, key_mode_query, ledger_app=None):
                         f"Waiting for funds to arrive... (Ctrl + C to choose another option)."
                     )
                     try:
-                        # dry-run delegate registration until it succeeds
                         while True:
                             result = get_proc_output(
                                 f"sudo -u tezos {suppress_warning_text} octez-client {tezos_client_options} "
-                                f"register key {baker_alias} as delegate --dry-run"
+                                f"register key {baker_alias} as delegate"
```
pasqu4le, line 721: Now there is an unnecessary register baker call in the setup wizard tho, in the `register_baker` function, which is called right after this one.

On a related point, maybe this function's name should be changed now BTW

docker/package/wizard_structure.py
```diff
@@ -715,13 +715,13 @@ def import_key(self, key_mode_query, ledger_app=None):
...
                             )
                             if result.returncode == 0:
+                                print(result.stdout.decode("utf8"))
```
pasqu4le, line 724: Why add this? :thinking:

docker/package/wizard_structure.py
krendelhoff2: Since it unharmful, I don't think we should change it

docker/package/wizard_structure.py
krendelhoff2: Because otherwise user chose `generate-fresh-key` option won't get `octez-client` output with baker registering information

docker/package/wizard_structure.py
pasqu4le: > it became unnecessary only for `generate-fresh-key` option

I'm confused, that's exactly what I'm referring to.

I mean, there is a "register key" call to `tezos-client` in the `register_baker` function (the only one not in `import_key`) which I think should be removed now.

AFAICT `register_baker` is only called in `run_setup` _after_ `import_baker_key`, which already registers the delegate.

In fact, the whole `register_baker` function now seems no longer needed and IMO `import_key` should be renamed to `import_delegate_key` to make clear(er) that delegate registration happens there too.

What do you think, does it make sense to you now?

krendelhoff2: `import-key` registers delegate only for `generate-fresh-key` option. If user choose another option, we still need `register_baker` function. But your idea to merge that two function make sense, so that for every option delegate will be registered after import

pasqu4le (approved):

PR #566, opened 2022-11-16
[#526] Provide packages for `epel-x86_64`

### LINKED ISSUE #526: Publish packages with static binaries for 'epel-7-x86_64' automatically

We were asked to provide packages for RHEL 7. However, its repository
doesn't have packages with `opam` and some other dependencies that are needed for building
Octez binaries. As a temporary solution, we published packages that include static binaries manually for this chroot.
However, this needs to be automated for future releases.

# Acceptance criteria

Source packages that contain static binaries are published for `epel-7-x86_64` along with the rest Fedora packages.
Note that source packages for Fedora 34 and 35 should still contain proper Octez sources with dependencies.

krendelhoff2:
## Description

Problem: We were asked to provide packages for RHEL 7.

Solution: Provide CI steps for publishing epel packages during release.

## Related issue(s)

Resolves #526 

  can be successfully built.


.buildkite/pipeline-for-tags.yml
```diff
@@ -33,6 +33,49 @@ steps:
    - buildkite-agent artifact download "out/*" . --step build-source-packages
    - nix develop .#autorelease -c ./scripts/publish-native-packages.sh out
```
pasqu4le, line 34: AFAIU this will still cause copr to try and build source packages for EPEL7 (and fail).

Can't this be avoided?

scripts/publish-native-packages.sh
```diff
@@ -53,5 +53,5 @@ done
 
 for f in "$source_packages_path"/*.src.rpm; do
   rpmsign --define="%_gpg_name Serokell <tezos-packaging@serokell.io>" --define="%__gpg $(which gpg)" --addsign "$f"
-  /run/wrappers/bin/sudo -u copr-uploader /run/current-system/sw/bin/copr-cli build "$copr_project" --nowait "$f"
+  /run/wrappers/bin/sudo -u copr-uploader /run/current-system/sw/bin/copr-cli build "$copr_project" "${CHROOTS:-}" --nowait "$f"
```
rvem, line 56: Last time I tried `--chroots` it didn't work for me on the version from our nixpkgs fork:(

.buildkite/pipeline-for-tags.yml
rvem: It should be possible to do `--chroots fedora-{35,36}-{x86_64,aarch64}`, but this will require us to actually hardcode them:( 
Maybe there is a way to get all chroots for a given copr repo via API :thinking:

pasqu4le (approved):

PR #574, opened 2022-11-24
[#572] Use block hash from snapshot metadata in setup wizard

### LINKED ISSUE #572: Run snapshot import with --block when using snapshot metadata

Snapshots found via `tezos-snapshots.json` metadata, such as those from xtz-shots.io can be imported with `--block <block_hash>` since metadata includes snapshot's block hash for additional data verification. Currently user sees the following message:

```
Downloading the snapshot from https://mainnet-v15.xtz-shots.io/mainnet-2909360.rolling
Progress: 100 %, 6123 MB
Nov 23 15:57:28.032 - node.snapshots: importing snapshot in on-disk mode. Mind using the --in-memory flag to allow
Nov 23 15:57:28.032 - node.snapshots:   a faster import which uses a higher quantity of ram
Nov 23 15:57:28.033 - node.snapshots: importing data from snapshot
Nov 23 15:57:28.033 - node.snapshots:   /tmp/octez_node.snapshot: chain TEZOS_MAINNET, block hash BKvWfbLaHeqjXprix6ykHZdL5MvyXWpp7c5AWxMC4TXWHRKv1Q3 at level 2909360, timestamp 2022-11-23T16:16:59-00:00 in rolling (snapshot version 4)
Nov 23 15:57:28.033 - node.snapshots: you may consider using the --block <block_hash> argument to ensure that the
Nov 23 15:57:28.033 - node.snapshots:   block imported is the one you expected
```

# Acceptance criteria

when using xtz-shots.io snapshots snapshot import is invoked with --block option and import log reflects this accordingly

pasqu4le:
## Description

Problem: octez-node has a 'block' option when importing a snapshot that can be used to ensure that the block that is imported is the one matching the given hash.
We can collect the block hash from the metadata (when one is available), but the setup wizard doesn't use this info.

Solution: get the block hash when fetching snapshots from metadata and use it during import in the setup wizard.

## Related issue(s)

Resolves #572

  can be successfully built.

### DIFF
--- a/docker/package/tezos_setup_wizard.py
+++ b/docker/package/tezos_setup_wizard.py
@@ -236,6 +236,8 @@ class Setup(Setup):
     # that is suited for the chosen history mode and network
     def get_snapshot_link(self):
         self.config["snapshot_url"] = None
+        self.config["snapshot_block_hash"] = None
+
         json_url = "https://xtz-shots.io/tezos-snapshots.json"
         try:
             with urllib.request.urlopen(json_url) as url:
@@ -245,7 +247,7 @@ class Setup(Setup):
             print(f"Couldn't collect snapshot metadata from {json_url}")
             return
 
-        self.config["snapshot_url"] = next(
+        snapshot_metadata = next(
             filter(
                 lambda artifact: artifact["artifact_type"] == "tezos-snapshot"
                 and artifact["chain_name"] == self.config["network"]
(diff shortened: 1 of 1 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -259,7 +261,10 @@ def get_snapshot_link(self):
                 iter(snapshot_array),
             ),
             {"url": None},
```
krendelhoff2, line 263:
```suggestion
            {"url": None, "block_hash": None},
```

krendelhoff2 (approved):
LGTM, small nitpick

PR #584, opened 2022-12-28
[Chore] Fix environment files regex

krendelhoff2:
## Description

Problem: For some reason, systemd services
packaging on bionic results in `EnvironmentFile`
clause in unit file, while other distros has `EnvironmentFiles`.

Solution: Make 's' optional.

## Related issue(s)

Resolves #

  can be successfully built.


docker/package/wizard_structure.py
```diff
@@ -196,7 +196,7 @@ def show_systemd_service(service_name):
 
 
 def find_systemd_env_files(show_systemd_output):
-    return re.findall(b"EnvironmentFiles=(.*) ", show_systemd_output)
+    return re.findall(b"EnvironmentFiles?=(.*) ", show_systemd_output)
```
pasqu4le, line 199: This seems to fix the problem here, however when I tested the `tezos-setup` I ended up with a failure either way, the output ends with:
```
Waiting for the node to bootstrap...
Fatal error:
  Unix.Unix_error(Unix.EACCES, "mkdir", "/home/fresh/.tezos-client")
The error has been logged to /home/fresh/tezos_setup.log
```
and the error logged is:
```
Command '['sudo', '-u', 'tezos', 'TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER=YES', 'octez-client', '--endpoint', 'http://127.0.0.1:8732', 'bootstrapped']' returned non-zero exit status 1.
```

pasqu4le (changes requested):

docker/package/wizard_structure.py
pasqu4le: The issue might be in `octez-node-wrapper` :thinking: 

That also seems to be looking for the `EnvironmentFiles` (with the `s`) and probably needs to be updated as well

krendelhoff2: You are right, but it's not the source of issue: it's failure of `octez-client`. But it's rather confusing, I had same issue and than it resolved by its own. Despite being called with `sudo -u tezos` it tries to create client dir at your user home dir, which is misbehavior. I need to investigate further.

krendelhoff2:
```
vagrant@ubuntu-bionic:~$ systemctl show -p EnvironmentFiles tezos-node-limanet.service 
EnvironmentFile=/etc/default/tezos-node-limanet (ignore_errors=no)
```

krendelhoff2: 🤔

pasqu4le: Indeed the `show` command seems to work correctly with the `s` on `jammy` too :ok_hand: 

Closing this thread as there is now a separate commit handling the other issue I mentioned above too

docker/docker-tezos-packages.py
```diff
@@ -103,19 +103,23 @@ def validate_dists(dists, versions, target_os):
...
-check_call(
-    f"""
-{virtualisation_engine}
-build -t tezos-{target_os}
--f docker/package/Dockerfile-{target_os} .
-"""
-)
+# that option reflects launchpad and copr build infrastruture
+# and has a purpose to behave more similar to them
+images = ["36"] if target_os == "fedora" else distributions
```
pasqu4le, line 108: Can you clarify this? I don't get why `36` would be the default value for fedora `images` but `distributions` is picked otherwise

docker/package/ubuntu.py
```diff
@@ -16,23 +16,26 @@ def build_ubuntu_package(
...
-    if source_archive_path is None:
-        pkg.fetch_sources(dir, binaries_dir)
-        pkg.gen_buildfile("/".join([dir, pkg.buildfile]), binaries_dir)
-        subprocess.run(["tar", "-czf", f"{dir}.tar.gz", dir], check=True)
-    else:
-        shutil.copy(f"{cwd}/../{source_archive_path}", f"{dir}.tar.gz")
-        subprocess.run(["tar", "-xzf", f"{dir}.tar.gz"], check=True)
-    pkg.meta.version = fixed_version
     for ubuntu_version in ubuntu_versions:
+        # ubuntu prohibits uppercase in packages names
```
pasqu4le, line 20: Why this change? weren't the definitions above working all the same and in common to different ubuntu versions? :thinking:

pasqu4le (commented):
Both the issue with the client dir and `EnvironmentFile`s seem to be fixed now :+1:

docker/package/wizard_structure.py
```diff
@@ -349,12 +349,12 @@ def get_key_address(tezos_client_options, key_alias):
...
 
 
-def wait_for_ledger_app(ledger_app):
+def wait_for_ledger_app(ledger_app, client_dir):
     output = b""
     try:
         while re.search(f"Found a Tezos {ledger_app}".encode(), output) is None:
             output = get_proc_output(
-                f"sudo -u tezos {suppress_warning_text} octez-client list connected ledgers"
+                f"sudo -u tezos {suppress_warning_text} octez-client --base-dir {client_dir} list connected ledgers"
```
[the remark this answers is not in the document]
krendelhoff2: I suppose that changes are not necessary now at all after python upgrade.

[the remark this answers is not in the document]
pasqu4le: As I was explaining, it seems that there is a problem with the old `sudo` version used, so these changes still seem necessary to me.

Bumping the `python` version used launches the command correctly, but that doesn't fix the command :slightly_smiling_face:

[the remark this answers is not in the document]
krendelhoff2: Sorry, I read your comment inattentively

docker/package/ubuntu.py
krendelhoff2: It's because ubuntu quirk. `python3.8` is available on `bionic` and `focal`, but not on `jammy`. So almost everythere we need dispatching according to version. It could be solved if we use third party launchpads so we could supply `python3.8` as dep for every ubuntu version.

docker/docker-tezos-packages.py
krendelhoff2: As you remember, `copr` build infrastructure uses `mock` for build on `fedora 36`, so it would be better option for our builds.
For ubuntu builds, we build them in impure environments, for now I found it quite complicated to run `pbuilder` under docker. Thus, for better reproducibility, I think we should build `bionic` packages in `bionic` container and etc, because during implementation of this PR, I got some strange errors on `bionic` package installation when build on `focal` container.

pasqu4le: > copr build infrastructure uses mock for build on fedora 36, so it would be better option for our builds.

Ah, I see, but I'm not sure that we should try to match `copr`, instead I propose that we use the latest available stable fedora. What do you think?

AFAIK one of the features of `mock` is being able to perform a build with some isolation, so it should work on different fedora versions in the same manner.

Also, we aren't sure of what `copr` is using for its builds now or in the future, so I don't this we can reasonably follow what they do.

Maybe picking the lastest available `fedora` version (so `37` now) would be the best option for us

pasqu4le: Also, it would be nice to have a comment here that's a bit longer and more clear explaining this choice, otherwise we'll likely forget why we did this in the future.

I'd also suggest to mention that as time goes on this fedora version should be updated.

docker/package/ubuntu.py
pasqu4le: I see, got it :ok_hand: 

> It could be solved if we use third party launchpads

Nope, I don't think that's worth the trouble

pasqu4le (commented):
Aside from [the open thread](https://github.com/serokell/tezos-packaging/pull/584/files#r1060796759) changes LGTM

Please improve the comment and consider using fedora 37, but feel free to squash fixup commits already

docker/docker-tezos-packages.py
krendelhoff2: > AFAIK one of the features of mock is being able to perform a build with some isolation, so it should work on different fedora versions in the same manner.

We just had the problem with `mock bootstrap` option which showed it's not entirely true even using `mock`. 
On the other hand, docs says that copr uses latest stable fedora, so both ways are about the same.

pasqu4le (approved):
LGTM and the fix seems to be working

PR #592, opened 2023-01-19
[Chore] Optimize docker builds

krendelhoff2:
## Description

Problem: Dependencies and sources were not reused
and rebuilt every time, despite they are about the
same for all packages.

Solution: Refactor Dockerfiles, split stages in different scripts and
place them in separate layers to reuse results using docker cache system.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/build/Dockerfile.aarch64
+++ b/docker/build/Dockerfile.aarch64
@@ -6,16 +6,21 @@ FROM alpine:3.12 as binary-fetch
 # See https://bugs.launchpad.net/qemu/+bug/1892684.
 RUN wget https://github.com/multiarch/qemu-user-static/releases/download/v4.2.0-7/qemu-aarch64-static
 RUN chmod +x qemu-aarch64-static
-FROM arm64v8/alpine:3.12
+
+FROM arm64v8/alpine:3.12 as builder
 COPY --from=binary-fetch qemu-aarch64-static /usr/bin/qemu-aarch64-static
 RUN apk update
 RUN apk --no-cache --virtual add rsync git m4 build-base patch unzip \
-  bubblewrap wget pkgconfig gmp-dev libev-dev hidapi-dev eudev-dev perl opam libusb-dev bash \
-  autoconf automake libtool linux-headers cargo libffi-dev zlib-static coreutils perl-utils
+  bubblewrap wget pkgconfig gmp-dev libev-dev \
+  hidapi-dev eudev-dev perl opam libusb-dev bash \
+  autoconf automake libtool linux-headers cargo \
+  libffi-dev zlib-static coreutils perl-utils cmake
 COPY ./build/build-libusb-and-hidapi.sh /build-libusb-and-hidapi.sh
 RUN /build-libusb-and-hidapi.sh
 COPY ./build/build-rust.sh /build-rust.sh
 RUN /build-rust.sh
+COPY ./build/build-upx.sh /build-upx.sh
+RUN /build-upx.sh
 ARG OCTEZ_VERSION
 RUN git clone --single-branch --depth 1 --branch "$OCTEZ_VERSION" https://gitlab.com/tezos/tezos.git
 WORKDIR /tezos
@@ -24,3 +29,4 @@ COPY ./build/build-deps.sh /build-deps.sh
 RUN /build-deps.sh
 COPY ./build/build-tezos.sh /build-tezos.sh
 RUN /build-tezos.sh
+RUN upx octez-*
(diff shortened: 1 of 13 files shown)


pasqu4le (changes requested):
I seem to have an issue with the emulated build for arm64, using `docker-static-build.sh aarch64` results in this error:
```
exec /bin/sh: exec format error
The command '/bin/sh -c git clone --single-branch --depth 1 --branch "$OCTEZ_VERSION" https://gitlab.com/tezos/tezos.git' returned a non-zero code: 1
```

krendelhoff2:
@pasqu4le you have to run [command](https://github.com/serokell/tezos-packaging/blob/master/docker/README.md#compiling-for-aarch64-on-x86_64-prerequisites) to make it work?

docker/build/build-deps.sh
```diff
@@ -0,0 +1,12 @@
+#! /usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+set -euo pipefail
+
+source "$HOME/.cargo/env"
+# Disable usage of instructions from the ADX extension to avoid incompatibility
+# with old CPUs, see https://gitlab.com/dannywillems/ocaml-bls12-381/-/merge_requests/135/
```
pasqu4le, line 10: This comment is misplaced, it's setting the `BLST_PORTABLE` what's written here possible

docker/build/Dockerfile.aarch64
```diff
@@ -17,5 +17,11 @@ RUN /build-libusb-and-hidapi.sh
...
 ARG OCTEZ_VERSION
+RUN git clone --single-branch --depth 1 --branch "$OCTEZ_VERSION" https://gitlab.com/tezos/tezos.git
+WORKDIR /tezos
+ENV BLST_PORTABLE yes
+ENV OPAMYES true
+COPY ./build/build-deps.sh /build-deps.sh
+RUN /build-deps.sh
 COPY ./build/build-tezos.sh /build-tezos.sh
-RUN /build-tezos.sh ${OCTEZ_VERSION}
+RUN /build-tezos.sh
```
pasqu4le, line 32: :thinking: I'm not a docker expert, but does this change mean that in case there is a failure in `build-tezos.sh` what was done by `build-deps.sh` won't be repeated in a new run?

pasqu4le (commented):
> using docker-static-build.sh aarch64 results in this error

indeed it seems that I was wrong about that, I though that I had followed the instructions in the doc, but apparently I made some mistake, never mind [my previous comment](https://github.com/serokell/tezos-packaging/pull/592#pullrequestreview-1272870585) then.

docker/build/build-upx.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+# This script builds upx. alpine:3.12 for aarch64 doesn't provide upx,
+# so we should build it natively for consistensy.
+set -euo pipefail
+
+git clone --depth 1 --branch v4.0.1 https://github.com/upx/upx.git
```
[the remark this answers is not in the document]
krendelhoff2: Well, if we are fine with downloading binary blobs why we build tezos binaries at all?

docker/build/Dockerfile.aarch64
krendelhoff2: `RUN` command creates intermediary image (layer) which gets cached

krendelhoff2: exactly that change is unrelated, since `$OCTEZ_VERSION` is build time arg it won't prevent caching in any way

krendelhoff2: but octez version is irrelevant for script now

docker/build/build-upx.sh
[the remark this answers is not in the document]
krendelhoff2: and for different architectures it would require some dispatching. That way we are sure that it will work both on `aarch64` and `x86_64`.

docker/docker-tezos-packages.py
```diff
@@ -123,7 +123,7 @@ def validate_dists(dists, versions, target_os):
 # so we should also keep that way
 # for ubuntu builds, since we lack `pbuilder` for now,
 # packages should be built in respective containers for better reproducibility
-images = ["37"] if target_os == "fedora" else distributions
+images = ["latest"] if target_os == "fedora" else distributions
```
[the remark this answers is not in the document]
krendelhoff2: dropped commit

pasqu4le (approved):
One minor comment, but otherwise LGTM now

PR #597, opened 2023-01-25
[Chore] Add option to fetch sources from launchpad

krendelhoff2:
## Description

Problem: When we make new release based on the same upstream octez version, we have to use previous source archive to upload packages to launchpad. Since they are available on predictable urls, we should have option to fetch them automatically during build.

Solution: Add flag to get sources from launchpad if relevant source archive is absent under `sources-dir` (or it's not defined).

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/package/package_generator.py
+++ b/docker/package/package_generator.py
@@ -5,6 +5,7 @@ import os
 import sys
 import shutil
 import argparse
+import urllib.request
 from .fedora import build_fedora_package
 from .ubuntu import build_ubuntu_package
 from .packages import packages as all_packages
@@ -39,6 +40,12 @@ parser.add_argument(
     help="specify directory with source archive(s) for ubuntu packages",
     type=os.path.abspath,
 )
+parser.add_argument(
+    "--launchpad-sources",
+    help="download sources from launchpad instead of providing them",
+    action="store_true",
+)
+parser.set_defaults(launchpad_sources=False)
 
 
 def get_ubuntu_run_deps(binaries_dir):
@@ -110,6 +117,8 @@ def main():
 
     source_archives = os.listdir(args.sources_dir) if args.sources_dir else []
 
+    dl_sources = args.launchpad_sources
+
     binaries_dir = args.binaries_dir
 
     if target_os == "ubuntu":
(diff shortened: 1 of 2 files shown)


docker/package/package_generator.py
```diff
@@ -132,7 +141,7 @@ def main():
     for package_name in args.packages:
         packages.append(all_packages[package_name])
 
-    if is_source and source_archives:
+    if is_source and (source_archives or dl_sources):
```
pasqu4le, line 144: I think that these should be mutually exclusive.

AFAIU right now if I provide both a source archive dir and use `--launchpad-sources` then a source archive in the dir (if any) will be preferred over one to download.

This could easily be confusing in some situations, I'd prefer an easier and more predictable behavior

docker/package/package_generator.py
```diff
@@ -39,6 +40,12 @@
     help="specify directory with source archive(s) for ubuntu packages",
     type=os.path.abspath,
 )
+parser.add_argument(
+    "--launchpad-sources",
+    help="download sources from launchpad if not provided",
```
pasqu4le, line 45: If these are mutually exclusive this description is not correct:
```suggestion
    help="download sources from launchpad instead of providing them",
```

docker/package/package_generator.py
```diff
@@ -143,6 +152,22 @@ def main():
...
+                    version = package.meta.version
+                    repo = "tezos-rc" if "rc" in version else "tezos"
+                    release = package.meta.release
+                    name = package.name
+                    url = f"https://launchpad.net/~serokell/+archive/ubuntu/{repo}/+sourcefiles/{name}/2:{version}-0ubuntu{release}~jammy/{name}_{version}.orig.tar.gz"
+                    try:
+                        urllib.request.urlretrieve(
+                            url,
+                            source_archive,
+                        )
```
pasqu4le, line 165: Maybe it would be nice to print a message here, saying that the source archive has been dowloaded.

docker/package/package_generator.py
pasqu4le, line 144: Shouldn't this be limited to Ubuntu?

docker/package/package_generator.py
```diff
@@ -143,6 +152,22 @@ def main():
                     package.source_archive = os.path.join(
                         args.sources_dir, source_archive
                     )
+                elif dl_sources:
+                    version = package.meta.version
+                    repo = "tezos-rc" if "rc" in version else "tezos"
+                    release = package.meta.release
+                    name = package.name
+                    url = f"https://launchpad.net/~serokell/+archive/ubuntu/{repo}/+sourcefiles/{name}/2:{version}-0ubuntu{release}~jammy/{name}_{version}.orig.tar.gz"
```
pasqu4le, line 160: How come `jammy` is hardcoded here? Wouldn't it be better to also fill that from the arguments when available?

pasqu4le (approved):

PR #609, opened 2023-02-23
[Chore] Change logic for tezos-baking sources

krendelhoff2:
## Description

Problem: Currently there is no way to provide sources for `tezos-baking` package. But what do we really need is ignore absence of it.

Solution: Let sources for tezos-baking be provided and ignore if they are not.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/package/package_generator.py
+++ b/docker/package/package_generator.py
@@ -135,18 +135,17 @@ def main():
     if is_source and source_archives:
         errors = []
         for package in packages:
-            if getattr(package, "letter_version", None) is None:
-                source_archive = (
-                    f"{package.name.lower()}_{package.meta.version}.orig.tar.gz"
+            source_archive = (
+                f"{package.name.lower()}_{package.meta.version}.orig.tar.gz"
+            )
+            if source_archive in source_archives:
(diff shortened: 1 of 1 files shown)


docker/package/package_generator.py
```diff
@@ -135,18 +135,15 @@ def main():
...
-            if getattr(package, "letter_version", None) is None:
-                source_archive = (
-                    f"{package.name.lower()}_{package.meta.version}.orig.tar.gz"
+            source_archive = (
+                f"{package.name.lower()}_{package.meta.version}.orig.tar.gz"
+            )
+            if source_archive in source_archives:
+                package.source_archive = os.path.join(args.sources_dir, source_archive)
+            elif getattr(package, "letter_version", None) is None:
+                errors.append(
```
pasqu4le, line 144:
```suggestion
            elif getattr(package, "letter_version", None) is None:
                # We throw an error if the source dir is missing, unless the package is tezos-baking
                # for which we don't need new sources.
                errors.append(
```

pasqu4le (approved):

PR #614, opened 2023-03-06
[Chore] Change way to retrieve opam repository tag

krendelhoff2:
## Description

Problem: Since last release our way to parse tag from `version.sh` is not relevant anymore.

Solution: Evaluate and output variable contents instead of parsing it.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/fetch_tezos_sources.py
+++ b/docker/fetch_tezos_sources.py
@@ -28,12 +28,18 @@ shutil.rmtree(os.path.join("tezos", ".git"))
 
 subprocess.run(["git", "clone", "https://gitlab.com/tezos/opam-repository.git"])
 
-with open("tezos/scripts/version.sh", "r") as f:
-    opam_repository_tag = re.search(
-        "^export opam_repository_tag=([0-9a-z]*)", f.read(), flags=re.MULTILINE
-    ).group(1)
-    os.chdir("opam-repository")
-    subprocess.run(["git", "checkout", opam_repository_tag])
-    subprocess.run(["rm", "-rf", ".git"])
-    subprocess.run(["rm", "-r", "zcash-params"])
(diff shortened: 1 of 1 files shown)


docker/fetch_tezos_sources.py
```diff
@@ -28,12 +28,17 @@
...
-    os.chdir("opam-repository")
-    subprocess.run(["git", "checkout", opam_repository_tag])
-    subprocess.run(["rm", "-rf", ".git"])
-    subprocess.run(["rm", "-r", "zcash-params"])
-    subprocess.run(["opam", "admin", "cache"])
+with open("get_opam_tag.sh", "w") as f:
+    f.write("source ./tezos/scripts/version.sh; echo $opam_repository_tag")
+
+opam_repository_tag = (
+    subprocess.run(["bash", "get_opam_tag.sh"], stdout=subprocess.PIPE)
```
pasqu4le, line 35: I think that the file could be avoided with `shell=True`:
```suggestion
get_opam_tag = "source ./tezos/scripts/version.sh; echo $opam_repository_tag"
opam_repository_tag = (
    subprocess.run(get_opam_tag, stdout=subprocess.PIPE, shell=True)
```

pasqu4le (commented):
One suggestion, also the commit message says:
>  last release our way to parse tag from `version.sh` is not relevant anymore.

but I think that it should be:
>  last release our way to parse tag from `version.sh` is no longer working.

because IMO "relevant" here is not the right word.

docker/fetch_tezos_sources.py
krendelhoff2: fixed and reworded

pasqu4le (approved):

PR #617, opened 2023-03-07
[Chore] Fix plugdev group add

krendelhoff2:
## Description

Problem: Currenly code regarding adding plugdev group does not working as expected, resulting in errors output.

Solution: Use parentheses to prioritize operations properly.

## Related issue(s)

Resolves #

  can be successfully built.


pasqu4le (approved):
One last thing, there is a typo in the commit message:
> does not working as expected

should be:
> does not work as expected

PR #623, opened 2023-03-13
[Chore] Output artifact list from container

krendelhoff2:
## Description

Problem: For now, we don't really know which artifacts were output
from container. This results in resigning issues because we just
list all files from output folder.

Solution: Get information about exact resulting artifacts from container.
## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/docker-tezos-packages.py
+++ b/docker/docker-tezos-packages.py
@@ -140,6 +140,8 @@ packages_to_build = get_packages_to_build(args.packages)
 if not args.build_sapling_package:
     packages_to_build.pop("tezos-sapling-params", None)
 
+artifacts = []
+
 for image in images:
     distros = [image] if target_os == "ubuntu" else distributions
     cmd_args = " ".join(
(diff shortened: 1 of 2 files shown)


docker/docker-tezos-packages.py
```diff
@@ -171,6 +173,11 @@ def validate_dists(dists, versions, target_os):
     """
     )
 
+    with open(os.path.join(args.output_dir, ".artifact_list"), "r") as f:
+        artifacts += f.read().splitlines()
+
+    call(f"rm -rf {os.path.join(args.output_dir, '.artifact_list')}")
```
pasqu4le, line 179: I believe that `-r` should be unnecessary here:
```suggestion
    call(f"rm -f {os.path.join(args.output_dir, '.artifact_list')}")
```

pasqu4le (commented):
This LGTM, I only have one comment

pasqu4le (approved):

PR #629, opened 2023-03-15
[#628] Preserve tezos git folder

### LINKED ISSUE #628: [BUG] Octez version not available tag 16.0-1
On ubuntu last release of serokell tezos package when i try to get octez node version it return :

```
octez-node --version                          
unknown (not-available) (0.0+dev)
```

Expected output is 16.0.

krendelhoff2:
## Description

Problem: Seems like octez binaries infer their version
from git tag, which get erased because `fetch_tezos_sources.py`
deletes `tezos/.git` folder, resulting in
`unknown (not-available) (0.0+dev)`
output instead of actual version.

Solution: Preserve `tezos/.git` folder.

## Related issue(s)

Resolves #628 

  can be successfully built.

### DIFF
--- a/docker/fetch_tezos_sources.py
+++ b/docker/fetch_tezos_sources.py
@@ -24,8 +24,6 @@ subprocess.run(
     ]
 )
 
-shutil.rmtree(os.path.join("tezos", ".git"))
-
 subprocess.run(["git", "clone", "https://gitlab.com/tezos/opam-repository.git"])
 
 opam_repository_tag = (


docker/fetch_tezos_sources.py
```diff
@@ -16,15 +16,12 @@
     [
         "git",
         "clone",
-        "--branch",
-        octez_version,
         "https://gitlab.com/tezos/tezos.git",
-        "--depth",
-        "1",
```
pasqu4le, line 23: There's a reason this was done like this, it takes much less space and I suppose less time to clone only one branch, instead of the whole repo.

Also, the commit doesn't really explain what is causing the issue, was this git command? Why? how does using `git checkout` solve the issue? :thinking:

docker/fetch_tezos_sources.py
```diff
@@ -40,6 +37,5 @@
 
 os.chdir("opam-repository")
 subprocess.run(["git", "checkout", opam_repository_tag])
-subprocess.run(["rm", "-rf", ".git"])
-subprocess.run(["rm", "-r", "zcash-params"])
```
pasqu4le, line 44: Likewise, this was done to save space AFAIK, why is it now necessary to avoid removing these subfolders?

docker/package/scripts/build-binary.sh
```diff
@@ -5,13 +5,20 @@
 set -euo pipefail
 
 export OPAMYES=true
-mkdir opamroot
+mkdir -p opamroot
```
pasqu4le, line 8: I'm confused by this change too.

This command creates a directory in the current working dir, how's it possible for its parent to not exist?
Did you make this to ignore the case in which `opamroot` was already there?

docker/package/scripts/build-binary.sh
```diff
@@ -5,13 +5,20 @@
...
 binary_name="$2"
 
+source ./tezos/scripts/version.sh
+
+cd opam-repository
+git checkout "${opam_repository_tag:-}"
+cd ..
+
 cd tezos
+git checkout "$OCTEZ_VERSION"
```
pasqu4le, line 21: Some comments on these changes would be nice too

PR #630, opened 2023-03-15
[#628] Preserve tezos git folder

### LINKED ISSUE #628: [BUG] Octez version not available tag 16.0-1

krendelhoff2:
## Description

Problem: Seems like octez binaries infer their version from git tag, which get erased because `fetch_tezos_sources.py` deletes `tezos/.git` folder, resulting in
`unknown (not-available) (0.0+dev)`
output instead of actual version.

Solution: Preserve `tezos/.git` folder.

## Related issue(s)

Resolves #628 

  can be successfully built.

### DIFF
--- a/docker/fetch_tezos_sources.py
+++ b/docker/fetch_tezos_sources.py
@@ -23,8 +23,8 @@ subprocess.run(
         "1",
     ]
 )
-
-shutil.rmtree(os.path.join("tezos", ".git"))
+# NOTE: it's important to keep the `tezos/.git` directory here, because the
+# git tag is used to set the version in the Octez binaries.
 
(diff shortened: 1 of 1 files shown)


docker/fetch_tezos_sources.py
```diff
@@ -24,8 +24,6 @@
     ]
 )
```
pasqu4le, line 26: We should really try to not forget this in the future.
```suggestion
# NOTE: it's important to keep the `tezos/.git` directory here, because the
# git tag is used to set the version in the Octez binaries.
```

pasqu4le (approved):
Tested out and this fix works, I only have one comment.

PR #638, opened 2023-03-24
[Chore] Add sapling package for fedora upload

krendelhoff2:
## Description

Problem: In contrast from launchpad, copr does not reuse sapling package from older releases. Thus
we should make sure that it is always available.

Solution: Upload sapling params every release among other packages.

## Related issue(s)

Resolves #

  can be successfully built.


pasqu4le:
Note: this was ultimately closed so that users don't end up updating their sapling-params package at every release.

PR #640, opened 2023-03-30
[Chore] Split build process

krendelhoff2:
## Description

Problem: it was decided to split `docker-tezos-packages.py`
script into several independent ones.

Solution: Create independent script for building, signing
and uploading.

## Related issue(s)

Resolves #

  can be successfully built.


docker/package/util/build.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+
+        call(f"{virtualisation_engine} rm -v {container_id}")
+
+        if exit_code:
+            print("Unrecoverable error occured.")
+            sys.exit(exit_code)
+
+        # the same source archive has to be reused for an ubuntu package on different distros
+        if sources_dir_name is None and target_os == "ubuntu":
```
pasqu4le, line 193: There's quite a bit of differences between the `ubuntu` and `fedora` builds at this point, I see a few `if`s that handle the two differently.

I'm thinking that it would be better to split the builds in two different python source files, both of which import a third file (non-executable) with the common code.

How does that sound to you?

docker/docker-tezos-packages.py
pasqu4le, line 1: While we are at it, what do you think about renaming this? `docker/docker-tezos-packages.py` is a bit too verbose and poorly named IMO, especially now that this is mainly an entry-point for end users.

I propose to change it to simply `docker/package.py`

.buildkite/pipeline-for-tags.yml
```diff
@@ -13,26 +13,73 @@ steps:
...
+ - label: Build ubuntu source packages
+   key: build-ubuntu-source-packages
    if: build.tag =~ /^v.*-1/
    agents:
      queue: "docker"
    commands:
    - eval "$SET_VERSION"
-   - nix develop .#docker-tezos-packages -c ./docker/docker-tezos-packages.py --os ubuntu --type source -s "Serokell <tezos-packaging@serokell.io>"
-   - nix develop .#docker-tezos-packages -c ./docker/docker-tezos-packages.py --os fedora --type source -s "Serokell <tezos-packaging@serokell.io>"
+   - nix develop .#docker-tezos-packages -c ./docker/package/util/build.py --os fedora --type source
```
pasqu4le, line 23: shouldn't this be:
```suggestion
   - nix develop .#docker-tezos-packages -c ./docker/package/util/build.py --os ubuntu --type source
```

docker/package/util/build.py
pasqu4le: @krendelhoff2 I see the new changes, but AFAICT there is still a common `build.py` file.

Have you considered splitting this?

.buildkite/pipeline-for-tags.yml
pasqu4le, line 23: shouldn't this be:
```suggestion
   - nix develop .#docker-tezos-packages -c ./docker/package/util/build.py --os ubuntu --type source
```

docker/package/util/build.py
pasqu4le: @krendelhoff2 I looked at the new changes and AFAICT there's still a common `build.py` for both (and the same goes for signing and uploading).

Have you considered my suggestion to split these?

To be clear: I see that you've split some functions to only do one thing and reduced the condition checks already, but I still think that there's some use in having these in separate file to avoid the argument parsing and make these files even more single-purpose.

docker/package/util/build.py
pasqu4le, line 1: Do we need this to be separate?

I mean, I doubt that we'll want to call this file directly, as it's not fine-grained as the two build ones nor it's multi-purpose like `docker/package.py`, so IMO we should consider moving its content to `docker/package.py` instead.

The same comment applies to `sign.py` and `upload.py`.

docker/package/util/build_fedora.py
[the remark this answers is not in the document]
krendelhoff2: I propose to move under `docker/build` directory then, since these scripts unrelated to packaging code

docker/package.py
```diff
@@ -0,0 +1,91 @@
...
+from build.fedora.sign import sign_fedora
+
+import build.util.sign as sign
+from build.ubuntu.sign import sign_ubuntu
+from build.fedora.sign import sign_fedora
+
+import build.util.upload as upload
+from build.ubuntu.upload import upload_ubuntu
+from build.fedora.upload import upload_fedora
+
```
pasqu4le, line 28: I suppose that there's some issue with the imports here, because when I try to call this script (regardless of the arguments I pass) I get an error that ends with:
```
argparse.ArgumentError: argument --output-dir: conflicting option string: --output-dir
```

pasqu4le (changes requested):
Aside from these two new comments there's also one last thing to be done: the commands in `docker/README.md` need to be updated to use `docker/package.py` instead of `docker/docker-tezos-packages.py`

docker/package.py
krendelhoff2: Some quirk of `argparse`. Got resolved by deepcopying `parser`.

docker/package.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2022 Oxhead Alpha
```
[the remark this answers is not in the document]
krendelhoff2: I accidentally squashed all `Make a wrapper script` fixups into initial commit, FYI

pasqu4le (approved):

Issue #646, opened 2023-04-24
[BUG] NixOS modules'


rvem:
# Description

# To Reproduce

# Expected behavior

# Environment

- OS
- branch/revision

# Additional context

krendelhoff2:
Issue is resolved after #657 merged

PR #657, opened 2023-05-15
[#646] Consider enable option in modules

krendelhoff2:
## Description

Problem: if you attempt to define a configuration using NixOS modules but decided to disable it, it will still attempt to start module.

Solution: Do not run service on startup if it is disabled and tweak test script according to services' enable option.

## Related issue(s)

Resolves #

  can be successfully built.


rvem (commented):
Hmm, usually, when `enable = false` corresponding `config` part is not generated at all

rvem (approved):

PR #658, opened 2023-05-19
[#643] Make `tezos-baking` separate package

### LINKED ISSUE #643: Move the setup and voting wizards to a new project

Currently the source code of our setup and and voting wizards is in the `docker/package/` directory and it's part of the project there.

This is quite confusing and incorrect, as their code is _packaged_, but isn't part of the packaging logic (which is what that python project is about) and moreover they aren't really related to `docker` either.

We should move them to their own package, to ensure that they remain (as much as possible) independent and in general to structure this repository in a better way.

# Acceptance criteria

The setup and voting wizard have their own python package.

krendelhoff2:
## Description

Problem: Currently the source code of our setup and voting wizards is in the docker/package/ directory and it's part of the project there. This is quite confusing and incorrect, as their code is packaged, but isn't part of the packaging logic.

Solution: Move tezos-baking sources to their own package.

## Related issue(s)

Resolves #643

  can be successfully built.


tezos_baking/__init__.py
pasqu4le, line 1: With the new commit we are closer to a correct change, but I think that we should strive to turn this into a proper python project.

Mainly what I mean is that it should follow the standard practices (see the ["Packaging Python Packages" doc](https://packaging.python.org/en/latest/tutorials/packaging-projects/)) as one would for any other python project.

For example, there should be an `src` subdirectory, this `__init__.py` doesn't need a shebang, it should be possible to build it and its dependencies with `pip`, there should be a README with instructions on how to do so, etc.

docker/package/Dockerfile-fedora
```diff
@@ -27,6 +27,7 @@ COPY protocols.json /tezos-packaging/protocols.json
 COPY docker/package/*.py /tezos-packaging/docker/package/
 COPY docker/package/defaults /tezos-packaging/docker/package/defaults
 COPY docker/package/scripts /tezos-packaging/docker/package/scripts
+COPY tezos_baking /tezos-packaging/docker/package/tezos_baking
```
pasqu4le, line 30: IMO `tezos_baking` is not the right top-level directory nor name for this new project, especially because the point of this issue is to distance it from the `tezos-baking` package

pasqu4le (changes requested):

docker/package/Dockerfile-fedora
pasqu4le: @krendelhoff2 I see that you've changed this, but `code` is also not that great of a name for this, as it's too generic and also because it doesn't help with the separation that we are looking for.

So, how about simply calling the top-level dir `baking` (but keeping `tezos-baking` as the python package name, as you did now)?

code/src/tezos_baking/__init__.py
```diff
@@ -0,0 +1,2 @@
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
```
pasqu4le, line 2: From the documentation:
> `__init__.py` is required to import the directory as a package, and should be empty.

so let's remove this REUSE header and instead add an entry in `.reuse/dep5`

code/pyproject.toml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+[build-system]
+requires = ["setuptools"]
```
pasqu4le, line 5: I see that the tutorial puts a lower-bound limit on the dependency
```toml
requires = ["setuptools>=61.0"]
```
can't we do the same?

code/pyproject.toml
```diff
@@ -0,0 +1,6 @@
...
+build-backend = "setuptools.build_meta"
```
pasqu4le, line 6: I think that some [project metadata](https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata) is missing here.

code/README.md
```diff
@@ -0,0 +1,13 @@
+<!--
```
pasqu4le, line 1: I think that a `LICENSE` file is also missing. It should be a copy of the `LicenseRef-MIT-OA` one.

It may seem unnecessary, but again we should try to make this as stand-alone as possible.
Moreover, as per documentation:
> Most build backends automatically include license files in packages. See your backend’s documentation for more details.

code/pyproject.toml
krendelhoff2: Do you think we should duplicate it from `setup.cfg` file?

pasqu4le: AFAIK `setup.cfg` was an artifact of `setuptools` and is no longer required for configuring a package (but `setuptools` and other tools maybe still use it).

The `pyproject.toml` file was introduced more recently and is supposed to not be specific to any single build tool. 

So, if it's true that `setuptools` works the same without `setup.cfg`, then I'd move the data there to `pyproject.toml`. Otherwise, yes, I'd like to have that duplicated.

.reuse/dep5
```diff
@@ -1,5 +1,5 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 
-Files: .github/* nix/nix/* nix/build/install_topfind_196.patch protocols.json meta.json tests/buildkite/golden/* tests/host.key tests/host.cert tests/ca.cert
+Files: .github/* nix/nix/* nix/build/install_topfind_196.patch protocols.json meta.json tests/buildkite/golden/* tests/host.key tests/host.cert tests/ca.cert baking/src/tezos_baking/__init__.py
 Copyright: 2022 Oxhead Alpha
```
pasqu4le, line 4: This is the wrong year, you need to add a new entry here instead

baking/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+Copyright (c) 2021-2022 Oxhead Alpha
```
pasqu4le, line 2: Please update this to:
```suggestion
Copyright (c) 2021-2023 Oxhead Alpha
```
and do the same in the original file too

code/src/tezos_baking/__init__.py
krendelhoff2: BTW, seems like it's not really true that `__init__.py` file should be empty, see [here](https://docs.python.org/3/tutorial/modules.html#packages).
> The __init__.py files are required to make Python treat directories containing the file as packages. This prevents directories with a common name, such as string, from unintentionally hiding valid modules that occur later on the module search path. In the simplest case, __init__.py can just be an empty file, but it can also execute initialization code for the package or set the __all__ variable, described later.

So example, it's idiomatic way to set package's version in code, see [here](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/).

baking/LICENSE
pasqu4le: > do the same in the original file too

@krendelhoff2 I think that you missed this part

code/src/tezos_baking/__init__.py
pasqu4le: Well, sure, `__init__.py` can be non-empty, but it should be if we follow the documentation (especially for the simplest cases, like our own).

> it's idiomatic way to set package's version in code, see here.

I wouldn't call that the idiomatic way, at least not anymore, given that:
1. it's the 6th option in that list
2. it's the only one warning that it will fail to work in some cases 

Given that we are currently matching version number with the package, I think we can get the version from the same `meta.json` that we use everywhere else (which is kinda what option 4 is BTW).
However TBH I don't see any need to do this right now, as we don't need that version number in the wizard's code at all and this info is just metadata at this point.

So, IMO this change should be reverted and once again `__init__.py` left empty.

P.s. even if we did need to set a version there, it shouldn't be `v17.1` as that's the version of Octez (that we don't need to match, as it's unrelated to this) and not the package version (that's `17.1-1` and we also don't need to match that).

.reuse/dep5
```diff
@@ -3,3 +3,7 @@ Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Files: .github/* nix/nix/* nix/build/install_topfind_196.patch protocols.json meta.json tests/buildkite/golden/* tests/host.key tests/host.cert tests/ca.cert
 Copyright: 2022 Oxhead Alpha
 License: LicenseRef-MIT-OA
+
+Files: nix/build/install_topfind_196.patch baking/src/tezos_baking/__init__.py
```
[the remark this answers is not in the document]
krendelhoff2: This file is used by `nix/build/ocaml-overlay.nix` and indeed was added to the repo in 2023, so it's right place for it

baking/setup.cfg
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+[metadata]
+name = tezos-baking
+author = Serokell
+author_email = tezos-packaging@serokell.io
+description = Package that provides systemd services that orchestrate other services from Tezos packages
+license = LicenseRef-MIT-OA
```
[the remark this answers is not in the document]
krendelhoff2: It works, but only since `61.0` (see [here](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)).
While `focal` has [package](https://packages.ubuntu.com/focal/python3-setuptools) with much lower version.
Current setup builds fine on both `ubuntu` and `fedora`, as well as local install.

docker/package/model.py
```diff
@@ -681,11 +681,10 @@ def __init__(
 
     def fetch_sources(self, out_dir, binaries_dir=None):
         os.makedirs(out_dir)
-        package_dir = out_dir + "/tezos_baking"
-        os.makedirs(package_dir)
-        shutil.copy(f"{os.path.dirname(__file__)}/wizard_structure.py", package_dir)
-        shutil.copy(f"{os.path.dirname(__file__)}/tezos_setup_wizard.py", package_dir)
-        shutil.copy(f"{os.path.dirname(__file__)}/tezos_voting_wizard.py", package_dir)
+        package_dir = out_dir
```
pasqu4le, line 684: The `package_dir` variable seems unnecessary to me now.

scripts/update-tezos.sh
```diff
@@ -30,6 +30,9 @@ our_tezos_tag="$(jq -r '.tezos_ref' meta.json | cut -d'/' -f3)"
 new_meta=$(jq ".tezos_ref=\"$latest_upstream_tag\"" meta.json)
 echo "$new_meta" > meta.json
 
+# Update version of tezos-baking package
+sed -i "s/version = .*/version = \"$latest_upstream_tag\"/" ./code/pyproject.toml
```
pasqu4le, line 34: IMO this should be moved a few lines below and be performed with the `release` `sed` command, so that it's committed with it.

docker/package/model.py
```diff
@@ -681,7 +681,6 @@ def __init__(
 
     def fetch_sources(self, out_dir, binaries_dir=None):
         os.makedirs(out_dir)
-        package_dir = out_dir
         subprocess.call(
             f"cp -r {os.path.dirname(__file__)}/baking/* {package_dir}/", shell=True
```
pasqu4le, line 685:
```suggestion
            f"cp -r {os.path.dirname(__file__)}/baking/* {out_dir}/", shell=True
```

pasqu4le: BTW, aside from this comment the rest LGTM, so feel free to squash the changes as well @krendelhoff2

pasqu4le (approved):

Issue #651, opened 2023-05-02
Log wizard's actions


pasqu4le:
# Clarification and motivation

Currently we have two wizards, `tezos-setup` and `tezos-vote`, neither of which saves any logs.

It would be nice to have these logs to:
1. keep a record of what the user did, which can help with error reporting and debugging
2. parse them elsewhere to know what was done, instead of relying on the state of systemd services alone

For them to be useful, these logs need to be kept for a few days at least, but we should avoid keeping them forever.

# Acceptance criteria

User choices and actions performed by the wizards are logged and kept for some time.

Issue #634, opened 2023-03-22
Update maintenance guide(s) 


pasqu4le:
# Clarification and motivation

In our `CONTRIBUTING.md` doc we have two brief guides for maintenance, a "quick" one and a "nix specific" one.

Both of them haven't been updated in a while so they need to be checked and, if necessary, updated.

# Acceptance criteria

The maintenance guide(s) in our contribution doc are up-to-date.

PR #660, opened 2023-05-24
[#651] Log wizards actions

krendelhoff2:
## Description

Problem: Currently we have two wizards, tezos-setup and tezos-vote,
neither of which saves any logs. It's useful sometimes to know
the configuration of the last wizards' call.

Solution: Save the log file with setup configuration.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/baking/src/tezos_baking/steps.py
+++ b/baking/src/tezos_baking/steps.py
@@ -215,3 +215,18 @@ def get_ledger_derivation_query(ledgers_derivations, node_endpoint, client_dir):
             ]
         ),
     )
+
+
+replace_key_options = {
+    "no": "Keep the existing key",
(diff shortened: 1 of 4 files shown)


pasqu4le (changes requested):
A few comments on the general approach here:
1. IMO the logging should be handled by the wizards themselves rather than the `tezos-baking` package, because (especially with #634 in mind) it's not a responsibility of the package
2. The standard logging library from python is quite powerful and has a lot of options (including how to handle rotations AFAIK), so we should use that here
3. IMO every choice taken by the user should be logged (without exposing sensitive info) as well as any error

DMozhevitin (approved):
LGTM

Issue #644, opened 2023-04-18
Allow the setup wizard to resume snapshot downloads


pasqu4le:
# Clarification and motivation

Currently the setup wizard, as part of the setup process, is able to download a node snapshot from snapshot metadata or from a URL given by the user.

However, if this download fails (e.g. for network issues) this step fails and the user needs to download a snapshot again.
This can be very inconvenient as some snapshots are very large and downloading them can take a long time.

We should modify the wizard and give it the capability to resume a download (where possible).
I propose to change the wizard so that, when a download fails and it falls back to the snapshot import step, a new option is there to resume the download (using the last URL given by the user or found in the snapshot metadata). 

# Acceptance criteria

It's possible to resume failed downloads in the setup wizard.

pasqu4le:
Resolved by #661

PR #661, opened 2023-05-25
[#644] Allow setup to resume snapshot download

krendelhoff2:
## Description

Problem: Currently, if node snapshot download fails (e.g. for network issues)
this step fails and the user needs to download a snapshot again. This can be very inconvenient as some snapshots are very large and downloading them can take a long time.

Solution: Use `wget` with `--continue` option to be able to resume download.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -232,6 +232,14 @@ snapshot_url_query = Step(
     validator=Validator([required_field_validator, reachable_url_validator()]),
 )
 
+snapshot_sha256_query = Step(
+    id="snapshot_sha256",
+    prompt="Provide the sha256 of the node snapshot file. (optional)",
+    help="With sha256 provided, an integrity check will be performed for you.\n"
+    "Also, it will be possible to resume incomplete snapshot downloads.",
+    default=None,
+)
+
 history_mode_query = Step(
     id="history_mode",
     prompt="Which history mode do you want your node to run in?",
@@ -349,6 +357,7 @@ class Setup(Setup):
             )
 
             self.config["snapshot_url"] = snapshot_metadata["url"]
+            self.config["snapshot_sha256"] = snapshot_metadata.get("sha256", None)
             self.config["snapshot_block_hash"] = snapshot_metadata["block_hash"]
         except (urllib.error.URLError, ValueError):
             print(f"Couldn't collect snapshot metadata from {json_url}")
@@ -396,7 +405,10 @@ class Setup(Setup):
             elif self.config["snapshot"] == "url":
                 self.query_step(snapshot_url_query)
                 try:
-                    snapshot_file = fetch_snapshot(self.config["snapshot_url"])
+                    self.query_step(snapshot_sha256_query)
+                    url = self.config["snapshot_url"]
+                    sha256 = self.config["snapshot_sha256"]
+                    snapshot_file = fetch_snapshot(url, sha256)
                 except (ValueError, urllib.error.URLError):
                     print()
                     print("The snapshot URL you provided is unavailable.")
@@ -405,9 +417,10 @@ class Setup(Setup):
                     continue
             else:
                 url = self.config["snapshot_url"]
+                sha256 = self.config["snapshot_sha256"]
                 snapshot_block_hash = self.config["snapshot_block_hash"]
                 try:
-                    snapshot_file = fetch_snapshot(url)
+                    snapshot_file = fetch_snapshot(url, sha256)
                 except (ValueError, urllib.error.URLError):
                     print()
                     print("The snapshot download option you chose is unavailable,")
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -52,7 +52,7 @@ toggle_vote_modes = {
 }
 
 
-TMP_SNAPSHOT_LOCATION = "/tmp/octez_node.snapshot"
+TMP_SNAPSHOT_LOCATION = "/tmp/octez_node.snapshot.d/"
 
 
 # Wizard CLI utility
@@ -74,10 +74,66 @@ Type in 'exit' to quit.
 """
 
 
-def fetch_snapshot(url):
+def fetch_snapshot(url, sha256=None):
+
+    dirname = TMP_SNAPSHOT_LOCATION
+    filename = os.path.join(dirname, "octez_node.snapshot")
+    metadata_file = os.path.join(dirname, "octez_node.snapshot.sha256")
+
+    # updates or removes the 'metadata_file' containing the snapshot's SHA256
+    def dump_metadata(metadata_file=metadata_file, sha256=sha256):
+        if sha256:
+            with open(metadata_file, "w+") as f:
+                f.write(sha256)
+        else:
+            try:
+                os.remove(metadata_file)
+            except FileNotFoundError:
+                pass
+
+    # reads `metadata_file` if any or returns None
+    def read_metadata(metadata_file=metadata_file):
+        if os.path.exists(metadata_file):
+            with open(metadata_file, "r") as f:
+                sha256 = f.read()
+            return sha256
+        else:
+            return None
+
+    def download(filename=filename, url=url, args=""):
+        from subprocess import CalledProcessError
+
+        try:
+            proc_call(f"wget {args} --show-progress -O {filename} {url}")
+        except CalledProcessError as e:
+            # see here https://www.gnu.org/software/wget/manual/html_node/Exit-Status.html
+            if e.returncode >= 4:
+                raise urllib.error.URLError
+            else:
+                raise e
+
     print("Downloading the snapshot from", url)
-    filename = TMP_SNAPSHOT_LOCATION
-    urllib.request.urlretrieve(url, filename, progressbar_hook)
+
+    # expected for the (possibly) existing chunk
+    expected_sha256 = read_metadata()
+
+    os.makedirs(dirname, exist_ok=True)
+    if sha256 and expected_sha256 and expected_sha256 == sha256:
+        # that case means that the expected sha256 of snapshot
+        # we want to download is the same as the expected
+        # sha256 of the existing octez_node.snapshot file
+        # when it will be fully downloaded
+        # so that we can safely use `--continue` option here
+        download(args="--continue")
+    else:
+        # all other cases we just dump new metadata
+        # (so that we can resume download if we can ensure
+        # that existing octez_node.snapshot chunk belongs
+        # to the snapshot we want to download)
+        # and start download from scratch
+        dump_metadata()
+        download()
+
     print()
     return filename
 
@@ -383,7 +439,7 @@ class Setup(Setup):
             print("Snapshot imported.")
 
             try:
-                os.remove(TMP_SNAPSHOT_LOCATION)
+                shutil.rmtree(TMP_SNAPSHOT_LOCATION)
             except:
                 pass
             else:
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -138,6 +138,29 @@ def fetch_snapshot(url, sha256=None):
     return filename
 
 
+class Sha256Mismatch(Exception):
+    "Raised when the actual and expected sha256 don't match."
+
+    def __init__(self, actual_sha256=None, expected_sha256=None):
+        self.actual_sha256 = actual_sha256
+        self.expected_sha256 = expected_sha256
+
+
+def check_file_contents_integrity(filename, sha256):
+    import hashlib
+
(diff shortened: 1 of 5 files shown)


docker/package/tezos_setup_wizard.py
```diff
@@ -77,7 +77,7 @@
 def fetch_snapshot(url):
     print("Downloading the snapshot from", url)
     filename = TMP_SNAPSHOT_LOCATION
-    urllib.request.urlretrieve(url, filename, progressbar_hook)
+    proc_call(f"wget --show-progress --continue -O {filename} {url}")
```
pasqu4le, line 80: AFAIU there's no built-in way to do this with `urllib`, but it should also be implementable.

Have you considered doing that? If so, were there problems with it?

krendelhoff2: I considered this, but it turns out to be much more verbose and I don't really see any benefit from it, `wget` interface and output seems to be very descriptive and download resume is working fine.

pasqu4le: OK, I'm not against it, but in this case `wget` should be listed as a run dependency for the `tezos-baking` package

docker/package/tezos_setup_wizard.py
```diff
@@ -77,7 +77,7 @@
...
     print()
     return filename
```
pasqu4le, line 82: Have you considered what will happen to this in all the cases too?

For example, I didn't test this, but here's some questions worth answering:
1. is it possible to resume a download from a previous `tezos-setup` run?
2. what happens when there is still an incomplete snapshot on disk and this code is run?
3. what happens if the user changes their URL or decides to switch from using the metadata to a URL?
4. what if the latest snapshot from the metadata is pointing at a different one?

docker/package/tezos_setup_wizard.py
krendelhoff2: It turns out to be non-trivial to make properly with our code now, but since I'm reconsidering `tezos-baking` packaging now I'll figure out how to split binaries packaging (and its deps) and `tezos-baking` ones

docker/package/tezos_setup_wizard.py
krendelhoff2: 1. Yes, AFAIU [download resume works on server-side](https://stackoverflow.com/questions/12434306/in-general-how-does-resuming-a-download-work), so it doesn't matter how and why download stops and when and how we resume it. It just will check the state of file and will try to do as little of work as possible.
2. If you mean it's totally different snapshot, it will be wiped in favor of file we want to download.
3 and 4 questions is irrelevant IIUC, because this code works already knowing the url of snapshot we want to download, disregarding the source

pasqu4le: > it doesn't matter how and why download stops and when and how we resume it

It does matter if the partial file on disk is not the same file we are trying to download anymore.

As mentioned in your link, the request to use a specific range is how download resuming works, but that assumes that we are indeed re-downloading the same file. If we "resume" the wrong download then we'll have a file in the end that's half our first attempt and half the second one.

I asked:
> is it possible to resume a download from a previous tezos-setup run?

both as a negative scenario and a positive one. 
IOW the question is: is it possible to incorrectly resume a download that was interrupted from a previous run?
Or will the previous snapshot be deleted in favor of a new download?
And is it possible to correctly resume a download that was interrupted in a previous run?

> If you mean it's totally different snapshot, it will be wiped in favor of file we want to download

Great, that's what should happen :+1: 

How do we decide if the existing incomplete file is from a "totally different snapshot" tho?
Can we also distinguish between different snapshots that are served on the same URL (at different times, e.g. if the URL given by the user is for the "latest" snapshot)?

P.s. Question 3 and 4 are basically reiteration of the same ones.
I wanted to point out that we need to handle all the different cases in which we are trying to resume a download.

krendelhoff2: > IOW the question is: is it possible to incorrectly resume a download that was interrupted from a previous run?

Seems like, it is. Thanks for noticing it.

docker/package/tezos_setup_wizard.py
```diff
@@ -74,10 +74,55 @@
 """
 
 
-def fetch_snapshot(url):
+def fetch_snapshot(url, sha256=None):
+
+    filename = os.path.join(TMP_SNAPSHOT_LOCATION, "octez_node.snapshot")
+    dirname = TMP_SNAPSHOT_LOCATION
+    metadata_file = os.path.join(dirname, "metadata.json")
```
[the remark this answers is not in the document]
krendelhoff2: I did this at first, but this seems to me cleaner and more extendable.
I also had a hard time trying to name this file, `octez_node.snapshot.sha256` looks gross.

docker/package/tezos_setup_wizard.py
```diff
@@ -293,6 +338,7 @@ def get_snapshot_link(self):
             )
 
             self.config["snapshot_url"] = snapshot_metadata["url"]
+            self.config["snapshot_sha256"] = snapshot_metadata.get("sha256", None)
```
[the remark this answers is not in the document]
krendelhoff2: > If there is no better option than using the file SHA256

I don't see any better option. As you already mentioned, `url` is not reliable source of truth, since its contents could be changed. 

We indeed can't understand if the file we download is the same as the file we intend to download until we fully download it. So that we can only rely on [metadata we can trust](xtz-shots.io/tezos-snapshots.json).

It's possible to ask `sha256` from user, but I believe then we should also check snapshot for integrity.

docker/package/tezos_setup_wizard.py
```diff
@@ -74,10 +74,55 @@
...
+                metadata = json.loads(f.read())
+            return metadata
+        else:
+            return None
+
+    def download(filename=filename, url=url):
+        proc_call(f"wget --show-progress -O {filename} {url}")
+
+    def continue_download(filename=filename, url=url):
+        proc_call(f"wget --continue --show-progress -O {filename} {url}")
```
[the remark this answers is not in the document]
krendelhoff2: It's wrapped in `check_call` invocation. AFAIU, previous code was handled about the same (no exception handling).

docker/package/tezos_setup_wizard.py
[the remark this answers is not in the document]
krendelhoff2: It is exactly what was done in recent commits

docker/package/tezos_setup_wizard.py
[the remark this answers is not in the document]
krendelhoff2: ah, I didn't get your idea then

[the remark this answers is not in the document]
krendelhoff2: You want `octez_node.snapshot` to be named with its sha256, or a separate file?

pasqu4le (commented):
I've found a bug regarding the handling of the SHA file.

Here's a way to reproduce it:
- run `tezos-setup` and for the snapshot use a URL and provide the correct SHA
- interrupt the download in the middle (with Ctrl-C)
- run `tezos-setup` again, give a different URL and provide no SHA
- interrupt this download in the middle again
- run `tezos-setup` one last time and give it the same URL and SHA as the first step

The expected behavior of the last run is that the file is downloaded from scratch, but instead the download is resumed from the wrong file (the one from the URL given in the third step).

In general, we should make sure that if there is a SHA it corresponds to the (possibly incomplete) snapshot file.

krendelhoff2:
Indeed, before second download, first download's sha256 should have been invalidated.

docker/package/model.py
```diff
@@ -303,11 +303,11 @@ def fetch_sources(self, out_dir, binaries_dir=None):
...
 
-    def gen_control_file(self, deps, ubuntu_version, out):
-        str_build_deps = ", ".join(deps)
-        str_additional_native_deps = ", ".join(
-            self.__get_os_specific_native_deps("ubuntu")
+    def gen_control_file(self, build_deps, run_deps, ubuntu_version, out):
+        str_run_deps = ", ".join(
+            run_deps + self.__get_os_specific_native_deps("ubuntu")
         )
+        str_build_deps = ", ".join(build_deps)
```
pasqu4le, line 310: It looks to me like something went wrong with the deps handling, because I noticed that installing `tezos-baking` now requires more packages to be installed (even tho it doesn't seem like they are actually needed).

krendelhoff2: What you see installed is packages' listed runtime deps, which for whatever reason weren't ever installed on ubuntu (but were on fedora)

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -74,14 +74,89 @@
 """
 
 
-def fetch_snapshot(url):
+def fetch_snapshot(url, sha256=None):
+
+    filename = os.path.join(TMP_SNAPSHOT_LOCATION, "octez_node.snapshot")
+    dirname = TMP_SNAPSHOT_LOCATION
```
pasqu4le, line 80:
```suggestion
    dirname = TMP_SNAPSHOT_LOCATION
    filename = os.path.join(dirname, "octez_node.snapshot")
```

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -74,14 +74,89 @@
...
+    metadata_file = os.path.join(dirname, "octez_node.snapshot.sha256")
+
+    # creating closures
+    def dump_metadata(metadata_file=metadata_file, sha256=sha256):
+        if sha256:
+            with open(metadata_file, "w+") as f:
+                f.write(sha256)
+        else:
+            proc_call(f"rm -f {metadata_file}")
```
pasqu4le, line 89:
```suggestion
            os.remove(metadata_file)
```

krendelhoff2: it throws exception when file does not exist, but we don't care if it does not exist. `proc_call`  looks much cleaner here

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -176,6 +252,14 @@ def get_snapshot_mode_query(modes):
     validator=Validator([required_field_validator, reachable_url_validator()]),
 )
 
+snapshot_sha256_query = Step(
+    id="snapshot_sha256",
+    prompt="Provide the sha256 of the node snapshot file. (optional)",
+    help="With sha256 provided integrity check will be performed for you.\n"
+    "Also, download resume option will be available.",
```
pasqu4le, line 259:
```suggestion
    prompt="Provide the sha256 of the node snapshot file. (optional)",
    help="With sha256 provided, an integrity check will be performed for you.\n"
    "Also, it will be possible to resume incomplete snapshot downloads.",
```

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -202,6 +286,20 @@ def get_key_mode_query(modes):
     )
 
 
+ignore_hash_mismatch_options = {
+    "no": "Abort that snapshot and return to the previous step",
+    "yes": "Continue setup with that snapshot",
```
pasqu4le, line 291:
```suggestion
    "no": "Discard the snapshot and return to the previous step",
    "yes": "Continue the setup with this snapshot",
```

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -202,6 +286,20 @@ def get_key_mode_query(modes):
...
+}
+
+ignore_hash_mismatch_query = Step(
+    id="ignore_hash_mismatch",
+    prompt="Do you want to proceed with that?",
+    help="It's possible to proceed if actual sha256 is good enough for you.",
```
pasqu4le, line 297:
```suggestion
    prompt="Do you want to proceed with this snapshot anyway?",
    help="It's possible, but not recommended, to ignore the sha256 mismatch and use this snapshot anyway.",
```

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -74,14 +74,90 @@
...
+    filename = os.path.join(dirname, "octez_node.snapshot")
+    metadata_file = os.path.join(dirname, "octez_node.snapshot.sha256")
+
+    # updates or removes the 'metadata_file' containing the snapshot's SHA256
+    def dump_metadata(metadata_file=metadata_file, sha256=sha256):
+        if sha256:
+            with open(metadata_file, "w+") as f:
+                f.write(sha256)
+        elif os.path.exists(metadata_file):
+            os.remove(metadata_file)
```
pasqu4le, line 89: This is not the best solution to [my previous comment](https://github.com/serokell/tezos-packaging/pull/661#discussion_r1267973197) because it's possible for `os.remove` to still throw an exception here.

The fact that `metadata_file` existed when `os.path.exists` is called does not guarantee that it still exists when `os.remove` is called.

So, as much as it is unlikely to happen and even tho it's uglier to see in code, let's use exception handling

pasqu4le (approved):

PR #662, opened 2023-05-29
backing.md: Add tezos/ to directory path.

philsaxton:
I was playing around with this and found `.tezos-client` was in /var/lib/tezos by default.


pasqu4le (approved):
:+1:  thanks for spotting and fixing this

PR #669, opened 2023-06-12
[Chore] Fix fedora upload

krendelhoff2:
## Description

Problem: Albali machine provides copr token only
under `copr-uploader` user, which is different from
buildkite one. Using absolute paths make impossible
for `copr-cli` to upload that files under `sudo -u`.

On the other hand, it doesn't really change script's
intended behavior.

Solution: Use relative paths to artifacts.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/build/util/upload.py
+++ b/docker/build/util/upload.py
@@ -15,8 +15,7 @@ parser.add_argument(
     "--directory",
     "-d",
     help="provide a directory with packages to upload",
-    default=f"{os.path.join(os.getcwd(), '.')}",
-    type=os.path.abspath,
+    default=".",
 )
 parser.add_argument(
     "--os",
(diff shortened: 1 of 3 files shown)


docker/build/util/upload.py
```diff
@@ -15,8 +15,7 @@
     "--directory",
     "-d",
     help="provide a directory with packages to upload",
-    default=f"{os.path.join(os.getcwd(), '.')}",
-    type=os.path.abspath,
+    default=".",
```
pasqu4le, line 18: Why not:
```suggestion
    default=".",
    type=os.path.relpath,
```
?

I didn't try it, but supposedly it should work

pasqu4le (approved):
One minor comment from me, but approving already

docker/build/util/upload.py
krendelhoff2: It still could happen that user wants to supply absolute path and it would lead to wrong behavior
We doesn't want to force it to use relative paths, we just don't force it to use absolute ones

PR #672, opened 2023-06-14
[#671] Fix epel chroot name

### LINKED ISSUE #671: [BUG] Epel packages are not published by CI
# Description

Epel packages should be uploaded by the `Publish epel packages` step of the pipeline for tags, but they fail with:
```
Uploading package srcs/epel/tezos-baking-17.0a-2.src.rpm
Bad request parameters: {'chroots': ["'epel-x86_64' is not a valid choice for this field."]}
```
see [an example here](https://buildkite.com/serokell/tezos-packaging/builds/3837#0188b646-7b15-4f4f-8509-4fbe4e6fc74a).

Moreover, the pipeline step reports a success even tho it obviously didn't succeed at its task.

# To Reproduce

Run the `Publish epel packages` step, in the version from the [v17.0-2 tag](https://github.com/serokell/tezos-packaging/releases/tag/v17.0-2).

# Expected behavior

1. the epel packages produced are actually published by CI
2. in case of failure to publish, the step reports a failure too

krendelhoff2:
## Description

Problem: copr epel chroot name lacks of version
number, resulting in error during upload.

Solution: Specify correct name.

## Related issue(s)

Resolves #671 

  can be successfully built.

### DIFF
--- a/docker/build/fedora/sign.py
+++ b/docker/build/fedora/sign.py
@@ -24,7 +24,7 @@ def sign_fedora(args: Arguments):
 
     for f in artifacts:
         if f.endswith(".src.rpm"):
-            subprocess.call(
+            subprocess.check_call(
                 f'rpmsign --define="%_gpg_name {identity}" --define="%__gpg {gpg}" --addsign {f}',
                 shell=True,
             )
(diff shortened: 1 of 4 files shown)


pasqu4le (approved):
LGTM

PR #677, opened 2023-07-17
[#641] Add tezos binaries tests

### LINKED ISSUE #641: Test packages installation for linux distros

We publish native package for several versions of linux distro, but we don't always test all of them after publication.

For example, we recently had to upload the `tezos-sapling-params` package for some of our supported Fedora versions, the absence of which was causing most of our packages to not be installable.

We should prevent this by testing that the packages can be installed and their software executed (at least by running the `--version` option) on the regular.

It probably makes sense to automatically do this right after a new publication or perhaps on a schedule.

# Acceptance criteria

The latest packages published for linux distros at any given time are tested in all supported versions by CI, at least once, both for stable releases and RCs.

PruStephan:
## Description
This PR adds checks that Tezos binaries can be installed on the latest Ubuntu and Fedora can be installed and work fine. The tests run once every day.

## Related issue(s)

Resolves #641 

  can be successfully built.

### DIFF
--- a/.github/workflows/test-binaries.yml
+++ b/.github/workflows/test-binaries.yml
@@ -1,0 +1,19 @@
+name: Test Tezos binaries
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  test_binaries:
+    name: Install and test binaries
+    runs-on: [self-hosted, Linux, X64, nix-with-docker]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Test fedora binaries
+        run: nix develop .#buildkite -c ./docker/tests/scripts/test-fedora-binaries.sh
+
+      - name: Test ubuntu binaries
+        run: nix develop .#buildkite -c ./docker/tests/scripts/test-ubuntu-binaries.sh
--- a/docker/tests/Dockerfile-fedora-test
+++ b/docker/tests/Dockerfile-fedora-test
@@ -1,0 +1,19 @@
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+ARG dist
+FROM fedora:${dist}
+
+WORKDIR /tezos-packaging/docker
+
+RUN dnf update -y
+RUN dnf install -y python3-devel python3-setuptools 'dnf-command(copr)'
+
+ARG repo
+RUN dnf copr enable -y @Serokell/${repo}
+
+ENV IS_RELEASED=${repo}
+
+COPY docker/tests/test-fedora-binaries.py /tezos-packaging/docker/tests/test-fedora-binaries.py
+COPY docker/tests/binaries.json /tezos-packaging/binaries.json
+CMD [ "python3", "/tezos-packaging/docker/tests/test-fedora-binaries.py"]
--- a/docker/tests/Dockerfile-ubuntu-test
+++ b/docker/tests/Dockerfile-ubuntu-test
@@ -1,0 +1,19 @@
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+ARG dist
+FROM ubuntu:${dist}
+
+WORKDIR /tezos-packaging/docker
+
+RUN apt update -y
+RUN apt install -y python3-all python3-setuptools software-properties-common
+
+ARG repo
+RUN add-apt-repository -yu ppa:serokell/${repo}
+
+ENV IS_RELEASED=${repo}
+
+COPY docker/tests/test-ubuntu-binaries.py /tezos-packaging/docker/tests/test-ubuntu-binaries.py
+COPY docker/tests/binaries.json /tezos-packaging/binaries.json
+CMD [ "python3", "/tezos-packaging/docker/tests/test-ubuntu-binaries.py"]
--- a/docker/tests/binaries.json
+++ b/docker/tests/binaries.json
@@ -1,0 +1,41 @@
+{
+    "released": [
+        "tezos-smart-rollup-client-Proxford",
+        "tezos-smart-rollup-client-PtNairob",
+        "tezos-smart-rollup-node-Proxford",
+        "tezos-smart-rollup-node-PtNairob",
+        "tezos-baking",
+        "tezos-sapling-params",
+        "tezos-accuser-Proxford",
+        "tezos-baker-Proxford",
+        "tezos-accuser-PtNairob",
+        "tezos-baker-PtNairob",
+        "tezos-node",
+        "tezos-dac-node",
+        "tezos-dac-client",
+        "tezos-codec",
+        "tezos-signer",
+        "tezos-admin-client",
+        "tezos-client",
+        "tezos-smart-rollup-wasm-debugger"
+    ],
+    "candidates": [
+        "tezos-smart-rollup-client-Proxford",
+        "tezos-smart-rollup-client-PtNairob",
+        "tezos-smart-rollup-node-Proxford",
+        "tezos-smart-rollup-node-PtNairob",
+        "tezos-baking",
+        "tezos-sapling-params",
+        "tezos-accuser-Proxford",
+        "tezos-baker-Proxford",
+        "tezos-accuser-PtNairob",
+        "tezos-baker-PtNairob",
+        "tezos-node",
+        "tezos-dac-node",
+        "tezos-dac-client",
+        "tezos-codec",
+        "tezos-signer",
+        "tezos-admin-client",
+        "tezos-client"
+    ]
+}
--- a/docker/tests/scripts/test-fedora-binaries.sh
+++ b/docker/tests/scripts/test-fedora-binaries.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+set -euo pipefail
+
+for version in $(jq -r '.fedora[]' docker/supported_versions.json); do
+    docker build --build-arg dist="$version" --build-arg repo="Tezos" -t fedora-test -f docker/tests/Dockerfile-fedora-test .
+    docker run -rm fedora-test
+    docker build --build-arg dist="$version" --build-arg repo="Tezos-rc" -t fedora-test -f docker/tests/Dockerfile-fedora-test .
+    docker run -rm fedora-test
+done
--- a/docker/tests/scripts/test-ubuntu-binaries.sh
+++ b/docker/tests/scripts/test-ubuntu-binaries.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+set -euo pipefail
+
+for version in $(jq -r '.ubuntu[]' docker/supported_versions.json); do
+    docker build --build-arg dist="$version" --build-arg repo="tezos" -t ubuntu-test -f docker/tests/Dockerfile-ubuntu-test .
+    docker run -rm ubuntu-test
+    docker build --build-arg dist="$version" --build-arg repo="tezos-rc" -t ubuntu-test -f docker/tests/Dockerfile-ubuntu-test .
+    docker run -rm ubuntu-test
+done
--- a/docker/tests/test-fedora-binaries.py
+++ b/docker/tests/test-fedora-binaries.py
@@ -1,0 +1,41 @@
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+import subprocess
+import sys
+import json
+import os
+
+
+def test(binaries):
+    subprocess.run("dnf update -y", shell=True, capture_output=True)
+    for binary in binaries:
+        try:
+            subprocess.run(
+                f"dnf install -y {binary}", shell=True, capture_output=True, check=True
+            )
+            if binary != "tezos-sapling-params" and binary != "tezos-baking":
+                print(f"{binary}: ", end="", flush=True)
+                subprocess.check_call(
+                    f"{binary.replace('tezos', 'octez')} --version", shell=True
+                )
+        except Exception as e:
+            print(f"Exception happened when trying to execute tests for {binary}.\n")
+            raise e
+
+
+if __name__ == "__main__":
+    data = {}
+    with open("/tezos-packaging/binaries.json") as f:
+        data = json.load(f)
+
+    binaries = []
+    is_released = os.environ["IS_RELEASED"]
+    if is_released == "Tezos":
+        binaries = data["released"]
+    elif is_released == "Tezos-rc":
+        binaries = data["candidates"]
+    else:
+        raise "Incorrect argument"
+
+    test(binaries)
--- a/docker/tests/test-ubuntu-binaries.py
+++ b/docker/tests/test-ubuntu-binaries.py
@@ -1,0 +1,44 @@
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+import subprocess
+import sys
+import json
+import os
+
+
+def test(binaries):
+    subprocess.run("apt update -y", shell=True, capture_output=True)
+    for binary in binaries:
+        try:
+            subprocess.run(
+                f"apt install -y {binary.lower()}",
+                shell=True,
+                capture_output=True,
+                check=True,
+            )
+            if binary != "tezos-sapling-params" and binary != "tezos-baking":
+                print(f"{binary}: ", end="", flush=True)
+                subprocess.check_call(
+                    f"{binary.replace('tezos', 'octez')} --version", shell=True
+                )
+        except Exception as e:
+            print(f"Exception happened when trying to execute tests for {binary}.\n")
+            raise e
+
+
+if __name__ == "__main__":
+    data = {}
+    with open("/tezos-packaging/binaries.json") as f:
+        data = json.load(f)
+
+    binaries = []
+    is_released = os.environ["IS_RELEASED"]
+    if is_released == "tezos":
+        binaries = data["released"]
+    elif is_released == "tezos-rc":
+        binaries = data["candidates"]
+    else:
+        raise "Incorrect argument"
+
+    test(binaries)
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -4,6 +4,6 @@ Files: .github/* nix/nix/* nix/build/install_topfind_196.patch protocols.json me
 Copyright: 2022 Oxhead Alpha
 License: LicenseRef-MIT-OA
 
-Files: nix/build/install_topfind_196.patch baking/src/tezos_baking/__init__.py
+Files: nix/build/install_topfind_196.patch baking/src/tezos_baking/__init__.py docker/tests/binaries.json
 Copyright: 2023 Oxhead Alpha
 License: LicenseRef-MIT-OA


PruStephan:
Currently, we run sequential actions, testing binaries one by one. If we enable parallel actions, we suffer to consequences:
* The jobs start to fail with 137 status code, meaning exceeded memory (Currently tested on 3 jobs per os)
* Time between jobs can be significant, compared to the execution of one job, because they are scheduled with delay. 
Because of these two reasons, parallel execution does not provide much benefit, compared to sequential one

.github/workflows/sequential-test-fedora-binaries copy.yml
```diff
@@ -0,0 +1,147 @@
...
+
+      - name: Set up and test tezos-client
+        run: |
+          docker run --rm -i fedora:latest /bin/bash -c "
+            sudo dnf update -y
+            sudo dnf install -y 'dnf-command(copr)'
+            sudo dnf copr enable -y @Serokell/Tezos
+            sudo dnf install tezos-client -y
+            sudo tezos-client --version
+          "
```
krendelhoff2, line 26: AFAIR @pasqu4le wanted this to be able to run locally too. 
So I propose you to develop `Dockerfile` and a wrapper script to run all of this conveniently.
Please refer to the `docker/package/Dockerfile-ubuntu`, `docker/package/Dockerfile-fedora` and `docker/build/Dockerfile`. `docker/docker-static-build.sh` could be also interesting for you.

krendelhoff2: @pasqu4le also, do you think it's a good idea to use GA? AFAIU we use it only for bottle builds on `monterey`. Since all other builds run on `buildkite` and we use `docker`, I don't see any benefit from using GA here.

krendelhoff2 (changes requested):

.github/workflows/sequential-test-fedora-binaries copy.yml
pasqu4le: > AFAIR @pasqu4le wanted this to be able to run locally too.

Yes, indeed, and we need documentation on how to do so too.

Anyone should be able to follow the instruction and (after meeting the requirements) run these tests locally.

_BTW, that's true in general, about almost everything that runs in CI._

> @pasqu4le also, do you think it's a good idea to use GA? 

AFAIK the OPS team intends to stop using buildkite entirely at some point (cc @rvem, can you confirm?), so it's not a bad idea either.

Unless there's some reason to prefer buildkite now, I think that it's fine to implement new actions in GA (especially if they aren't connected to what we do in buildkite).

rvem: > AFAIK the OPS team intends to stop using buildkite entirely at some point (cc @rvem, can you confirm?)

Well, not really, public buildkite is free and we're rather considering not using it for new projects and maybe eventually migrating existing projects to it. Though, `tezos-packaging` has very complicated CI setup, so I'm not sure if it's actually worth completely switching it

.github/workflows/test-fedora-binaries.yml
```diff
@@ -0,0 +1,19 @@
+name: Test Fedora binaries (Sequential)
```
krendelhoff2, line 1: Do you think it's a good idea to split tests in different workflows? Semantically, fedora and ubuntu tests seem to be part of the one workflow.

docker/package/Dockerfile-fedora-test
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: 2021 Oxhead Alpha
```
krendelhoff2, line 1: Date have to be relevant for every new file
```suggestion
# SPDX-FileCopyrightText: 2023 Oxhead Alpha
```

docker/package/Dockerfile-ubuntu-test
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: 2021 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
```
krendelhoff2, line 3: Tests are not related to the packaging logic in any way, so they should be located in it's own directory. I propose `docker/tests` directory.

docker/package/Dockerfile-ubuntu-test
```diff
@@ -0,0 +1,12 @@
...
+FROM ubuntu:latest
```
krendelhoff2, line 4: It will not suffice for us, since we want to be sure that everything works fine on every distribution we support (see `docker/supported_versions.json`. You can provide image tag via build arguments.

docker/package/test-ubuntu-binaries.py
```diff
@@ -0,0 +1,78 @@
...
+]
+
+
+def setup():
+    commands = [
+        "apt update -y",
+        "apt install -y software-properties-common",
+        "apt update -y",
+        "add-apt-repository -y ppa:serokell/tezos",
+        "apt-get update -y",
```
krendelhoff2, line 30: I believe all this commands (except one `apt update -y`) should be moved to the build stage in the docker file

docker/package/test-ubuntu-binaries.py
```diff
@@ -0,0 +1,78 @@
...
+
+            processed_binaries.append(binary)
+
+            check_binary_command = f"sudo {binary} --version"
+            check_binary_result = subprocess.run(check_binary_command, shell=True)
+            assert check_binary_result.returncode == 0
+            sys.stdout.write(f"{binary} binary works fine\n")
+    except Exception as e:
+        print("Exception happened when trying to execute tests.\n")
+        print(f"{str(e)}")
```
krendelhoff2, line 65: Script does not signal any error if something happens. As you can see in your GA logs, exception happened, but job was marked as successful.

docker/package/test-ubuntu-binaries.py
```diff
@@ -0,0 +1,78 @@
...
+    sys.stdout.write("Setting up env\n")
+    setup()
+    sys.stdout.write("Setup is done\n")
+
+    processed_binaries = []
+    try:
+        for binary in binaries:
+            sys.stdout.write(
+                f"---------------------|Testing {binary}|---------------------\n"
+            )
```
krendelhoff2, line 47: This output is postponed until the end of the script (you have to manually flush buffer, I suppose). 
BTW, have you considered using something like `pytest`? I'm not insisting at all, it just could be easier in the end.

krendelhoff2 (changes requested):

docker/package/Dockerfile-ubuntu-test
PruStephan: Okay, but what version should we use on CI? I am asking, because if we set the version, then we would require to update it every time we move to the new version of os

docker/package/test-ubuntu-binaries.py
PruStephan: Currently, you can run:
* tests on your local machine
* tests on your local machine inside a docker container
* tests on CI inside a docker container

If we remove these updates from the Python script, we want to be able to use the first variant. Not sure, if we need that variant, but still.

docker/package/test-ubuntu-binaries.py
PruStephan: For such simple scripts, I don't think it is needed. For something more complex it might be relevant

docker/tests/Dockerfile-fedora-test
```diff
@@ -0,0 +1,16 @@
...
+ARG dist
+FROM fedora:${dist}
+
+WORKDIR /tezos-packaging/docker
+
+RUN dnf update -y
+RUN dnf install -y python3-devel python3-setuptools
+RUN sudo dnf update -y
+RUN sudo dnf install -y 'dnf-command(copr)'
+RUN sudo dnf copr enable -y @Serokell/Tezos
```
krendelhoff2, line 13:
```suggestion
RUN dnf update -y
RUN dnf install -y python3-devel python3-setuptools 'dnf-command(copr)'
RUN dnf copr enable -y @Serokell/Tezos
```

krendelhoff2, line 13: We should also check that packages from `Tezos-rc` repo get installed correctly, am I right? @pasqu4le 
```
ARG repo
RUN dnf copr enable -y @Serokell/${repo}
```

docker/tests/Dockerfile-ubuntu-test
```diff
@@ -0,0 +1,18 @@
...
+
+WORKDIR /tezos-packaging/docker
+
+RUN apt update -y
+RUN apt-get install -y python3-all python3-setuptools
+RUN apt update -y
+RUN apt install -y software-properties-common
+RUN apt update -y
+RUN add-apt-repository -y ppa:serokell/tezos
+RUN apt-get update -y
```
krendelhoff2, line 15:
```suggestion
RUN apt update -y
RUN apt install -y python3-all python3-setuptools software-properties-common
RUN add-apt-repository -yu ppa:serokell/tezos
```

docker/tests/test-fedora-binaries.py
```diff
@@ -0,0 +1,50 @@
...
+    "tezos-smart-rollup-client-PtNairob",
+    "tezos-smart-rollup-node-PtNairob",
+]
+
+
+def test(cleanup: bool):
+    processed_binaries = []
+    for binary in binaries:
+        try:
+            install_command = f"sudo dnf install -y {binary}"
```
krendelhoff2, line 28:
```suggestion
            install_command = f"dnf install -y {binary}"
```

docker/tests/test-fedora-binaries.py
```diff
@@ -0,0 +1,50 @@
...
+            installation_result = subprocess.run(install_command, shell=True)
+            assert installation_result.returncode == 0
+
+            processed_binaries.append(binary)
+
+            check_binary_command = f"sudo {binary} --version"
```
krendelhoff2, line 34:
```suggestion
            check_binary_command = f"{binary.replace('tezos', 'octez')} --version"
```

krendelhoff2: using `tezos-` prefixed binaries results in warning message, which pollutes the output (same on ubuntu)

docker/tests/test-fedora-binaries.py
```diff
@@ -0,0 +1,50 @@
...
+    "tezos-smart-rollup-client-PtMumbai",
+    "tezos-smart-rollup-node-PtMumbai",
+    "tezos-baker-PtNairob",
+    "tezos-accuser-PtNairob",
+    "tezos-smart-rollup-client-PtNairob",
+    "tezos-smart-rollup-node-PtNairob",
+]
+
+
+def test(cleanup: bool):
```
krendelhoff2, line 24: At this point we have to update package index again (on both ubuntu and fedora), because everything fetched by docker `RUN` commands will be cached and reused, so that we won't work on fresh index

docker/tests/test-ubuntu-binaries.py
```diff
@@ -0,0 +1,54 @@
...
+
+            check_binary_command = f"{binary} --version"
+            check_binary_result = subprocess.run(check_binary_command, shell=True)
+            assert check_binary_result.returncode == 0
+        except Exception as e:
+            print(f"Exception happened when trying to execute tests for {binary}.\n")
+            raise e
+    if cleanup:
+        for binary in processed_binaries:
+            remove_res = subprocess.run("apt remove {binary}")
```
krendelhoff2, line 46:
```suggestion
            remove_res = subprocess.run(f"apt remove {binary}")
```

docker/tests/test-ubuntu-binaries.py
```diff
@@ -0,0 +1,54 @@
...
+    processed_binaries = []
+    for binary in binaries:
+        try:
+            install_command = f"apt install -y {binary}"
+            installation_result = subprocess.run(install_command, shell=True)
+            assert installation_result.returncode == 0
+
+            processed_binaries.append(binary)
+
+            check_binary_command = f"{binary} --version"
```
krendelhoff2, line 38: It fails on `tezos-baker-ptmumbai`, it shouldn't be lowercased here

krendelhoff2 (changes requested):

krendelhoff2:
Also, @PruStephan, please do not resolve conversations, let the reviewer do it.

docker/tests/test-fedora-binaries.py
PruStephan: WDYM?

PruStephan:
> Also, @PruStephan, please do not resolve conversations, let the reviewer do it.

Yes, sure. I just needed a way to track resolved comments, since GitHub unlike Gitlab does not allow to toggle them, unfortunately.

docker/tests/Dockerfile-fedora-test
krendelhoff2: seems like you forgot this one

docker/tests/test-fedora-binaries.py
krendelhoff2: and this one too

krendelhoff2:
Everything LGTM, except one thing: the packages' versions get lost among the (mostly) irrelevant output. I propose to notify user that package is getting installed, but hide installation output using something like `subprocess.run` with the `capture_output` argument supplied. That way, you can show the installation output if some error occurs, but not show when everything is fine.

.github/workflows/test-binaries.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  test_binaries:
+    name: Install and test binaries
+    runs-on: [self-hosted, Linux, X64, nix-with-docker]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Test fedora binaries
+        run: ./docker/tests/scripts/test-fedora-binaries.sh
```
rvem, line 19: GA actually doesn't complete this job successfully, it says
```
./docker/tests/scripts/test-fedora-binaries.sh: line 8: jq: command not found
```
The fact that a non-zero exit code is swallowed and is not propagated is a problem

.github/workflows/test-binaries.yml
```diff
@@ -0,0 +1,23 @@
...
+
+      - name: Test ubuntu binaries
+        run: ./docker/tests/scripts/test-fedora-binaries.sh
```
rvem, line 22: I suppose
```suggestion
        run: ./docker/tests/scripts/test-ubuntu-binaries.sh
```

docker/tests/scripts/test-fedora-binaries.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+set -e
+
+for version in $(jq -r '.fedora[]' docker/supported_versions.json); do
+    docker build --build-arg dist="$version" -t fedora-test -f docker/tests/Dockerfile-fedora-test .
+    docker run -i fedora-test
```
rvem, line 10: Could you elaborate why `-i` is needed? AFAIR, it forwards stdin, but I assume that it's not needed here

docker/tests/test-fedora-binaries.py
```diff
@@ -0,0 +1,51 @@
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+import subprocess
+import sys
+
+binaries = [
```
rvem, line 7: I slightly dislike the fact that binaries are now listed in multiple places (here, in the ubuntu script and implicitly in [packages.py](https://github.com/serokell/tezos-packaging/blob/master/docker/package/packages.py)) and it's easy to forget to update one of them once the set of binaries changes

PruStephan: We can open a follow-up issue to hold JSON with all file names for example. I assume, that we require a full list of binaries in other places as well

rvem (commented):
Note that this PR should also cover repositories with RC packages, see its "Acceptance criteria"
> both for stable releases and RCs

As well as it should cover all supported Ubuntu and Fedora versions. Currently, tests are run only for a single version of each distro.


The overall approach LGTM

krendelhoff2 (changes requested):

docker/tests/test-fedora-binaries.py
```diff
@@ -0,0 +1,47 @@
...
+def test(binaries):
+    subprocess.run("sudo dnf update -y", shell=True, capture_output=True)
+    processed_binaries = []
+    for binary in binaries:
+        try:
+            install_command = f"sudo dnf install -y {binary}"
+            installation_result = subprocess.run(
+                install_command, shell=True, capture_output=True
+            )
+            assert installation_result.returncode == 0
```
krendelhoff2, line 19:
```suggestion
            installation_result = subprocess.run(
                install_command, shell=True, capture_output=True, check=True
            )
```

binaries.json
```diff
@@ -0,0 +1,36 @@
...
+        "tezos-smart-rollup-client-PtNairob",
+        "tezos-smart-rollup-node-PtNairob"
+    ],
+    "candidates": [
+        "tezos-smart-rollup-client-Proxford",
+        "tezos-smart-rollup-client-PtNairob",
+        "tezos-smart-rollup-node-Proxford",
+        "tezos-smart-rollup-node-PtNairob",
+        "tezos-baking",
+        "tezos-sapling-params",
```
krendelhoff2, line 23: They both should be present in `released` list too. And they both shouldn't be checked with `--version` command, only the fact that they installed successfully.

krendelhoff2 (changes requested):
All the threads apply the same way to `ubuntu` part too. Also, please rebase your PR on fresh master and squash old fixups.

binaries.json
```diff
@@ -0,0 +1,38 @@
+{
+    "released": [
```
krendelhoff2, line 2: New release just happened, so these lists are identical now. Also, please move it under the `docker/tests` dir for now.

krendelhoff2 (changes requested):
Besides that, everything LGTM.

docker/tests/scripts/test-ubuntu-binaries.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+set -e
```
krendelhoff2, line 6:
```suggestion
set -euo pipefail
```

docker/tests/scripts/test-fedora-binaries.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
```
krendelhoff2, line 5:
```suggestion

set -euo pipefail

```

docker/tests/scripts/test-ubuntu-binaries.sh
```diff
@@ -0,0 +1,13 @@
...
+
+for version in $(jq -r '.ubuntu[]' docker/supported_versions.json); do
+    docker build --build-arg dist="$version" --build-arg repo="tezos" -t ubuntu-test -f docker/tests/Dockerfile-ubuntu-test .
+    docker run fedora-test
```
krendelhoff2, line 10:
```suggestion
    docker run ubuntu-test
```

krendelhoff2 (changes requested):

docker/tests/scripts/test-fedora-binaries.sh
```diff
@@ -0,0 +1,13 @@
...
+set -euo pipefail
+
+for version in $(jq -r '.fedora[]' docker/supported_versions.json); do
+    docker build --build-arg dist="$version" --build-arg repo="Tezos" -t fedora-test -f docker/tests/Dockerfile-fedora-test .
+    docker run fedora-test
```
rvem, line 10:
```suggestion
    docker run --rm fedora-test
```
To cleanup the container after finishing

.github/workflows/test-binaries.yml
```diff
@@ -0,0 +1,23 @@
+name: Test Tezos binaries
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  push:
+    branches:
+      - PruStephan/Add_tests_for_packages
```
DMozhevitin, line 8: Just a reminder so you don't forget to change it

docker/tests/scripts/test-fedora-binaries.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+set -euo pipefail
+
+for version in $(jq -r '.fedora[]' docker/supported_versions.json); do
+    docker build --build-arg dist="$version" --build-arg repo="Tezos" -t fedora-test -f docker/tests/Dockerfile-fedora-test .
```
DMozhevitin, line 9: Could you please clarify why here `Tezos` arg is capitalized, but in Ubuntu test it isn't? 🤔

docker/tests/binaries.json
```diff
@@ -0,0 +1,40 @@
+{
+    "released": [
```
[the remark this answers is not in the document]
DMozhevitin: Btw I wonder if we can keep this file up-to-date automatically since there is a chance that we forget to update it

I don't insist on doing it in this PR, we always can open a follow-up one

docker/tests/binaries.json
```diff
@@ -0,0 +1,40 @@
...
+        "tezos-baker-PtNairob",
+        "tezos-node",
+        "tezos-dac-node",
+        "tezos-dac-client",
+        "tezos-codec",
+        "tezos-signer",
+        "tezos-admin-client",
+        "tezos-client"
+    ]
+}
```
DMozhevitin, line 40: nitpick: missing trailing newline in some of the files

DMozhevitin (changes requested):
Overall LGTM, left only minor comments

Also agree with the rest of open comments

docker/tests/scripts/test-fedora-binaries.sh
PruStephan: I was thinking about it, but since these scripts are only used in CI, the container will anyway be deleted afterwards

docker/tests/binaries.json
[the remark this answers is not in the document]
PruStephan: I think we can. Sounds like a simple script, and a small github action that updates list on release

.github/workflows/test-binaries.yml
PruStephan: Actually ...... thanks, this issue exists for so long, that I forgot, that I even inserted it there

docker/tests/scripts/test-fedora-binaries.sh
PruStephan: Because, unfortunately, on fedora and on ubuntu packages are named differently, see here https://serokell.slack.com/archives/C013LJZVC1K/p1690898759870919?thread_ts=1690819686.455979&cid=C013LJZVC1K

docker/tests/scripts/test-fedora-binaries.sh
rvem: > the container will anyway be deleted afterwards

No, this is not true, containers will only be deleted after daily `docker system prune` invocation on CI server

docker/tests/binaries.json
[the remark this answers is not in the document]
DMozhevitin: Sounds good

Could you please open a follow-up issue for this after merging this PR?

docker/tests/scripts/test-fedora-binaries.sh
DMozhevitin: Seems to be resolved now

binaries.json
DMozhevitin: Seems to be resolved now

DMozhevitin (approved):
LGTM now 👍

docker/tests/binaries.json
```diff
@@ -0,0 +1,40 @@
+{
+    "released": [
+        "tezos-smart-rollup-client-Proxford",
+        "tezos-smart-rollup-client-PtNairob",
+        "tezos-smart-rollup-node-Proxford",
+        "tezos-smart-rollup-node-PtNairob",
+        "tezos-baking",
+        "tezos-sapling-params",
+        "tezos-accuser-Proxford",
```
krendelhoff2, line 9: Add `tezos-smart-rollup-wasm-debugger` here and rebase please, after that I'll approve.

krendelhoff2 (changes requested):

krendelhoff2 (approved):
LGTM

PR #694, opened 2023-08-12
[#633] Allow users to specify snapshot provider

### LINKED ISSUE #633: Allow users to specify snapshot provider in setup wizard

The setup wizard currently has an option to use the metadata provided by xtz-shots (see #441) to download a snapshot for its node.

There are however other snapshot providers that are adopting the same metadata format (see [Agora thread](https://forum.tezosagora.org/t/we-re-standardizing-tezos-snapshot-metadata/5090/2)), so we want to let the users pick a different provider too.

A good solution might be to modify the existing option to ask the user for a URL to look for, but pre-fill this with the xtz-shots one. This would only add one more keypress for those that want to use the defaults.

# Acceptance criteria

Following the setup wizard, it's possible for users to specify the URL for the metadata of any snapshot provider.

### LINKED ISSUE #692: Do not hide download option silently in case of connectivity issues

In case of any connectivity issues querying `xtz-shots.io`,  snapshot download option is silently hided. 

Since network connectivity is essential to the whole setup, not alerting user of the issue here is not helpful. 
Also, first-time users running into this won't even know what they are missing and walk away with an impression that setup wizard does not provide all that much convenience.

# Acceptance criteria
1. Output the corresponding warning
2. Provide a way to recover from this

krendelhoff2:
## Description

Problem: The setup wizard currently has an option to use
the metadata provided by xtz-shots (see #441) to download
a snapshot for its node. There are however other snapshot
providers that are adopting the same metadata format, so we want
to let the users pick a different provider too.

Solution: Add the step to choose provider.

## Related issue(s)

Resolves #633
Resolves #692 

  can be successfully built.

### DIFF
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -222,7 +222,7 @@ liquidity_toggle_vote_query = Step(
 # We define this step as a function to better tailor snapshot options to the chosen history mode
 def get_snapshot_mode_query(modes):
     return Step(
-        id="snapshot",
+        id="snapshot_mode",
         prompt="The Tezos node can take a significant time to bootstrap from scratch.\n"
         "Bootstrapping from a snapshot is suggested instead.\n"
         "How would you like to proceed?",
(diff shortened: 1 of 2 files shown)


pasqu4le (approved):
LGTM now

pasqu4le (changes requested):
Sorry, slight change of plan (more info soon)

pasqu4le:
@krendelhoff2 after some discussion it was decided to go with an approach similar to `kiln`'s.

Instead of adding one more step, we should modify the `snapshot_import_modes` step like this:
1. keep the existing default `download rolling`/`full` option, which uses `xtz-shots.io`
2. add another `download rolling`/`full` below it that uses `marigold.dev`
3. modify the `url` option to be smarter, so that it also supports snapshot metadata

In particular, for (3), we should try to download and parse the metadata from the given url, in case of failure try the URL with `/tezos-snapshots.json` appended and finally in case of failure treat it as a snapshot URL (as we do now).

This makes the currently available choices easier, keeps the default path shorter for users and still adds the possibility to specify custom providers.

pasqu4le (approved):

PR #695, opened 2023-08-17
[#539] Load BigSur bottles from release instead of artifacts

### LINKED ISSUE #539: [BUG] CI doesn't update BigSur bottle hashes if a retry was needed
# Description

Currently CI is set up to (among other things) build BigSur bottles, both for arm64 and amd64, so that if an error occurs the build is retried.

Moreover, the retried build does not re-build packages that have been successfully built and uploaded on the previous try, to avoid unnecessary rebuilds.

Finally, CI also automatically opens a PR to update the hashes in brew formulae with the new bottles.

However, it seems that when a retry has occurred only hashes of the bottles built in that last attempt will be inserted.

# To Reproduce

It could be reproduced by purposefully failing to build some bottle on the first try.

For context, [here](https://buildkite.com/serokell/tezos-packaging-tags/builds/2473) is the pipeline in which this issue was noticed and [here](https://github.com/serokell/tezos-packaging/commit/9e6a1cbcef14c6add199c608444c5d525d7c89c9) is the incomplete commit that resulted from that.   

However

# Expected behavior

The automated PR to add bottle hashes includes them for all of the bottles built in that pipeline, not only those in the last attempt.

PruStephan:
## Description

Problem: when some bottles are build but then the step fails, the build restarts
without re-building the existing bottles.
This causes issues because we pick the bottles from the job's artifacts and we
only receive those from the second run of the job.

Solution: we take the BigSur bottles (after all have been built) from the GitHub
release where they have already been uploaded.

## Related issue(s)

Resolves #539 

  can be successfully built.

### DIFF
--- a/.buildkite/pipeline-for-tags.yml
+++ b/.buildkite/pipeline-for-tags.yml
@@ -152,15 +152,15 @@ steps:
      automatic:
        limit: 1
 
- # We use the tag that triggered the pipeline here. Normally, this isn't very resilient,
- # but in 'scripts/sync-bottle-hashes.sh' it's only used for informational purposes
+ # We use the tag that triggered the pipeline here.
+ # However, this requires that the tag and the release name are the same, which
+ # in practice it's always the case in this repo.
  - label: Add Big Sur bottle hashes to formulae
    depends_on:
    - "build-bottles-big-sur-arm64"
    - "build-bottles-big-sur-x86_64"
    if: build.tag =~ /^v.*/
-   soft_fail: true # No artifacts to download if all the bottles are already built
    commands:
    - mkdir -p "Big Sur"
-   - buildkite-agent artifact download "*bottle.tar.gz" "Big Sur/"
+   - nix develop .#buildkite -c gh release download "$BUILDKITE_TAG" -D "Big Sur/" -p "*big_sur.bottle.tar.gz"
    - nix develop .#autorelease -c ./scripts/sync-bottle-hashes.sh "$BUILDKITE_TAG" "Big Sur"


.buildkite/pipeline-for-tags.yml
```diff
@@ -162,5 +162,5 @@ steps:
    soft_fail: true # No artifacts to download if all the bottles are already built
    commands:
    - mkdir -p "Big Sur"
-   - buildkite-agent artifact download "*bottle.tar.gz" "Big Sur/"
+   - gh release download build.tag -D "Big Sur/" -p "*.bottle.tar.gz"
```
pasqu4le, line 165: It would be nice to have a comment here for future reference

.buildkite/pipeline-for-tags.yml
```diff
@@ -162,5 +162,5 @@ steps:
    soft_fail: true # No artifacts to download if all the bottles are already built
```
pasqu4le, line 162: The branch name should have the issue number in it, see [our guidelines](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e?pvs=4#c6e4c35d7aa74d1ca0b43ae049a3c847).

AFAIK you'd have to close this PR and open another one, so I think that for this small change it's not worth the trouble, but please name the branch correctly next time.

pasqu4le (commented):
> it is quite hard to reproduce it right now

Do you mean because of the timing of this PR?

The solution make sense to me, but I'd like to do a test run on CI and check that the problem is indeed resolved by it. However, let's do it after the release from #697 is complete, to not overload the infra with jobs.

.buildkite/pipeline-for-tags.yml
PruStephan: > It would be nice to have a comment here for future reference

Sure I will

> But I'd like to do a test run on CI

I will try to create a buildkite pipeline for this. GitHub actions fail to build macOS bottles

PruStephan: Also, in order to test this change, we need to simulate an update of the release. How would you perform it? I suppose, that we don't want to publish something that is not a correctly-building release. Maybe we can try to do that in another repo, WDYT?

pasqu4le: > I will try to create a buildkite pipeline for this

Makes sense, also because this issue appears on our buildkite's pipelines and not on GA

> we need to simulate an update of the release. How would you perform it? 

I think that's not necessary. We know that the bug appears when a build has failed at least once (the building script can be modified to do so) and that the issue is not with opening the PR, but rather that some bottles are missing when we do that.

So, IMO it's sufficient to make a temporary commit (in this branch or otherwise) to trigger the build (that should fail at least for one bottle on purpose) and list the bottles in the PR step (without then opening a new PR).

Let me know if you have questions about setting that up.

pasqu4le: @PruStephan any update on this test run?

PruStephan: I still did not get how you want this change to be tested. Current changes load hashes of bottles from the gh release, so if we're going to try it, we need, well, a new release, that will provide hashes, different from what we have. The only solution for this that I see is to manually change hashes of binaries, push the commit, and trigger loading from the latest release. What do you think about it? Also, do you think the complexity of this test is worth implementing it?

pasqu4le: Yes, a release needs to be created, but it can be a pre-release, which can then also be deleted.

The simplest way that came to my mind about testing this was by pushing [this commit](https://github.com/serokell/tezos-packaging/commit/5401bab153836f01ed0092949ace8ca24fd72436) in a temporary branch and then push a tag to trigger [its pipeline run](https://buildkite.com/serokell/tezos-packaging/builds/4148).

You can see the changes there, but in short I tried to make as little changes as possible to the scripts and tag pipeline definition, such that:
1. it creates a release, then builds and upload one bottle, then lists all the bottles, then deletes the release
2. it is triggered by each push of the `test-695` tag (I've already re-deleted it now)
3. it uses a temporary tag/release called `test-release`
4. both bottle build steps should fail on their first run (but upload the bottle) and not fail on the second run
5. the command that you've changed hasn't been modified, so that it is tested

While I'm writing this comment the pipeline is still running, so it's possible that I've made some mistakes, but if so it should still help you as an example of the kind of test that I had in mind.

pasqu4le: @PruStephan it looks like my test worked.

There are a bunch of failed builds for ARM BigSur, but those are for unrelated reasons (although they help simulate more problems).

As you can see, however, [the main job](https://buildkite.com/serokell/tezos-packaging-tags/builds/4612#018a7148-9326-43ff-89b6-8d2f234893ca) failed and that's because:
> line 2: gh: command not found

I'd imagine that this is because `gh` is only available on our build machine behind a `nix` command

PruStephan: Ok, I've got it. I will update the command in the test branch, and then trigger the build once more

pasqu4le:
The previous solution (8bc960cc3378b7a17933dc6ce6abf981f9d97178) doesn't really work because it picks all the bottles available from the release.

This is a problem if the bottle builds from GitHub Actions have been pushed already, as this would result in the step modified here to grab those too (and consequently push the wrong bottle hashes update).
See [this test pipeline](https://buildkite.com/serokell/tezos-packaging-tags/builds/4744#018acc75-1499-4f6d-88c8-7280bef8aac5) for example.

The current solution solves this problem by filtering for BigSur bottles, see [this other test pipeline](https://buildkite.com/serokell/tezos-packaging-tags/builds/4745#018accdc-0542-4512-9b88-1a8545bdaf45).

P.s. I also removed the `soft_fail: true` from the step because (see deleted comment) there's no longer a case in which we expect there to be no bottles available to download.

pasqu4le (approved):

Issue #707, opened 2023-09-04
Select snapshots from the corresponding Octez release


pasqu4le:
# Clarification and motivation

The setup wizard can download snapshots from a provider with metadata, but it doesn't filter these by Octez version.

This can cause issues in case a snapshot has been created with an Octez version that is incompatible with the one being used.

# Acceptance criteria

The setup wizard only downloads snapshots created with the corresponding Octez version.

itkach:
Snapshot format and compatibility is not tied directly to octez version, attempting to select snapshot strictly by octez version would be a mistake. Snapshot format itself is versioned separately (available in `snapshot_version` in snapshot metadata). Even when format changes newer octez it typically able to read previous snapshot format.

pasqu4le:
> Even when format changes newer octez it typically able to read previous snapshot format.

I agree, although typically isn't always and AFAIK there is no forward compatibility guarantees on snapshot versions.

Moreover, AFAICT there is no command to ask the node what snapshot versions it supports, so this info might need to be hardcoded.
With the current solution (see #708 ) we can ask the node its version and check against that, which doesn't couple the wizard version with the `octez-node` version.

As (obviously, hence this issue) backward compatibility also isn't to be expected, filtering for the exact `octez-node` version seems the only safe option.

_Note: #708 was merged already and releases were cut with it, this issue should have been closed already so I'm doing so now. However, feel free to re-open it (or a new one) if the current solution is not satisfying_

pasqu4le:
I should probably add for completeness that even _the exact_ `snapshot_version` is currently not working as a filtering method on its own.

AFAICT from xtz-shots' metadata, both Octez node version `v17.3` and `v18.0-rc1` produce snapshots with `"snapshot_version": 5`, but those produced by the latter are rejected by the former (which is what was reported to us and prompted the opening of this issue).

PR #708, opened 2023-09-04
[#707] Filter snapshots by node's commit hash

krendelhoff2:
## Description

Problem: The setup wizard can download snapshots from a provider with metadata, but it doesn't filter these by Octez version. This can cause issues in case a snapshot has been created with an Octez version that is incompatible with the one being used.

Solution: Filter by node's commit hash.

## Related issue(s)

Resolves #

  can be successfully built.


baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -387,7 +393,9 @@ def get_snapshot_link(self):
                             self.config["history_mode"] == "archive"
                             and artifact["history_mode"] == "full"
                         )
-                    ),
+                    )
+                    and artifact["tezos_version"]["commit_info"]["commit_hash"]
+                    == node_version_hash,
```
pasqu4le, line 398: I suppose that it's unlikely, but perhaps we should instead check that one hash is prefix of the other (or the other way around), in case only one of them uses the full hash.

pasqu4le (commented):
Overall LGTM, just one comment

pasqu4le (commented):
LGTM except for one thing: the version in `baking/pyproject.toml` also needs to be updated (twice)

pasqu4le (approved):

PR #710, opened 2023-09-07
[Chore] Utilize OCTEZ_EXECUTABLES env var

krendelhoff2:
## Description

Problem: `make static` command utilizes `OCTEZ_EXECUTABLES` variable,
which we can use to specify exact binaries to build. Also, it
could be used to build experimental or development binaries.

Solution: Let user provide `OCTEZ_EXECUTABLES` env var and pass
it to the build.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -19,7 +19,7 @@ Static binaries building using custom alpine image.
 image defined in [Dockerfile](build/Dockerfile). In order to build them you should specify
 `OCTEZ_VERSION` env variable and run the script:
 ```
-export OCTEZ_VERSION="v14.1"
+export OCTEZ_VERSION="v17.3"
 ./docker-static-build.sh
 ```
 After that, directory will contain built static binaries.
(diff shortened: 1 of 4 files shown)


docker/README.md
```diff
@@ -19,7 +19,7 @@ Static binaries building using custom alpine image.
 image defined in [Dockerfile](build/Dockerfile). In order to build them you should specify
 `OCTEZ_VERSION` env variable and run the script:
 ```
-export OCTEZ_VERSION="v14.1"
+export OCTEZ_VERSION="v17.3"
```
pasqu4le, line 22: These are meant to be example commands and we don't update them for every Octez update, so I think they are fine either way.

TBH I find this commit unnecessary, but I don't mind it either, so feel free to keep it

pasqu4le (approved):
Seems to be working fine :+1:

Issue #568, opened 2022-11-16
Implement non-interactive/partial interactive mode for tezos-setup-wizard


itkach:
# Clarification and motivation

It would facilitate scripting and copy-paste distribution of setup instructions if tezos-setup-wizard had a non-interactive mode

# Acceptance criteria
- Non interactive mode: take user input from the command line, fail with a instructive error message in case of errors or incomplete input
- Partial interactive mode: take user input from the command line, drop into interactive mode on errors or in case of incomplete input (skipping wizard steps where input was given on the command line)

PR #712, opened 2023-09-13
[Chore] Factor out code from `wizard_structure`

krendelhoff2:
## Description

This is preparatory PR for the #568.

## Related issue(s)

Resolves #

  can be successfully built.


pasqu4le (approved):
These all look like improvements to be :+1: 

One minor nitpick, all 3 commit messages start with:
 > Problem: wizard_structure contains too *much* definitions

but AFAIK "many" should be used here instead of "much".

PR #720, opened 2023-09-22
[#718] Improve algorithm of picking snapshot

### LINKED ISSUE #718: Improve the algorithm of picking the latest compatible snapshot

Currently setup wizard tries to download the snapshot only on the same Octez version (see #707) from the provider metadata. However, this algorithm can be improved. The proposed solution has the following steps:

1. Try to find the snapshot of exact same Octez version that is used in `tezos-packaging`
2. If there is none, try to find the snapshot with **the same major version**, but **less minor version** and with the `snapshot_version` compatible with the current Octez (it's `6` for `v18.0`). For the time being, until [tezos/#6319](https://gitlab.com/tezos/tezos/-/issues/6319) is resolved we can just hardcode this value.
3. If there is none, try to find the snapshot with any Octez version, but compatible `snapshot_version`.

# Acceptance criteria

Setup wizard uses the specified algorithm of picking the latest compatible snapshot from the provider metadata

krendelhoff2:
Problem: Currently setup wizard tries to download the snapshot only on the same Octez version (see #707) from the provider metadata. However, this is not correct, since snapshot doesn't have to be created with the exact same node version.

Solution: Provide one of these latest snapshots in decreasing order of relevance:
* exact version match
* major version is the same, but lesser minor version
* snapshot version is 6

## Related issue(s)

Resolves #718 

  can be successfully built.

### DIFF
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -602,14 +602,14 @@ class Setup(Setup):
         except Exception as e:
             print(f"\nUnexpected error handling snapshot metadata:\n{e}\n")
 
-    def output_snapshot_metadata(self):
+    def output_snapshot_metadata(self, name):
         from datetime import datetime
         from locale import setlocale, getlocale, LC_TIME
 
         # it is portable `C` locale by default
         setlocale(LC_TIME, getlocale())
 
-        metadata = self.config["snapshot_metadata"]
+        metadata = self.config["snapshots"][name]
         timestamp_dt = datetime.strptime(
             metadata["block_timestamp"], "%Y-%m-%dT%H:%M:%SZ"
         )
@@ -640,6 +640,7 @@ block timestamp: {timestamp} ({time_ago})
         try:
             url = self.config["snapshots"][name]["url"]
             sha256 = self.config["snapshots"][name]["sha256"]
+            self.output_snapshot_metadata(name)
             return fetch_snapshot(url, sha256)
         except KeyError:
             raise InterruptStep
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -179,8 +179,20 @@ def is_full_snapshot(snapshot_file, import_mode):
     return False
 
 
-def get_node_version_hash():
-    return get_proc_output("octez-node --version").stdout.decode("ascii").split()[0]
+def get_node_version():
+    version = get_proc_output("octez-node --version").stdout.decode("ascii")
+    major_version, minor_version, rc_version = re.search(
+        r"[a-z0-9]+ \(.*\) \(([0-9]+).([0-9]+)(?:(?:~rc([1-9]+))|(?:\+dev))?\)",
+        version,
+    ).groups()
+    return (
+        int(major_version),
+        int(minor_version),
+        (int(rc_version) if rc_version is not None else None),
+    )
+
+
+compatible_snapshot_version = 6
 
 
 # Steps
@@ -406,21 +418,18 @@ class Setup(Setup):
             return False
         return True
 
-    # Check the provider url and collect the most recent snapshot
-    # that is suited for the chosen history mode and network
-    def get_snapshot_metadata(self, name, json_url):
-        def hashes_comply(s1, s2):
-            return s1.startswith(s2) or s2.startswith(s1)
+    # Returns relevant snapshot's metadata
+    # It filters out provided snapshots by `network` and `history_mode`
+    # provided by the user and then follows this steps:
+    # * tries to find the snapshot of exact same Octez version, that is used by the user.
+    # * if there is none, try to find the snapshot with the same major version, but less minor version
+    #   and with the `snapshot_version` compatible with the user's Octez version.
+    # * If there is none, try to find the snapshot with any Octez version, but compatible `snapshot_version`.
+    def extract_relevant_snapshot(self, snapshot_array):
+        from functools import reduce
 
-        try:
-            snapshot_array = None
-            with urllib.request.urlopen(json_url) as url:
-                snapshot_array = json.load(url)["data"]
-            snapshot_array.sort(reverse=True, key=lambda x: x["block_height"])
-
-            node_version_hash = get_node_version_hash()
-
-            snapshot_metadata = next(
+        def find_snapshot(pred):
+            return next(
                 filter(
                     lambda artifact: artifact["artifact_type"] == "tezos-snapshot"
                     and artifact["chain_name"] == self.config["network"]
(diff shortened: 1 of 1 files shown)


baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -182,6 +182,23 @@ def get_node_version_hash():
...
+    major_version, minor_version = (
+        re.search(
+            r"\([0-9]+\.[0-9]+\)",
+            version,
+        )
+        .group()
+        .strip("()")
+        .split(".")
+    )
+    return int(major_version), int(minor_version)
```
DMozhevitin, line 196: `minor_version` seems to be unused below since we still use the old approach with hashes

I'd suggest removing the old logic and using the `minor_version` too for consistency

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -430,16 +435,69 @@ def hashes_comply(s1, s2):
...
+        exact_version_pred = (
+            lambda artifact, node_version_hash=node_version_hash: hashes_comply(
+                artifact["tezos_version"]["commit_info"]["commit_hash"],
+                node_version_hash,
             )
+        )
+
+        major_version, _ = get_node_version()
+
+        major_version_pred = (
```
DMozhevitin, line 458: This predicate doesn't check 2 things:

1. That `minor_version` is less than the current version. i.e. the wizard can pick the snapshot generated by `17.3` `octez-node` when user only have `17.2` installed, and there is a possibility that they are incompatible

2. it doesn't check that `snapshot_version` is equal to the `compatible_snapshot_version`: it's still a possible case when accepted `snapshot_version` is different between the binaries with the same major version

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -182,6 +182,23 @@ def get_node_version_hash():
     return get_proc_output("octez-node --version").stdout.decode("ascii").split()[0]
 
 
+def get_node_version():
+    version = get_proc_output("octez-node --version").stdout.decode("ascii")
+    major_version, minor_version = (
+        re.search(
+            r"\([0-9]+\.[0-9]+\)",
```
DMozhevitin, line 189: It will fail if `rc` binary is used, e.g.:

```
174938a1 (2023-05-10 08:58:52 +0000) (17.0~rc1)
```

DMozhevitin (changes requested):

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -430,16 +432,131 @@ def hashes_comply(s1, s2):
...
+            lambda artifact: minor_version == get_artifact_version(artifact)[1]
+        )
+
+        exact_rc_version_pred = (
+            lambda artifact: rc_version == get_artifact_version(artifact)[2]
+        )
+
+        less_minor_version_pred = (
+            lambda artifact: minor_version > get_artifact_version(artifact)[1]
+        )
```
pasqu4le, line 483: All these predicates are using (part of) the artifact's version, couldn't they take that as an argument instead of the artifact itself?

IOW can't `get_artifact_version` be called only once?

On a related note, I find this name quite confusing, `artifact_version` makes me think that this has the version of the artifact, but really this is the version of the node that was used to make the snapshot. I recommend renaming these.

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -405,21 +417,11 @@ def check_blockchain_data(self):
...
-            return s1.startswith(s2) or s2.startswith(s1)
-
-        try:
-            snapshot_array = None
-            with urllib.request.urlopen(json_url) as url:
-                snapshot_array = json.load(url)["data"]
-            snapshot_array.sort(reverse=True, key=lambda x: x["block_height"])
-
-            node_version_hash = get_node_version_hash()
+    def extract_snapshot_metadata(self, snapshot_array):
```
pasqu4le, line 420: This definitely could use a comment.

The name suggests that this extracts metadata from a snapshot, but without reading its implementation one cannot tell that this does also the filtering and picking of the best snapshot available.

Also, the strategy used to pick the snapshot is not written down anywhere either, but it should be.

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -430,16 +432,131 @@ def hashes_comply(s1, s2):
...
+            with urllib.request.urlopen(json_url) as url:
+                snapshot_array = json.load(url)["data"]
+            snapshot_array.sort(reverse=True, key=lambda x: x["block_height"])
+
+            snapshot_metadata = self.extract_snapshot_metadata(snapshot_array)
+
+            if snapshot_metadata is None:
+                print(
+                    color(
+                        f"Suitable snapshot is not found in {name} provider.",
```
pasqu4le, line 553:
```suggestion
                        f"No suitable snapshot found from the {name} provider.",
```

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -405,21 +417,11 @@ def check_blockchain_data(self):
...
-            snapshot_array.sort(reverse=True, key=lambda x: x["block_height"])
-
-            node_version_hash = get_node_version_hash()
+    def extract_snapshot_metadata(self, snapshot_array):
+        from functools import reduce
 
-            snapshot_metadata = next(
+        def find_snapshot(pred):
+            return next(
                 filter(
```
pasqu4le, line 425: I suppose that there should be one major difference between the algorithm used here and the one used in Kiln.

Kiln is always releases with a stable Octez version and so we decided to filter out all the snapshots from Octez RCs.

Our packages however also get released for RCs, so we shouldn't do the same here.

IMO here we should filter out all snapshots from RCs only if we are using a stable Octez release and not otherwise.

krendelhoff2: Why do you think we should filter out RC snapshots? When stable version just gets released, v18.0 for example, shouldn't latest snapshot for v18.0-rc1 be fine for it? (since there is no snapshots for v18.0 yet). But then I think we should also output node version information and snapshot version, so that user won't be very surprised if for some reason such import would fail.

baking/src/tezos_baking/tezos_setup_wizard.py
krendelhoff2: > All these predicates are using (part of) the artifact's version, couldn't they take that as an argument instead of the artifact itself?

They could, but actually then `get_artifact_version` would become unnecessary at all. That also means that the responsibility of that function will be delegated to the predicates itself.

baking/src/tezos_baking/tezos_setup_wizard.py
pasqu4le: > Why do you think we should filter out RC snapshots?

Because, it being an RC, there's more chances that something is wrong with it.

Let's say that now we have a `v18.0` release and that a new `v18.1-rc1` comes out with the same bug as [tezos/#6318](https://gitlab.com/tezos/tezos/-/issues/6318) (which is what `v18.0-rc1` had), then with this new algorithm we'll have another broken release.

Small bugs like this are much more likely to happen with RCs and then get fixed before the next stable release.

> When stable version just gets released, v18.0 for example, shouldn't latest snapshot for v18.0-rc1 be fine for it?

Not necessarily, for example let's say that `v18.0` stopped supporting snapshot version `5` (which could have happened, since its snapshot version is `6`), then snapshots from `v18.0-rc1` would be incompatible (due to the same bug linked above).

Likewise in this case, and in general, we shouldn't make as many assumptions on RCs as we would for stable releases, because the former are intended to be tested more and (where necessary) fixed.

baking/src/tezos_baking/tezos_setup_wizard.py
pasqu4le: > I suppose you mean that we can call it inside find_snapshot and then call predicates with individual version parts

Not quite, I mean that we can call it inside `find_snapshot` and then call predicates with _all_ the version parts.

IOW I'd like to give in input to the predicates the entire output of  `get_artifact_version`

This should make it possible to also compose these predicates with one-another, no?

krendelhoff2: Makes sense, this change is included in recent fixup

baking/src/tezos_baking/tezos_setup_wizard.py
DMozhevitin: As for handling RCs, I'd suggest doing the following (on the 2nd step):

* Reject all `rc` snapshots if user has a binary of stable version
* If user has an `rc` binary, then accept either
    a) snapshots with the same `major_version`, same `minor_version` and less `rc_version` (please note that we should reject stable snapshot with same major and minor version there too)
    b) **non-rc** snapshots with the same `major_version` and less `minor_version`
    
As for the `b)`, I think that there is no reason to care about `18.0-rc1` snapshots if we have `18.1-rc1` node because there definitely will be more fresh `18.0` snapshot. If there won't be such a snapshot (which is kinda unlikely), it's better to move to the next step anyway, since the current logic of handling the RCs is too complicated and error-prone (see my other comment)
    
@krendelhoff2 what do you think?

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -430,16 +439,150 @@ def hashes_comply(s1, s2):
...
+            # e.g. marigold snapshots don't supply it
+            lambda major, minor, rc, snapshot_version: snapshot_version
+            and compatible_snapshot_version == snapshot_version
+        )
+
+        no_rc_on_stable_pred = lambda major, minor, rc, snapshot_version: not (
+            rc and rc_version is None
+        )
+
+        every_pred = lambda major, minor, rc, snapshot_version: True
```
DMozhevitin, line 525: Do we really need this predicate? :thinking: 

IMO it's better to return `None` and say that we didn't find a compatible snapshot rather than picking the one which is mist likely incompatible and fail during the import

Or am I missing something and it won't work like this?

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -430,16 +439,150 @@ def hashes_comply(s1, s2):
...
+
+        less_minor_version_pred = (
+            lambda major, minor, rc, snapshot_version: minor_version > minor
+        )
+
+        exact_rc_version_pred = (
+            lambda major, minor, rc, snapshot_version: rc_version == rc
+        )
+
+        less_rc_version_pred = (
```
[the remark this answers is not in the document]
krendelhoff2: Well, it makes sense to add another check just to be less error prone, but how does this situation is possible now? Shouldn't such case short-curcuit on `non_rc_on_stable_version` predicate?

DMozhevitin (approved):

PR #730, opened 2023-10-02
[#728] Expand tilde char in user provided path

### LINKED ISSUE #728: [BUG] ~ in snapshot path does not expand
# Description

When specifying snapshot file path, tab completion correctly interprets `~` as an alias for user's home directory, however path starting with `~` then produces an error 

# To Reproduce

- Place a snapshot into `~/Downloads`  (e.g. download using browser)
- run tezos-setup
- choose to create node from a snapshot
- choose to provide snapshot file
- enter path to the snapshot file using `~`, e.g. `~/Downloads/ghostnets-123456.rolling`
- observe tab completion works as expected
- observe entered correct path results in error

# Expected behavior

no error, `~` is expanded to `/home/$USER/` as usual and snapshot file is successfully located

krendelhoff2:
## Description

Problem: User provided paths with tilde char does not
pass the validation check.

Solution: Expand tilde char to the home directory.

## Related issue(s)

Resolves #728 

  can be successfully built.

### DIFF
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -12,6 +12,7 @@ import os, sys, shutil
 import readline
 import re
 import traceback
+import time
 import urllib.request
 import json
 from typing import List
@@ -574,6 +575,8 @@ block timestamp: {timestamp} ({time_ago})
             for name, url in default_providers.items():
                 self.get_snapshot_metadata(name, url)
 
+            os.makedirs(TMP_SNAPSHOT_LOCATION, exist_ok=True)
+
         else:
             return
 
@@ -589,7 +592,11 @@ block timestamp: {timestamp} ({time_ago})
                     return
                 elif self.config["snapshot_mode"] == "file":
                     self.query_step(snapshot_file_query)
-                    snapshot_file = self.config["snapshot_file"]
+                    snapshot_file = os.path.join(
+                        TMP_SNAPSHOT_LOCATION, f"file-{time.time()}.snapshot"
+                    )
+                    # not copying since it can take a lot of time
+                    os.link(self.config["snapshot_file"], snapshot_file)
                 elif self.config["snapshot_mode"] == "direct url":
                     self.query_step(snapshot_url_query)
                     url = self.config["snapshot_url"]
--- a/baking/src/tezos_baking/validators.py
+++ b/baking/src/tezos_baking/validators.py
@@ -38,9 +38,9 @@ def dirpath(input):
 
 
 def filepath(input):
-    if input and not os.path.isfile(input):
+    if input and not os.path.isfile(os.path.expanduser(input)):
         raise ValueError("Please input a valid file path.")
-    return input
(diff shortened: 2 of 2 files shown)


baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -589,7 +591,9 @@ def import_snapshot(self):
                     return
                 elif self.config["snapshot_mode"] == "file":
                     self.query_step(snapshot_file_query)
-                    snapshot_file = self.config["snapshot_file"]
+                    snapshot_file = os.path.join(TMP_SNAPSHOT_LOCATION, "file.snapshot")
+                    # not copying since it can take a lot of time
+                    os.link(self.config["snapshot_file"], snapshot_file)
```
DMozhevitin, line 596: Should we also remove this symlink after all? Consider the following steps:

1. User imported the snapshot from a file (in my case, I cancelled the import, but I believe it doesn't matter)
2. User runs `tezos-setup` once again and imports the same file

After it user will get the following error

```
  File "src/tezos_baking/tezos_setup_wizard.py", line 839, in main
    setup.run_setup()
  File "src/tezos_baking/tezos_setup_wizard.py", line 783, in run_setup
    self.bootstrap_node()
  File "src/tezos_baking/tezos_setup_wizard.py", line 654, in bootstrap_node
    self.import_snapshot()
  File "src/tezos_baking/tezos_setup_wizard.py", line 596, in import_snapshot
    os.link(self.config["snapshot_file"], snapshot_file)
FileExistsError: [Errno 17] File exists: '/home/dmitriy/serokell/node-snapshots/tezos-mainnet.rolling' -> '/tmp/octez_node.snapshot.d/file.snapshot'
```

krendelhoff2: It does matter, since `/tmp/octez_node.snapshot.d/` gets cleaned up after import, but it is still a fair point, it shouldn't fail here

krendelhoff2:
> @krendelhoff2 could you please check if you can reproduce it on your machine?

It does not reproduce. Please check that you run relevant version of `tezos-setup` and that this file actually exists.
```
Provide the path to the node snapshot file.
> ~/serokell/node-snapshots/nairobinet-1665178.rolling
Oct 03 10:15:45.960: Note: the import of a snapshot can be sped up using the '--no-check' option.
Oct 03 10:15:45.960:   Only use this option if you fully trust the snapshot source.
```

DMozhevitin (approved):

DMozhevitin:
> > @krendelhoff2 could you please check if you can reproduce it on your machine?
> 
> It does not reproduce. Please check that you run relevant version of `tezos-setup` and that this file actually exists.
> 
> ```
> Provide the path to the node snapshot file.
> > ~/serokell/node-snapshots/nairobinet-1665178.rolling
> Oct 03 10:15:45.960: Note: the import of a snapshot can be sped up using the '--no-check' option.
> Oct 03 10:15:45.960:   Only use this option if you fully trust the snapshot source.
> ```

It looks like I indeed ran a wrong version. Works fine now :+1:

Issue #734, opened 2023-10-12
tezos-client cannot be installed on a container managed by Kubernetes


NeoDev42:
# Clarification and motivation

We tried to install the `tezos-client` package in the following way from a container orchestrated by Kubernetes and it does not work:

```
$ sudo apt-get install -y tezos-client
Setting up tezos-client (2:18.0-0ubuntu3~focal) ...
Failed to send reload request: No such file or directory
dpkg: error processing package tezos-client (--configure):
 installed tezos-client package post-installation script subprocess returned error exit status 1
```
I tried to run the post-installation script directly by this way to check which line failed:

```
sh -x /var/lib/dpkg/info/tezos-client.postinst
```

And I got the same error, but in the script we have a very useful comment:

```
# Don't add udev rules in case the package is installed inside either docker or podman container.
# Otherwise, post-installation script will fail due to inability to non-zero exit code of the
# 'udevadm control --reload-rules' call
if [ ! -f /.dockerenv ] && [ ! -f /.containerenv ] ; then
    cat <<EOF > /etc/udev/rules.d/20-hw1.rules 
```

Nevertheless this test does not handle all cases, indeed in a container orchestrated by Kubernetes we have neither `.dockerenv` nor `.containerenv`.

By creating a fake `.dockerenv` , it works.

We must find a variable or the presence of an additional file to test if we are in a kubernetes pod.

# Acceptance criteria

Command `apt-get install -y tezos-client` MUST work from a Kubernetes pod.

krendelhoff2:
@NeoDev42 we made a [new release](https://github.com/serokell/tezos-packaging/releases/tag/v18.0-4) containing fix, please run `apt-get update` and try to install one more time

DMozhevitin:
Closing the issue for now
@NeoDev42 feel free to reopen it if this problem is relevant for the new release

PR #737, opened 2023-10-17
[#734] Do not add udev rules in k8s environment

krendelhoff2:
## Description

Problem: Post-install script adding udev rules will fail running in container environement and it has checks for it. It doesn't cover all cases though. Specifically, it does not recognize running in k8s environment. According to the
[docs](https://kubernetes.io/docs/tutorials/services/connect-applications-service/#environment-variables) k8s always export `KUBERNETES_SERVICE_HOST` variable.

Solution: Check if `KUBERNETES_SERVICE_HOST` variable is set.

## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/docker/package/scripts/udev-rules
+++ b/docker/package/scripts/udev-rules
@@ -8,8 +8,10 @@
 
 # Don't add udev rules in case the package is installed inside either docker or podman container.
 # Otherwise, post-installation script will fail due to inability to non-zero exit code of the
-# 'udevadm control --reload-rules' call
-if [ ! -f /.dockerenv ] && [ ! -f /.containerenv ] ; then
+# 'udevadm control --reload-rules' call.
+# Since containers orchestrated by Kubernetes don't contain any indicator files, check k8s-specific
+# env var.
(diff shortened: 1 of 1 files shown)


docker/package/scripts/udev-rules
```diff
@@ -9,7 +9,7 @@
 # Don't add udev rules in case the package is installed inside either docker or podman container.
 # Otherwise, post-installation script will fail due to inability to non-zero exit code of the
 # 'udevadm control --reload-rules' call
-if [ ! -f /.dockerenv ] && [ ! -f /.containerenv ] ; then
+if [ ! -f /.dockerenv ] && [ ! -f /.containerenv ] && [ -z "${KUBERNETES_SERVICE_HOST++}" ] ; then
```
DMozhevitin, line 12: It seems that we should also update the comment above and mention k8s

krendelhoff2: Doesn't it still fits the description? AFAIU k8s uses docker

DMozhevitin: Yes, but IMO it won't hurt much to mention why we still check for `KUBERNETES_SERVICE_HOST` env variable (since there is neither `.dockerenv` nor `.containerenv`) just not to forget about it over time

DMozhevitin (approved):

Issue #518, opened 2022-08-12
Package smart contract rollup Octez binaries


pasqu4le:
# Clarification and motivation

The [v14.0 Octez release](https://gitlab.com/tezos/tezos/-/releases/v14.0) introduced new rollup binaries for SCORUs (containing `sc` in their name), see [changelog](http://tezos.gitlab.io/CHANGES.html?highlight=scorus#rollups).

These are different from the ones covered by #453 but should also be supported.

Note: these are more recent, not as documented, and possibly not as stable compared to transactions rollups. As such it may be a good idea to wait some more time before supporting them.

# Acceptance criteria

1) We provide static binaries, Ubuntu/Fedora packages, and brew formulae for new smart contract rollup Octez binaries.
2) We provide systemd/launchd services for binaries that are supposed to be run as a background daemon.

pasqu4le:
update: the latest Octez RC ([v16.0-rc1](https://gitlab.com/tezos/tezos/-/releases/v16.0-rc1) has these binaries too and the functionality will be active on the upcoming `mumbainet` testnet as [per announcement](https://research-development.nomadic-labs.com/mumbai-announcement.html#smart-rollups-and-epoxy-testers-wanted).

 _note: their name is now `octez-smart-rollup-<node|client>-<proto>` now_

These new binaries should be stable enough and documented sufficiently to allow us to package them and create systemd services too.

pasqu4le:
@PruStephan I'm assigning you this issue now, when we already have _some_ support for rollup binaries.

Please start by checking:
1. what docs need to be updated, I see that we still have an outdated `docs/tx-rollup.md`
2. if there is more functionality that we can add, especially to `systemd` services

Feel free to ask questions to @krendelhoff2 as he worked on the current docs and implementation.

PR #740, opened 2023-10-20
[#518] Update rollup binaries

PruStephan:
## Description
This PR updates the usage of smart-rollup binaries so that instead of transaction rollup binaries we use smart rollup binaries

## Related issue(s)

Resolves #

  can be successfully built.


krendelhoff2:
There is also mentions `TYPE` variable also, see [here](https://github.com/serokell/tezos-packaging/blob/master/docker/package/packages.py#L516)

docs/smart-rollup.md
```diff
@@ -3,20 +3,20 @@
    - SPDX-License-Identifier: LicenseRef-MIT-OA
    -->
 
-# Setting up transaction rollup node on Ubuntu
+# Setting up smart rollup node on Ubuntu
 
 At first you should originate rollup with the provided alias for some implicit account:
 ```
-sudo -u tezos tezos-client originate tx rollup <ROLLUP_ALIAS> from <IMPLICIT_ACCOUNT_ALIAS>
+sudo -u tezos tezos-client originate smart rollup <ROLLUP_ALIAS> from <IMPLICIT_ACCOUNT_ALIAS>
```
krendelhoff2, line 10: This command doesn't seem to be valid anymore
```
Error:
  Unterminated command, here are possible completions.
    originate smart rollup <alias> from <src> of kind <smart_rollup_kind> of type <parameters_type> with kernel <kernel> [--fee <amount>]
```

krendelhoff2 (changes requested):

docs/smart-rollup.md
```diff
@@ -7,7 +7,7 @@
 
 At first you should originate rollup with the provided alias for some implicit account:
 ```
-sudo -u tezos tezos-client originate smart rollup <ROLLUP_ALIAS> from <IMPLICIT_ACCOUNT_ALIAS>
+sudo -u tezos tezos-client originate smart rollup <ROLLUP_ALIAS> from <IMPLICIT_ACCOUNT_ALIAS> of kind <SMART_ROLLUP_KIND> of type <ROLLUP_PARAMETER_TYPE> with kernel <KERNEL>
```
krendelhoff2, line 10: `ROLLUP_ALIAS` and `IMPLICIT_ACCOUNT_ALIAS` are more or less self-explanatory, but others should be documented with some example of use provided

krendelhoff2 (changes requested):

docs/smart-rollup.md
PruStephan: To be honest,  finding any examples on how exactly this binary should be used, is a total mess. I found a sort of documentation, which I provided link to

docker/package/packages.py
```diff
@@ -506,78 +506,74 @@ def mk_node_unit(
...
-                exec_start_pre=[
-                    "+/usr/bin/setfacl -m u:tezos:rwx /run/systemd/ask-password"
-                ],
-                exec_start=startup_script,
-                exec_stop_post=[
-                    "+/usr/bin/setfacl -x u:tezos /run/systemd/ask-password"
-                ],
-                state_directory="tezos",
-                user="tezos",
-                type_="forking",
```
DMozhevitin, line 526: Could you please elaborate a bit on this `type_="forking"` -> `type_="simple"` change? I didn't get it, to be honest 😅

DMozhevitin (approved):
LGTM aside from one point that I've asked to clarify 👍

docker/package/packages.py
DMozhevitin: oh, I guess `forking` is about having different services like we do for `tezos-node`, but since we have single smart rollup node binary, it isn't needed now 🤔

krendelhoff2: `octez-smart-rollup-node` does not seem to fork. `forking` systemd service consider service activated when it forks into its daemon. It is the case for `octez-node`. But `octez-smart-rollup-node` just runs and service hangs into `activating` state, failing in the end because of the timeout.

krendelhoff2: `forking` systemd service is about having `fork()` system call during service startup

DMozhevitin: Got it, thanks for the clarification

krendelhoff2 (approved):

PR #741, opened 2023-10-24
[Chore] Fix bug in tests

### LINKED ISSUE #641: Test packages installation for linux distros

PruStephan:
## Description
Fix small bug in binaries testing script

## Related issue(s)

Resolves #641

  can be successfully built.

### DIFF
--- a/docker/tests/scripts/test-fedora-binaries.sh
+++ b/docker/tests/scripts/test-fedora-binaries.sh
@@ -7,7 +7,7 @@ set -euo pipefail
 
 for version in $(jq -r '.fedora[]' docker/supported_versions.json); do
     docker build --build-arg dist="$version" --build-arg repo="Tezos" -t fedora-test -f docker/tests/Dockerfile-fedora-test .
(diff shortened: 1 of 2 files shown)


krendelhoff2 (approved):
Approved, please don't forget to change `branches` clause before merging

DMozhevitin:
Also could you please mention #641 in the PR and change the PR/commit name to be a bit more informative, e.g. smth like "Fix docker argument in binaries tests"?

PR #743, opened 2023-10-25
[#729] Introduce better prompts for `tezos-vote`

### LINKED ISSUE #729: [BUG] tezos-vote lacks proper prompts
# Description

tezos-vote does not provide adequate user prompts and feedback

# To Reproduce

- run `tezos-vote`
- `Does this look correct? (Y/n) y `
- enter selection for yay/nay/pass
- observe Ledger Nano S screen prompts to sign request
- observe there is no prompt from `tezos-vote` instructing user to check their ledger screen
- accept transaction on ledger device
- observe nothing at all happens in the terminal; tezos-vote produces no output and hangs indefinitely;
- observe the vote actually goes through despite having no indication of success from tezos-vote

# Expected behavior

- user is prompted to look at their ledger device
- tezos-vote prints informational messages to inform user of the progress and exits upon completion

krendelhoff2:
## Description

Problem: Currently, `tezos-vote` does not provide
adequate user prompts and feedback when used with
ledger devices.

Solution: Provide visible prompts requesting action from user.

## Related issue(s)

Resolves #729

  can be successfully built.

### DIFF
--- a/baking/src/tezos_baking/tezos_voting_wizard.py
+++ b/baking/src/tezos_baking/tezos_voting_wizard.py
@@ -513,6 +513,7 @@ class Setup(Setup):
 
         print()
         print("Thank you for voting!")
+        logging.info("Exiting the Tezos Voting Wizard.")
 
 
 def main():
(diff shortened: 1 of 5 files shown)


DMozhevitin (approved):
LGTM :+1: 

I have only minor suggestion. While we're on it, let's update the 1st step of `docs/voting-test.md` to mention voting using the ledger device. Something like this:

1. Generate a pair of keys associated with baker alias:

```
sudo -u tezos tezos-client gen keys baker
```

If you want to use your ledger device for voting, you should import its encrypted private key instead:

First, run this command:

```
sudo -u tezos tezos-client list connected ledgers
```

This will display some instructions to import the Ledger encrypted private key. Then run

```
sudo -u tezos tezos-client import secret key baker ledger://XXXXXXXXXX
```

And confirm providing the public key using the prompt on the ledger device.

PR #746, opened 2023-11-08
[#738] Further improve the algorithm of picking the latest compatible snapshot

### LINKED ISSUE #738: Further improve the algorithm of picking the compatible snapshot from provider

In #718 we did some improvements in the algorithm of picking the latest compatible snapshot from provider in `tezos-setup`. However, there are still some improvements that would be nice to implement:

1. Currently, `tezos-setup` scans both known providers (`xtz-shots` and `marigold`) for the presence of compatible snapshot before showing the snapshot import options. But in the cases when user wants to use other option (e.g. enter the direct link or specify snapshot file path), this search become redundant, so we need to run it on demand.
2. When the selected provider doesn't have compatible snapshot in its metadata, we need to use the second one as a fallback by default, and report the error only when the search failed on both of them.
3. Currently on the 2nd and 3rd steps of the selection algorithm we filter the snapshots with the `snapshot_version` equal to the `compatible_snapshot_version` value. However, this condition is too strict, and we can also accept the snapshots with `snapshot_version == compatible_snapshot_version - 1`.

# Acceptance criteria

All of the improvements described above are implemented in the `tezos-setup`.

DMozhevitin:
## Description

This PR makes several improvements in the logic of picking the latest compatible snapshot in provider's metadata. See the commit messages to get more details about each improvement.

## Related issue(s)

Resolves #738

  can be successfully built.

### DIFF
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -543,7 +543,7 @@ class Setup(Setup):
             # it could happen that `snapshot_version` field is not supplied by provider
             # e.g. marigold snapshots don't supply it
             lambda major, minor, rc, snapshot_version: snapshot_version
-            and compatible_snapshot_version == snapshot_version
+            and compatible_snapshot_version - snapshot_version <= 1
         )
 
         non_rc_on_stable_pred = lambda major, minor, rc, snapshot_version: (
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -196,6 +196,21 @@ def get_node_version():
     )
 
 
+def is_non_protocol_testnet(network):
+    return network == "mainnet" or network == "ghostnet"
+
+
+# Starting from Nairobi protocol, the corresponding testnet
+# is no longer a named network, so we need to provide the URL
+# of the network configuration instead of the network name
+# in 'octez-node config init' command.
+def network_name_or_teztnets_url(network):
+    if is_non_protocol_testnet(network):
+        return network
+    else:
+        return f"https://teztnets.xyz/{network}"
+
+
 compatible_snapshot_version = 6
 
 
@@ -396,12 +411,13 @@ class Setup(Setup):
         if not node_dir_config.issubset(node_dir_contents):
             print_and_log("The Tezos node data directory has not been configured yet.")
             print_and_log("  Configuring directory: " + node_dir)
+            network = self.config["network"]
             proc_call(
                 "sudo -u tezos octez-node-"
                 + self.config["network"]
                 + " config init"
                 + " --network "
-                + self.config["network"]
+                + network_name_or_teztnets_url(self.config["network"])
                 + " --rpc-addr "
                 + self.config["node_rpc_addr"]
             )
--- a/baking/src/tezos_baking/tezos_setup_wizard.py
+++ b/baking/src/tezos_baking/tezos_setup_wizard.py
@@ -262,8 +262,7 @@ def get_snapshot_mode_query(config):
     dynamic_import_modes = {}
 
     for name in default_providers.keys():
-        if config["snapshots"].get(name, None):
-            dynamic_import_modes[mk_option(name)] = mk_desc(name)
+        dynamic_import_modes[mk_option(name)] = mk_desc(name)
 
     import_modes = {**dynamic_import_modes, **static_import_modes}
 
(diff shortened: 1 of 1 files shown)


baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -694,6 +693,48 @@ def get_snapshot_from_provider(self, name, url):
...
+    # provider's metadata
+    #
+    # if the snapshot not found, tries to find it in other known
+    # providers
+    def get_snapshot_from_provider_with_fallback(self, name, url):
+        print_and_log(f"Getting snapshots' metadata from {name}...")
+
+        self.get_snapshot_metadata(name, url)
+        snapshot = self.config["snapshots"].get(name, None)
+
```
krendelhoff2, line 733:
```suggestion
fallback_provider = None
```
otherwise it fails with `UnboundLocalError: local variable 'fallback_provider' referenced before assignment`

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -793,10 +830,22 @@ def import_snapshot(self):
...
+                        message = f"Couldn't find available snapshot in any of the known providers."
+                        print(
+                            color(
+                                message,
+                                color_red,
+                            )
+                        )
+                        logging.warning(message)
+                        print_and_log("Getting back to the snapshot import mode step.")
+                        continue
```
krendelhoff2, line 856:
```suggestion
                        raise InterruptStep
```

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -793,10 +830,22 @@ def import_snapshot(self):
...
+                    (snapshot_file, snapshot_block_hash) = res
```
krendelhoff2, line 857:
```suggestion
                    (snapshot_file, snapshot_block_hash) = snapshot_info
```

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -793,10 +830,22 @@ def import_snapshot(self):
...
                     for name, url in default_providers.items():
                         if name in self.config["snapshot_mode"]:
-                            (
-                                snapshot_file,
-                                snapshot_block_hash,
-                            ) = self.get_snapshot_from_provider(name, url)
+                            selected_provider = name
+                    snapshot_info = self.get_snapshot_from_provider_with_fallback(
+                        selected_provider, url
+                    )
```
krendelhoff2, line 845:
```suggestion 
                            selected_provider = (name, url)
                    snapshot_info = self.get_snapshot_from_provider_with_fallback(*selected_provider)
```

krendelhoff2 (changes requested):
Choosing either provider option results into `Couldn't find available snapshot in any of the known providers.`
BTW, maybe it makes sense to ask user about using fallback provider at first before trying it

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -694,6 +693,48 @@ def get_snapshot_from_provider(self, name, url):
...
+        return self.config["snapshots"].get(name, None)
+
+    # check if some of the providers has the compatible snapshot
+    # available in its metadadata and return the provider name
+    # along with snapshot itself if found
+    def find_fallback_provider(self, providers):
+        for name, url in providers.items():
+            snapshot = self.try_fallback_provider(name, url)
+            if snapshot is not None:
+                return (snapshot, name)
```
krendelhoff2, line 727: Return of `snapshot` seem to be redundant here

baking/src/tezos_baking/tezos_setup_wizard.py
krendelhoff2:
```suggestion
                    if snapshot_info is None:
                            print_and_log(
                                 "Couldn't find available snapshot in any of the known providers.",
                                 log=logging.warning,
                                 colorcode=color_yellow,
                            )
                            raise InterruptStep
```

krendelhoff2 (changes requested):

krendelhoff2 (approved):

baking/src/tezos_baking/tezos_setup_wizard.py
```diff
@@ -687,6 +686,49 @@ def get_snapshot_from_provider(self, name, url):
         snapshot_block_hash = self.config["snapshots"][name]["block_hash"]
         return (snapshot_file, snapshot_block_hash)
 
+    # check if a given provider has the compatible snapshot
+    # available in its metadata and return the metadata of this
+    # snapshot if it's available
+    def try_fallback_provider(self, name, url):
+        print(f"Getting snapshots' metdata from {name} instead...")
```
krendelhoff2, line 709:
```suggestion
        print(f"Getting snapshots' metadata from {name} instead...")
```

krendelhoff2 (changes requested):

krendelhoff2 (approved):

PR #747, opened 2023-11-14
[#705] Better show the error logs of the wizards

### LINKED ISSUE #705: Better show the error logs of the setup wizard

The setup wizard can fail for a number of reasons, after which it will mention that logs have been dumped in a file.

However:
1. the location of this file is unclear from the message given by the wizard (it's always in the working dir)
2. the wizard doesn't attempt to show the logs or suggest how to do it

Both of these points could be improved.

_Note: this might be better handled with #651 in mind_ 

# Acceptance criteria

The setup wizard suggests how to open the logs in an unequivocal manner

krendelhoff2:
## Description

Problem: the location of the log file is unclear from the message given by the wizards. They also don't attempt to show the logs or suggest how to do it.

Solution: Show last 10 lines of the error output. Suggest how to see the full log. Separate log entries in the file by the dates.

## Related issue(s)

Resolves #705 

  can be successfully built.

### DIFF
--- a/baking/src/tezos_baking/tezos_voting_wizard.py
+++ b/baking/src/tezos_baking/tezos_voting_wizard.py
@@ -536,13 +536,13 @@ def main():
         logging.info("Exiting the Tezos Voting Wizard.")
         sys.exit(1)
     except Exception as e:
-        print("Error in Tezos Voting Wizard, exiting.")
-        logging.error(f"{str(e)}")
+        message = "Error in the Tezos Voting Wizard, exiting."
+        print("\n" + color(message, color_red))
+        logging.error(message)
+
+        log_exception(exception=e, logfile="tezos-vote.log")
+
         logging.info("Exiting the Tezos Voting Wizard.")
-        logfile = "tezos_vote.log"
-        with open(logfile, "a") as f:
-            f.write(str(e) + "\n")
-        print("The error has been logged to", os.path.abspath(logfile))
         sys.exit(1)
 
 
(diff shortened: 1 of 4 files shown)


baking/src/tezos_baking/wizard_structure.py
```diff
@@ -107,8 +107,8 @@ def setup_logger(log_file):
     )
 
 
-def print_and_log(s, log=logging.info):
-    print(s)
+def print_and_log(message, log=logging.info, colorcode=None):
+    print(color(message, colorcode) if colorcode else message)
     log(s)
```
DMozhevitin, line 112:
```suggestion
    log(message)
```

baking/src/tezos_baking/tezos_voting_wizard.py
```diff
@@ -536,13 +536,13 @@ def main():
         logging.info("Exiting the Tezos Voting Wizard.")
         sys.exit(1)
     except Exception as e:
-        print("Error in Tezos Voting Wizard, exiting.")
-        logging.error(f"{str(e)}")
+        message = "Error in the Tezos Setup Wizard, exiting."
```
DMozhevitin:
```suggestion
message = "Error in the Tezos Voting Wizard, exiting."
```

DMozhevitin (changes requested):
LGTM overall, have only a couple of minor nitpicks

Btw, is it possible to save the logfile to some predefined dir (e.g. `/var/log`)? It seems a bit more proper to me, but maybe there are some drawbacks (e.g. with permissions) that I'm missing :thinking:

DMozhevitin (approved):
LGTM

PR #750, opened 2023-11-21
[Chore] v18.1 release hotfixes

krendelhoff2:
## Related issue(s)

Resolves #

  can be successfully built.


DMozhevitin (approved):
LGTM

While we're here, could you please change [this](https://github.com/serokell/tezos-packaging/blob/master/baking/src/tezos_baking/tezos_setup_wizard.py#L562) condition from `<= 1` to `<= 2` as we discussed?

Judging by the [changelog](https://tezos.gitlab.io/CHANGES.html), snapshots with version 5 will be also supported by `v18.1` `octez-node`, so we can safely pick them in our algroithm

PR #751, opened 2023-11-22
[#736] Add setup to automatically update binaries

### LINKED ISSUE #736: Automatically update binaries list on release
Currently, we have to update the binaries.json list manually when we add new binaries or change protocols. We can do this automatically, via script, for example, one part will copy release candidates to released property on release, and the second one will scan the Formula folder and copy all names from it to candidates. 

# Acceptance criteria
- [ ] Create a script that updates `binaries.json` on release.
- [ ] Test it with fake release

PruStephan:
## Description
This PR adds a pipeline setup to update the list of binaries after every successful tag build. The script will create a followup PR with an updated list of binaries. 

## Related issue(s)

Resolves #736 

  can be successfully built.

### DIFF
--- a/docker/package/meta.py
+++ b/docker/package/meta.py
@@ -25,16 +25,16 @@ class PackagesMeta:
         self.license_version = license_version
 
 
-tag = os.environ["OCTEZ_VERSION"]
-for (i, c) in enumerate(tag):
+meta_json_contents = json.load(
+    open(f"{os.path.dirname(__file__)}/../../meta.json", "r")
+)
+
+tag = os.environ.get("OCTEZ_VERSION", meta_json_contents["tezos_ref"])
+for i, c in enumerate(tag):
     if c.isdigit():
         digit_index = i
         break
 version = tag[digit_index:]
-
-meta_json_contents = json.load(
-    open(f"{os.path.dirname(__file__)}/../../meta.json", "r")
-)
 packages_meta = PackagesMeta(
     tag=tag,
     version=version,
--- a/docker/package/update-test-binaries-list.py
+++ b/docker/package/update-test-binaries-list.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2023 Oxhead Alpha
+# SPDX-License-Identifier: LicenseRef-MIT-OA
+
+import os
+import json
+from .packages import packages as all_packages
+from .meta import packages_meta
+
+binaries_json_path_suffix = "tests/binaries.json"
+
+
+def update_binaries(binaries, field):
+    binaries_json_path = os.path.join(os.environ["PWD"], binaries_json_path_suffix)
+    with open(binaries_json_path, "r") as file:
+        data = json.load(file)
+
+    data[field] = binaries
+    with open(binaries_json_path, "w") as file:
+        json.dump(data, file, indent=4)
+        file.write("\n")
+
+
+def main():
+    tag = packages_meta.tag
+    binaries = []
+    with open(f"{os.path.dirname(__file__)}/../octez-executables", "r") as f:
+        binaries = [l.strip() for l in f.readlines()]
+    if not binaries:
+        raise Exception(
+            "Exception, while reading binaries list: binaries list is empty"
+        )
+
+    field = "candidates" if "rc" in tag else "released"
+
+    update_binaries(binaries, field)
+
+
+if __name__ == "__main__":
+    main()
--- a/docker/tests/binaries.json
+++ b/docker/tests/binaries.json
@@ -1,20 +1,19 @@
 {
     "released": [
-        "tezos-smart-rollup-node",
-        "tezos-smart-rollup-wasm-debugger"
-        "tezos-baking",
-        "tezos-sapling-params",
-        "tezos-accuser-Proxford",
-        "tezos-baker-Proxford",
-        "tezos-accuser-PtNairob",
-        "tezos-baker-PtNairob",
-        "tezos-node",
-        "tezos-dac-node",
-        "tezos-dac-client",
-        "tezos-codec",
-        "tezos-signer",
-        "tezos-admin-client",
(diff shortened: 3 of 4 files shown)


docker/package/scripts/update-binaries-list.py
```diff
@@ -0,0 +1,44 @@
...
+import os
+import sys
+import json
+
+binaries_json_path_suffix = "/tests/binaries.json"
+binaries_list_path_suffix = "/tmp/binaries.txt"
+
+
+def update_binaries(binaries, field):
+    binaries_json_path = os.environ["PWD"] + binaries_json_path_suffix
```
DMozhevitin, line 13: nitpick: using `os.path.join` will be a bit more proper there (and in other places)

docker/package/scripts/update-binaries-list.py
```diff
@@ -0,0 +1,44 @@
...
+        binaries = [l.strip() for l in f.readlines()]
+
+    if not binaries:
+        raise Exception(
+            "Exception, while reading binaries list: binaries list is empty"
+        )
+
+    field = "released"
+    if "rc" in tag:
+        field = "candidates"
```
DMozhevitin, line 37: Nitpick:

```suggestion
    field = "candidates" if "rc" in tag else "released"
```

docker/update-binaries.sh
```diff
@@ -0,0 +1,20 @@
...
+
+git config --local user.email "hi@serokell.io"
+git config --local user.name "$bot_name"
+git fetch --all
+git checkout -B "$branch"
+
+python3 package/scripts/update-binaries-list.py
+
+git add --all
+git commit -m "Updated binaries for $BUILDKITE_TAG release" --gpg-sign="tezos-packaging@serokell.io"
```
DMozhevitin, line 17: For some reason [this commit](https://github.com/serokell/tezos-packaging/pull/752/commits/5d02cf1840a47af892a73d795e9b970dbd4a5367) is marked as `Unverified` which means that it's unsigned 🤔

docker/package/scripts/update-binaries-list.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def update_binaries(binaries, field):
+    binaries_json_path = os.environ["PWD"] + binaries_json_path_suffix
+    with open(binaries_json_path, "r") as file:
+        data = json.load(file)
+
+    data[field] = binaries
+    with open(binaries_json_path, "w") as file:
+        json.dump(data, file, indent=4)
```
DMozhevitin, line 19: After automatic update `binaries.json` doesn't contain trailing newline (see [there](https://github.com/serokell/tezos-packaging/pull/752/files#diff-43f2e418820e6fc4daf593ef39c4036b8312781e426e5ad90d3f797d25c95bb1R38))

docker/package/scripts/update-binaries-list.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def main():
+    tag = os.environ["BUILDKITE_TAG"]
+    binaries = []
+
+    binaries_list_path = os.environ["PWD"] + binaries_list_path_suffix
+    with open(binaries_list_path, "r") as f:
+        binaries = [l.strip() for l in f.readlines()]
```
DMozhevitin, line 28: It seems that we should do `s/octez/tezos/` for binary names as well.

The reason is that tests in `docker/tests-test-*-binaries.py` install **packages** using `apt` and `rpm`, and the package names are `tezos-*`, not `octez-*`, so that `apt install octez-client` will fail, and the right command is `apt install tezos-client`.

For more context about binaries/packages naming policy see [this thread](https://serokell.slack.com/archives/C013LJZVC1K/p1665650088834379?thread_ts=1665502772.172429&cid=C013LJZVC1K)

docker/update-binaries.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
DMozhevitin, line 1: I'd suggest renaming this file to `update-binaries-list.sh`

DMozhevitin:
There is `.github/binaries_update_pr.md` PR template, but it seems that it's not used anywhere. At least #752 doesn't use this template.

Btw, I'd suggest renaming the template to `binaries_list_update_pr.md` to be consistent in naming.

DMozhevitin (changes requested):
Mostly LGTM 👍 , have only minor comments

docker/docker-static-build.sh
```diff
@@ -52,4 +52,5 @@ container_id="$("$virtualisation_engine" create alpine-tezos)"
 for b in "${binaries[@]}"; do
     "$virtualisation_engine" cp "$container_id:/tezos/$b" "$b"
 done
+printf "%s\n" "${binaries[@]}" > "/tmp/binaries.txt"
```
krendelhoff2, line 55: It is not the correct place to get installable packages list. You can always get them by importing our python script, thus this does not require any additional file at all. See [here](https://github.com/serokell/tezos-packaging/blob/master/docker/build/ubuntu/build.py#L18).

docker/update-binaries.sh
```diff
@@ -0,0 +1,20 @@
...
+git config --local user.name "$bot_name"
+git fetch --all
+git checkout -B "$branch"
+
+python3 package/scripts/update-binaries-list.py
+
+git add --all
+git commit -m "Updated binaries for $BUILDKITE_TAG release" --gpg-sign="tezos-packaging@serokell.io"
+git push --set-upstream origin "$branch"
+gh pr create -B master -t "Update list of binaries for $BUILDKITE_TAG" -b "Updated list of binaries for $BUILDKITE_TAG version"
```
krendelhoff2, line 19: We also have to ensure that this PR won't happen if binaries list hasn't changed

krendelhoff2 (changes requested):

krendelhoff2 (approved):

DMozhevitin (approved):

PR #760, opened 2023-12-03
[OPS-1436] Use rootless docker

Sereja313:
## Description
Problem: We introduced rootless docker on albali and now we need to change our docker CI pipelines to use albali runners instead of bunda runners.

Solution: Remove the bunda specific runner tag.

## Related issue(s)

Resolves #

  can be successfully built.


DMozhevitin (approved):
LGTM, but I'd ask for one more review from a person from ops just in case

rvem (approved):

PR #766, opened 2023-12-09
[#704] Add script to automatically fix node corruption

### LINKED ISSUE #704: Fix node storage corruption automatically

`octez-node` can sometime exit with leaving a corrupted state on disk, which happens more frequently if it's interrupted abruptly.

The node can sometimes fix the state by itself, but it can also fail to do so, preventing it from starting.

This affects also nodes launched with our `systemd` services, which can fail to start for this reason.

It should be possible to automatically detect the failure of the node to fix its corrupted state and consequently perform a cleanup of the node storage and a re-sync of it.

# Acceptance criteria

When our `tezos-node` service is started with a corrupted state that `octez-node` cannot fix by itself, then the service fixes it instead.

PruStephan:
## Description
Add script that performs the following: 
* Cheks if node logs contain record of corruption
* Removes the node dir
* Loads relevant snapshot to temporary dir
* imports it

## Related issue(s)

Resolves #704 

  can be successfully built.


baking/src/tezos_baking/restore_from_corruption.py
```diff
@@ -0,0 +1,104 @@
...
+    if not reinstallation_result.returncode:
+        print("Recovery from corruption was successfull")
+    else:
+        print("Recovery from corruption failed. Manual restoration is required")
+
+
+def main():
+    is_corrupted = check_node_corruption()
+    is_baking_installed = (
+        b"tezos-baking" in get_proc_output("which octez-baking").stdout
```
DMozhevitin, line 73: 1. There is no `octez-baking` package, but `tezos-baking`.`octez-*` is about binaries from [tezos/tezos](https://gitlab.com/tezos/tezos) that we build and provide via `tezos-*` packages (i.e. installing `tezos-node` **package** via `sudo apt-get install tezos-node` installs `octez-node` **binary**). But apart from that, `tezos-baking` is a special case, since there is no such binary as `octez-baking`, it's rather set of `systemd` services and other things that are used together with `octez` binaries.

2. `which` command searches the path of executable. Considering words above, it can't be used for checking the presence of `tezos-baking` package in the system, since this package isn't an executable.

So we need to use another approach to checking whether `tezos-baking` is installed. The first thing that comes to mind is `apt list --installed | grep tezos-baking`, but I'm pretty sure that there are better alternatives

DMozhevitin (changes requested):
The restoration logic looks ok overall, so it seems that you're moving in the right direction, although I have some comments

Issue #715, opened 2023-09-20
Stop bottle support for BigSur and add it for the latest MacOS versions


pasqu4le:
# Clarification and motivation

Currently we build bottles for BigSur and Monterey for our releases.

At the time of writing, there is already a new MacOS version (Ventura) and one that should shortly be announced (Sonoma).

Once the announcement of the latter happened, we should re-honor [our support policy](https://github.com/serokell/tezos-packaging/blob/master/docs/support-policy.md#macos-brew-formulae) and drop/add support accordingly for the currently maintained MacOS versions.

# Acceptance criteria

Bottles are build for all currently maintained MacOS version (recommended: after the announcement of Sonoma).

PR #767, opened 2023-12-20
[#715] Stop bottle support for Big Sur

krendelhoff2:
## Description

Problem: Big Sur is deprecated by Apple.

Solution: Stop bottle support for Big Sur.

## Related issue(s)

Resolves #

  can be successfully built.


rvem:
Note that GA is also building some Monterey bottles at the moment, so it should either be removed or updated to use `macos-13`

DMozhevitin (changes requested):
It seems that there is still `arm64_big_sur` leftover in [scripts/update-brew-formulae.sh#22]. Should be `arm64_monterey` now, I suppose

Issue #755, opened 2023-11-27
[BUG] tezos-setup tries to register key as delegate even if it's already registered


DMozhevitin:
# Description

When `baker` key is already registered as delegate, `tezos-setup` still shows the ledger prompt and calls `octez-client register key baker as delegate` command which fails with error. We need to skip these actions if the account has been already registered as delegate. It can be checked by querying `/context/delegates/<pkh>` node RPC endpoint.

# To Reproduce

1. Manually register the account used for baking in `tezos-setup` as delegate
2. Select this account in `tezos-setup`
3. Observe that `tezos-setup` shows the ledger prompt and calls `register key <key> as delegate` command which fails with error. The setup wizard won't fail with error though.

# Expected behavior

If the account is already a delegate, `tezos-setup` skip registering step.

# Environment

- Ubuntu 20.04
- `v18.1-1` release

krendelhoff2:
fixed by #768

PR #768, opened 2023-12-20
[#755] Check if bakey key already registered

krendelhoff2:
## Description

Problem: Sometimes at point of `register_baker` action key is already registered as delegate, resulting in error output.

Solution: Check if it already registered.

## Related issue(s)

Resolves #

  can be successfully built.


DMozhevitin (approved):
LGTM :+1: , have only one minor nitpick:

When account is already registered a delegate, the wizard still shows "Registering the baker" text:

```
...
Registering the baker
Would you like to request to end the Liquidity Baking subsidy?
...
```

I believe that it's easy to fix, so feel free to merge this PR after resolving this comment.

PR #769, opened 2023-12-21
[Chore] Update "actions/checkout"

Sereja313:
Problem: node16 is now deprecated and github-runner provided by nixpkgs no longer supports this runtime. However, "actions/checkout@v3" uses this runtime.

Solution: Update CI pipeline to use "actions/checkout@v4".

## Related issue(s)

Resolves #

  can be successfully built.


Sereja313:
some of them use our runners, which causes scheduled jobs to fail

rvem (approved):

PR #771, opened 2023-12-27
[#736] Use binaries list from octez and partially automate binaries update

krendelhoff2:
## Related issue(s)

Resolves #

  can be successfully built.

### DIFF
--- a/scripts/shell.nix
+++ b/scripts/shell.nix
@@ -8,6 +8,7 @@ mkShell {
     # NOTE using the most recent nix for sticky nix flake lock update feature
     nix
     python3
+    moreutils
     coreutils
     gnused
     gh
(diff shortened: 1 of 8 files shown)


scripts/shell.nix
```diff
@@ -8,6 +8,7 @@ mkShell {
     # NOTE using the most recent nix for sticky nix flake lock update feature
     nix
     python3
+    moreutils
```
DMozhevitin, line 11: Do we actually use the tools from `moreutils` in this PR? 🤔

DMozhevitin (approved):
Looks good 👍 , have only one question

Btw, is it possible to make a dummy tag like [this one](https://github.com/serokell/tezos-packaging/releases/tag/test-736) to check how it works in action?

scripts/shell.nix
krendelhoff2: yes, `proto` script requires `sponge` utility from `moreutils`

DMozhevitin: Got it 👌

PR #783, opened 2024-03-13
[#782] Support tzinit provider

### LINKED ISSUE #782: Replace xtz-shots snapshot provider with tzinit

`xtz-shots` snapshot provider is down for some time, and it's unlikely that it will continue working. At the same time, the new snapshot provider [tzinit](https://snapshots.tzinit.org/) launched recently, it would be nice to replace xtz-shots with it in `tezos-setup`.

This provider has different API than `Marigold` and `xtz-shots`, namely it doesn't provide the snapshot metadata JSON file, but their API is designed in such a way as to download the needed snapshot without metadata JSON.

Also note that they have several servers located in different regions (EU, US and Asia) and it's represented by the part of URL to download the snapshot. We should take it into account and give users a possibility to choose the appropriate region to download the snapshot faster. 

# Acceptance criteria

* `xtz-shots` is removed from the list of available providers and replaced by `tzinit`

krendelhoff2:
## Description

TzInit provider should be supported, as it seems to be the most stable in the near future.

## Related issue(s)

Resolves #782 

  can be successfully built.

### DIFF
--- a/baking/src/tezos_baking/provider.py
+++ b/baking/src/tezos_baking/provider.py
@@ -28,7 +28,6 @@ def get_node_version():
 @dataclass
 class Provider:
     title: str
-    metadata_url: str
 
     @abstractmethod
     def get_snapshot_metadata(self, network, history_mode):
@@ -37,6 +36,7 @@ class Provider:
 
 @dataclass
 class XtzShotsLike(Provider):
+    metadata_url: str
     # Returns relevant snapshot's metadata
     # It filters out provided snapshots by `network` and `history_mode`
     # provided by the user and then follows this steps:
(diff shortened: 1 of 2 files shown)


baking/src/tezos_baking/provider.py
```diff
@@ -0,0 +1,250 @@
...
+                nbytes /= 1024.0
+                i += 1
+            f = ("%.2f" % nbytes).rstrip("0").rstrip(".")
+            return "%s%s" % (f, suffixes[i])
+        return content_length
+
+    def get_snapshot_metadata(self, network, history_mode):
+        history_mode = "full" if history_mode == "archive" else history_mode
+        self.metadata_url = (
+            f"https://snapshots.eu.tzinit.org/{network}/{history_mode}.json"
```
DMozhevitin, line 220: Is it possible to add one more step that asks user to select a closest region? It would be nice to utilize this feature of `tzinit` provider

baking/src/tezos_baking/provider.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+compatible_snapshot_version = 7
+
+default_providers = [
+    TzInit("tzinit"),
+    XtzShotsLike(
+        "marigold.dev", "https://snapshots.tezos.marigold.dev/api/tezos-snapshots.json"
+    ),
+    XtzShotsLike("xtz-shots.io", "https://xtz-shots.io/tezos-snapshots.json"),
```
DMozhevitin, line 247: I'd suggest to completely remove `xtz-shots` from the list of default providers since it's unlikely that it will become available again

In this case, `XtzShotsLike` worth renaming to `MarigoldLike` or just `Marigold` since we don't have other providers with similar API for now

DMozhevitin (changes requested):
Looks really good :+1: 

Left only a couple of minor comments

DMozhevitin (approved):
LGTM :+1:, thanks for quickly addressing my comments

One minor thing left: there are still a couple of mentions of `xtz-shots` in the help text of some steps, they need to be replaced by `tzinit` as well. Feel free to merge the PR after changing it

